PR #3, opened 2021-09-22
network: Add types and deserializer for NetworkOptions

flouthoc:
Following PR contains types which will be consumed by netavark.

* Types
* Deserializer to consume types from files.
* Serializer incase we need to flush back types back to json. (We can remove this later if not needed).

### DIFF
--- a/src/network/mod.rs
+++ b/src/network/mod.rs
@@ -1,0 +1,55 @@
+// Crate contains the types which are accepted by netvark.
+
+extern crate serde_derive;
+use std::collections::HashMap;
+
+// NetworkOptions for a given container.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct NetworkOptions {
+    #[serde(rename = "container_id")]
+    pub container_id: String,
+
+    #[serde(rename = "container_name")]
+    pub container_name: String,
+
+    #[serde(rename = "networks")]
+    pub networks: HashMap<String, PerNetworkOptions>,
+
+    #[serde(rename = "port_mappings")]
+    pub port_mappings: Option<Vec<PortMapping>>,
+}
+
+// PerNetworkOptions are options which should be set on a per network basis
+#[derive(Debug, Serialize, Deserialize)]
+pub struct PerNetworkOptions {
+    #[serde(rename = "aliases")]
+    pub aliases: Option<Vec<String>>,
+
+    #[serde(rename = "interface_name")]
+    pub interface_name: String,
+
+    #[serde(rename = "static_ips")]
+    pub static_ips: Option<Vec<String>>,
+
+    #[serde(rename = "static_mac")]
+    pub static_mac: Option<String>,
+}
+
+// PortMapping is one or more ports that will be mapped into the container.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct PortMapping {
+    #[serde(rename = "container_port")]
+    pub container_port: u16,
+
+    #[serde(rename = "host_ip")]
+    pub host_ip: String,
+
+    #[serde(rename = "host_port")]
+    pub host_port: u16,
+
+    #[serde(rename = "protocol")]
+    pub protocol: String,
+
+    #[serde(rename = "range")]
+    pub range: u16,
+}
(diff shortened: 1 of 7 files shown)


flouthoc:
@Luap99 @mheon PTAL

Luap99:
We also have to deserialize the Network struct and the `map[string]StatusBlock` must be serialized and returned as result to podman for the setup command.

flouthoc:
@Luap99 oh okay but i thought this only accepts `NetworkOptions` as input according to https://github.com/containers/podman/blob/main/libpod/network/types/network.go#L164 `NetworkOption` never resolves to `Network`.

src/network/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    #[serde(rename = "static_mac")]
+    pub static_mac: Option<String>,
+}
+
+// PortMapping is one or more ports that will be mapped into the container.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct PortMapping {
+    #[serde(rename = "container_port")]
+    pub container_port: i64,
```
Luap99, line 42: This should be `u16`

src/network/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    #[serde(rename = "host_ip")]
+    pub host_ip: Option<String>,
+
+    #[serde(rename = "host_port")]
+    pub host_port: Option<i64>,
```
Luap99, line 48: This should also be `u16` and it should not be optional at this point.

src/network/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    #[serde(rename = "protocol")]
+    pub protocol: Option<String>,
+
+    #[serde(rename = "range")]
+    pub range: Option<i64>,
```
Luap99, line 54: also `16`

src/network/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// PerNetworkOptions are options which should be set on a per network basis
+#[derive(Debug, Serialize, Deserialize)]
+pub struct PerNetworkOptions {
+    #[serde(rename = "aliases")]
+    pub aliases: Option<Vec<String>>,
+
+    #[serde(rename = "interface_name")]
+    pub interface_name: Option<String>,
```
Luap99, line 29: Interface name should not be optional. I assume the `Option` means that this can be empty or what does this do in rust?

Luap99:
@flouthoc Yes the input in the setup call onlt uses network options, however we need to read the networks for this to work. The interface already has the networks. To make this work with this extra binary we need to read network struct, e.g. we have to know what Subnet to use, etc...

flouthoc:
@Luap99 ah makes sense will create another PR for that. Limiting this PR to `NetoworkOptions` only

Luap99:
> @Luap99 ah makes sense will create another PR for that. Limiting this PR to `NetoworkOptions` only

understood

src/network/mod.rs
flouthoc: @Luap99 this wraps the value so `None` can be parsed this does not necessarily means that this is optional. We can enforce mandatory checks in business logic struct shouldn't specify that.

Luap99: Thanks for clarifying this. I do not really understand the rust syntax yet and this wasn't clear to me from the rust documentation.

Luap99: Is there a reason that container_port does not use Optional?

flouthoc: @Luap99 Since `containerPort` is missing omitempty here https://github.com/containers/podman/blob/main/libpod/network/types/network.go#L184 it is assumed that this will be always present and can never be `None`

flouthoc: @Luap99 We can remove `Option<` from eveything but then it would mean that all the values have to be filled by some values they cannot be `None`.

flouthoc: @Luap99 Could you remove `omitempty` from go types which are mandatory ?

Luap99: @flouthoc Yes I will do this.

mheon:
Additional PR SGTM. We probably should make a struct on the Go end as well that unifies a NetworkOptions and the full set of Networks, so we have code written to interface on both sides.

src/network/mod.rs
flouthoc: Updated after https://github.com/containers/podman/pull/11692

Luap99 (commented):
LGTM

mheon:
LGTM

PR #5, opened 2021-09-23
 network: Refactor mod structure and add types for StatusBlock,Network,Subnet,Leaserange 

flouthoc:
Following PR 
* Adds types for `StatusBlock,Network,Subnet,Leaserange `
* Refactor `network` module structure into more generic one.

### DIFF
--- a/src/network/network.rs
+++ b/src/network/network.rs
@@ -125,7 +125,8 @@ pub struct NetAddress {
     pub gateway: Option<String>,
 
     #[serde(rename = "subnet")]
-    pub subnet: IpNet,
+    //TODO: Cast this to IpNet
+    pub subnet: String,
 }
 
 // IPNet is used as custom net.IPNet type to add Marshal/Unmarshal methods.
@@ -148,7 +149,8 @@ pub struct Subnet {
     pub lease_range: Option<LeaseRange>,
 
     #[serde(rename = "subnet")]
-    pub subnet: IpNet,
+    //TODO: Cast this to IpNet
+    pub subnet: String,
 }
 
 // LeaseRange contains the range where IP are leased.
--- a/src/network/network.rs
+++ b/src/network/network.rs
@@ -53,3 +53,48 @@ pub struct PortMapping {
     #[serde(rename = "range")]
     pub range: u16,
 }
+
+// StatusBlock contains the network information about a container
+// connected to one Network.
+#[derive(Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


flouthoc:
@Luap99 @mheon PTAL

src/network/network.rs
```diff
@@ -0,0 +1,162 @@
...
+// IPNet is used as custom net.IPNet type to add Marshal/Unmarshal methods.
+// TODO: [Not important] Transformed from golang net.IPNet. Swtich to native rust if needed.
+#[derive(Serialize, Deserialize)]
+pub struct IpNet {
+    #[serde(rename = "IP")]
+    pub ip: String,
+
+    #[serde(rename = "Mask")]
+    pub mask: String,
+}
```
Luap99, line 140: Golang should unmarshal this into a string with the format `IP/CIDIR` e.g to `192.168.1.1/24` or `fd80::1/64` for ipv6.
Can rust read this correctly? Is there a better type than string for a ip address?

flouthoc: @Luap99 I have specially mentioned this in a TODO line just above this struct , yes as of now this will work without any problem. We can switch a rust native type later or cast it to a native type during business logic.

flouthoc: So parsing for json should not be a problem at all.

flouthoc: Edit: @Luap99 i think we should store IpNet as a single string of `IP/CIDR` and do the parsing ourselves. I'll edit this type.

flouthoc: @Luap99 Added a new commit https://github.com/containers/netavark/pull/5/commits/07888fb3783aee2a444cfa8074ac1a2ca1723ad0 we are not using `IpNet` as of now instead accepting Subnet as string and we can cast it to `IpNet`. Will add automatic sterilizer for this later as an improvement.

mheon:
Can we get some tests for this? We can get test data by writing a little Go program to serialize the Podman types, and then just include the test data - no need to include the Go code in the actual repo.

flouthoc:
@mheon Sure could do this. Infact i was just about to start writing some early tests for this. Should tests be part of this PR or seperate PR is fine.

mheon:
Separate PR is fine

flouthoc:
@Luap99 @mheon final review and merge needed on this one.

mheon:
LGTM

PR #13, opened 2021-10-05
Add cirrus testing

baude:
Basic wireup for testing in the cirrus container.  At first, we just use
a container.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,18 @@
+---
+
+test_task:
+  alias: "build_and_test"
+  name: "Build and Test"
+  container:
+    image: docker.io/library/rust
+  install_script: cargo build
+  test_script: cargo test
+
+success_task:
+  name: "Total success"
+  depends_on:
+    - "build_and_test"
+  clone_script: /bin/true
+  script: /bin/true
+  container: 
+    image: docker.io/library/alpine


lsm5 (approved):
LGTM 

@cevich PTAL

flouthoc:
@cevich LGTM final merge plz

.cirrus.yml
```diff
@@ -0,0 +1,18 @@
+---
+
+test_task:
+  alias: "build_and_test"
```
cevich, line 4: The alias name needs to match the map name, so this should be changed to `alias: test` or the map should be renamed to `build_and_test_task`.  Getting this wrong will break `hack/get_ci_vm.sh` assuming it's needed in the future.

.cirrus.yml
```diff
@@ -0,0 +1,18 @@
...
+  name: "Build and Test"
+  container:
+    image: docker.io/library/rust
```
cevich, line 7: This will break if/when the pull rate limit is exceeded.  A new quay.io namespace should probably be created for netavark, and this image copied there manually.  This will also prevent sudden surprises by unexpected changes to the image (this happened before with the registry image).

.cirrus.yml
```diff
@@ -0,0 +1,18 @@
...
+  install_script: cargo build
+  test_script: cargo test
+
+success_task:
```
cevich, line 11: This needs `alias: success` otherwise it will break a future `hack/get_ci_vm.sh`.

PR #19, opened 2021-10-11
Fix logging with the env_logger crate

mheon:
We didn't have a log implementation set, so nothing was being printed. Use env_logger for now, we can reevaluate this if/when we need more advanced log handling ability.

### DIFF
--- a/src/test/config/setupopts.test.json
+++ b/src/test/config/setupopts.test.json
@@ -14,5 +14,15 @@
       "defaultNetwork":{
          "interface_name":"eth0"
       }
+   },
+   "network_info":{
+      "defaultNetwork":{
+         "dns_enabled":true,
+         "driver":"bridge",
+         "id":"53ce4390f2adb1681eb1a90ec8b48c49c015e0a8d336c197637e7f65e365fa9e",
+         "internal":false,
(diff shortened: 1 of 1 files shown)


baude:
should the src/test json file be updated as well ?

src/network/network.rs
```diff
@@ -55,6 +55,9 @@ pub struct NetworkOptions {
     #[serde(rename = "networks")]
     pub networks: HashMap<String, PerNetworkOptions>,
 
+    #[serde(rename = "network_info")]
+    pub network_info: HashMap<String, Network>,
```
flouthoc, line 59: We either need to make this optional or if it is mandatory this needs to be added into test stub data.

src/test/config/setupopts.test.json
```diff
@@ -14,5 +14,13 @@
       "defaultNetwork":{
          "interface_name":"eth0"
       }
+   },
+   "network_info":{
```
flouthoc, line 18: I think it `network_info` takes a map of `stirng`: `<Network>`. I think this is a direct <Network>

flouthoc (approved):

PR #20, opened 2021-10-12
network: add debug to types so serde can implement debug if traits need it.

flouthoc:
Needed by https://github.com/containers/netavark/pull/19


lsm5 (approved):
LGTM

PR #30, opened 2021-10-14
iptables

baude:
implement iptables rules for NAT and MASQ setup.  At this time, no tear
down is being proposed.

### DIFF
--- a/src/network/types.rs
+++ b/src/network/types.rs
@@ -2,6 +2,7 @@
 
 extern crate serde_derive;
 use std::collections::HashMap;
+use ipnet::IpNet;
 
 // Network describes the Network attributes.
 #[derive(Debug, Serialize, Deserialize)]
@@ -132,17 +133,6 @@ pub struct NetAddress {
     pub subnet: String,
 }
 
-// IPNet is used as custom net.IPNet type to add Marshal/Unmarshal methods.
-// TODO: [Not important] Transformed from golang net.IPNet. Swtich to native rust if needed.
-#[derive(Debug, Serialize, Deserialize)]
-pub struct IpNet {
-    #[serde(rename = "IP")]
-    pub ip: String,
-
-    #[serde(rename = "Mask")]
-    pub mask: String,
-}
-
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Subnet {
     #[serde(rename = "gateway")]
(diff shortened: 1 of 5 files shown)


src/commands/setup.rs
```diff
@@ -29,7 +33,66 @@ impl Setup {
...
+                self.setup_iptables(&network_name, nets);
+            }
+        }
         debug!("{:?}", "Setup complete");
-        //()
+    }
+    /// Set up required iptables for NAT
+    ///
+    ///
+    pub fn setup_iptables(&self, network_name: &str, subnets: Vec<Subnet>) {
```
flouthoc, line 49: I think can have this function in `network` create not at the frontend.

src/commands/setup.rs
```diff
@@ -29,7 +33,66 @@ impl Setup {
...
+        debug!("iptables");
+        for net in subnets{
+            // Basic object setup
+            let ipt = iptables::new(false).unwrap();
```
flouthoc, line 53: not-a-blocker: unwrap from serializers might be okay but for iptables we might wanna handle error. Can revisit this later.

src/commands/setup.rs
baude: can you be more exact. point to exactly where

src/commands/setup.rs
```diff
@@ -29,7 +33,66 @@ impl Setup {
...
+            //  Check if the chain exists, if not - create it
+            let chain_check = ipt.chain_exists(NAT, network_name);
+            match chain_check {
+                Ok(true) => debug!("{} chain exists", &network_name),
+                Ok(false) => {
+                    // The chain did not exist
+                    debug!("need to create chain {}", network_name);
+                    match ipt.new_chain(NAT, network_name) {
+                        Ok(_) => debug!("chain {} created", network_name),
+                        Err(e) => panic!("{}", e)
```
flouthoc, line 63: We can return CliError instead of panic. Can revisit this later.

src/network/types.rs
```diff
@@ -134,14 +135,14 @@ pub struct NetAddress {
...
 // TODO: [Not important] Transformed from golang net.IPNet. Swtich to native rust if needed.
-#[derive(Debug, Serialize, Deserialize)]
-pub struct IpNet {
-    #[serde(rename = "IP")]
-    pub ip: String,
-
-    #[serde(rename = "Mask")]
-    pub mask: String,
-}
+// #[derive(Debug, Serialize, Deserialize)]
```
flouthoc, line 138: We can remove this if its not needed at all.

src/commands/setup.rs
flouthoc: @baude we could create a new file `src/network/core.rs` maybe like this ->  https://github.com/containers/netavark/pull/17

PR #31, opened 2021-10-14
replace calls to unwrap with expect

lsm5:
expect is similar to unwrap with the added feature of calling the panic!
macro. Should be friendlier to identify panic messages rather than the
generic messages from unwrap.

@baude @mheon @flouthoc PTAL, rust n00b here, so be gentle :)


lsm5:
related: might be good to hook rustfmt into ci.

lsm5:
> related: might be good to hook rustfmt into ci.

filed https://github.com/containers/netavark/issues/33

PR #34, opened 2021-10-14
CI and Makefile updates

lsm5:
- Run rustfmt in `make validate` and CI validation
- Makefile: allow modification of PREFIX, DATADIR and MANDIR
- .cirrus.yml uses make targets instead of cargo commands
- Use `mandown` for manpage generation instead of `go-md2man`

Needs rustfmt(-nightly) and mandown crates installed in CI image.

@flouthoc @baude @mheon @Luap99 @ashcrow PTAL

We'll need rustfmt crate in quay.io/libpod/netavark-devel to get this working. 

Also, once this is merged, I expect a few things to go kaboom.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,15 +5,15 @@ test_task:
   name: "Build and Test"
   container:
     image: quay.io/libpod/netavark-devel
-  install_script: cargo build
-  test_script: cargo test
+  install_script: make all
+  test_script: make test
 
 test_task:
   alias: "validate_test"
   name: "Validate Code"
   container:
     image: quay.io/libpod/netavark-devel
-  test_script: cargo clippy -p netavark -- -D warnings
+  test_script: make validate
 
 success_task:
   name: "Total success"
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,7 +1,7 @@
-PREFIX := /usr/local
-DATADIR := ${PREFIX}/share
-MANDIR := $(DATADIR)/man
-GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
+PREFIX ?= /usr/local
+DATADIR ?= ${PREFIX}/share
+MANDIR ?= $(DATADIR)/man
+MANDOWN ?= $(shell command -v mandown)
 
 docs: $(patsubst %.md,%,$(wildcard *.md))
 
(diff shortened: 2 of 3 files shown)


.cirrus.yml
```diff
@@ -13,7 +13,9 @@ test_task:
   name: "Validate Code"
   container:
     image: quay.io/libpod/netavark-devel
-  test_script: cargo clippy -p netavark -- -D warnings
+  test_script:
+    - cargo fmt --all -- --check
+    - cargo clippy -p netavark -- -D warnings
```
lsm5, line 18: maybe it's best to just run `make validate` here? @baude @cevich wdyt?

lsm5:
> SGTM, as long as the output is valid

ack, generates a `netavark.1` just fine on my machine.

cevich:
Suggestion (maybe for another PR) setup and use https://pre-commit.ci/ instead of `DCO` to perform the basic validations.  Reason: It's highly extensible and scaleable.  Plus the same set of validation plugins is available to be [used locally with the git pre-commit hook](https://pre-commit.com/) (available in Fedora).

lsm5:
> Suggestion (maybe for another PR) setup and use https://pre-commit.ci/ instead of `DCO` to perform the basic validations. Reason: It's highly extensible and scaleable. Plus the same set of validation plugins is available to be [used locally with the git pre-commit hook](https://pre-commit.com/) (available in Fedora).

@cevich thanks, looking at `.pre-commit-config.yaml` in podman, I do see a bunch of ids but nothing that gets rid of DCO. Is that yet to be done over there or am I not reading it right?

cevich:
Again, I'm totally cool with leaving a more full-feature validation job up to another PR.

lsm5:
kinda sorta conflicts with https://github.com/containers/netavark/pull/36

docs/Makefile
```diff
@@ -1,7 +1,7 @@
-PREFIX := /usr/local
-DATADIR := ${PREFIX}/share
-MANDIR := $(DATADIR)/man
-GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
+PREFIX ?= /usr/local
+DATADIR ?= ${PREFIX}/share
+MANDIR ?= $(DATADIR)/man
+MANDOWN ?= $(shell command -v mandown || echo '$(PATH)/mandown')
```
Luap99, line 4: How/Where is this installed?

Also appending like to $PATH is wrong, $PATH contains more than one path usually.  `command` already looks up $PATH so why try to use it anyway

lsm5: mandown will be installed in the rust ci image. I'll remove that PATH part.

lsm5: @Luap99 done. PTAL

PR #35, opened 2021-10-14
serialize: Use buffered reads/writes

cgwalters:
A not obvious trap in Rust is that e.g. `File` is just a raw wrapper
for a file descriptor, and e.g. serde will happily try to do things
like invoke `read()/write()` a byte or 4 bytes at a time.

Buffering is important to avoid lots of syscall traffic, particularly
in the age of Spectre etc.


flouthoc:
Thanks again @cgwalters for that patch. I hope it resolves your doubt.

PR #36, opened 2021-10-15
Add rustfmt

baude:
Add a fmt linter for rust to the CI

### DIFF
--- a/.github/workflows/rust-lints.yml
+++ b/.github/workflows/rust-lints.yml
@@ -1,0 +1,31 @@
+---
+name: Rust
+
+on:
+  push:
+    branches: [main]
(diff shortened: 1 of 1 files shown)


cevich:
...so the issue with using github-actions and Cirrus-CI is, it's really tricky to have any dependency between them.  So for example, we cannot tell a bot to green-light a merge once a final `Success` Cirrus (or github-actions) job completes, without a HUGE-pile of duct-tape and string to glue them together.  Lastly, since github-actions is a terrible overall architecture (IMHO) and doesn't scale well at all, it's best if we stick to the much simpler/easy for everyone to understand Cirrus-CI :grinning:

PR #45, opened 2021-10-20
Add rustfmt to netavark image

baude:
We want to be able to run rustfmt to validate code.


mheon:
LGTM

lsm5:
@baude hmm, how about we replace `rustup component add` with `cargo install` then? `cargo install` does update the index automatically.

PR #37, opened 2021-10-18
validate namespace before setup and teardown

lsm5:
Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

PR #47, opened 2021-10-26
JSON errors to podman

lsm5:
rebased on #37

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -1,4 +1,5 @@
 //! Configures the given network namespace with provided specs
+use crate::error::NetavarkError;
 use crate::firewall;
 use crate::network;
 use crate::network::types;
@@ -33,7 +34,12 @@ impl Setup {
 
         let network_options = match network::types::NetworkOptions::load(&input_file) {
             Ok(opts) => opts,
-            Err(e) => panic!("{}", e),
+            Err(e) => {
+                return Err(Box::new(NetavarkError {
+                    error: format!("{}", e),
+                    errno: 1,
+                }));
+            }
         };
 
         let firewall_driver = match firewall::get_supported_firewall_driver() {
--- a/src/commands/teardown.rs
+++ b/src/commands/teardown.rs
@@ -1,3 +1,4 @@
+use crate::error::NetavarkError;
 use crate::{firewall, network};
 use clap::{self, Clap};
 use log::debug;
@@ -24,7 +25,12 @@ impl Teardown {
         debug!("{:?}", "Tearing down..");
         let network_options = match network::types::NetworkOptions::load(&input_file) {
             Ok(opts) => opts,
-            Err(e) => panic!("{}", e),
+            Err(e) => {
+                return Err(Box::new(NetavarkError {
+                    error: format!("{}", e),
+                    errno: 1,
+                }));
+            }
         };
 
         let _firewall_driver = match firewall::get_supported_firewall_driver() {
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,4 @@
 use clap::{crate_version, Clap};
-use std::error::Error;
 
 use netavark::commands::setup;
 use netavark::commands::teardown;
@@ -25,14 +24,32 @@ enum SubCommand {
     Teardown(teardown::Teardown),
 }
 
-fn main() -> Result<(), Box<dyn Error>> {
+fn main() {
     env_logger::init();
     let opts = Opts::parse();
 
     let file = opts.file.unwrap_or_else(|| String::from("/dev/stdin"));
     match opts.subcmd {
-        SubCommand::Setup(setup) => setup.exec(file),
-        SubCommand::Teardown(teardown) => teardown.exec(file),
+        SubCommand::Setup(setup) => {
+            if let Err(err) = setup.exec(file) {
+                let er = netavark::error::NetavarkError {
+                    error: format!("{}", err),
+                    errno: 1,
+                };
+                println!("{}", er.to_string());
+                std::process::exit(er.errno);
+            }
+        }
+        SubCommand::Teardown(teardown) => {
+            if let Err(err) = teardown.exec(file) {
+                let er = netavark::error::NetavarkError {
+                    error: format!("{}", err),
+                    errno: 1,
+                };
+                println!("{}", er.to_string());
+                std::process::exit(er.errno);
+            }
+        }
     }
 }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,0 +1,16 @@
+use std::error::Error;
+use std::fmt;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct NetavarkError {
+    pub error: String,
+    pub errno: i32,
+}
+
+impl fmt::Display for NetavarkError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", serde_json::to_string(&self).unwrap())
+    }
+}
+
+impl Error for NetavarkError {}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,5 +5,6 @@ extern crate serde_json;
 extern crate simple_error;
 
 pub mod commands;
+pub mod error;
 pub mod firewall;
 pub mod network;


src/network/error.rs
```diff
@@ -0,0 +1,15 @@
+use std::error::Error;
+use std::fmt;
+
+#[derive(Debug)]
+pub struct NetavarkError {
+    pub error: String,
```
flouthoc, line 6: I think we would another field for `errno`. It could be empty or standard `-1` if not specified but we could fill it with number wherever possible.

src/network/error.rs
```diff
@@ -0,0 +1,16 @@
...
+    pub errorno: i32,
```
lsm5, line 7: need to change the name, pushing change.

lsm5:
@flouthoc errno: -1 should be for `other` errors, right? I'll need to change that.

src/commands/setup.rs
```diff
@@ -31,7 +31,12 @@ impl Setup {
 
         let network_options = match network::types::NetworkOptions::load(&input_file) {
             Ok(opts) => opts,
-            Err(e) => panic!("{}", e),
+            Err(_) => {
```
flouthoc, line 34: I think `e` here should be passed into the `error` message instead of ignoring by (`_`). We could use something like `format!("Netvark error {}", e).to_string()` or something better. But overall I think we can't ignore original error.

flouthoc:
> @flouthoc errno: -1 should be for `other` errors, right? I'll need to change that.

@lsm5 Yes for any error which does have a code.

src/commands/setup.rs
lsm5: ack

lsm5: done

lsm5:
@baude @mheon @Luap99 PTAL

src/commands/setup.rs
```diff
@@ -31,7 +31,12 @@ impl Setup {
 
         let network_options = match network::types::NetworkOptions::load(&input_file) {
             Ok(opts) => opts,
-            Err(e) => panic!("{}", e),
+            Err(e) => {
+                return Err(Box::new(network::error::NetavarkError {
+                    error: format!("Netavark Error {}", e),
```
Luap99, line 36: Please do not add `Netavark Error` to the error message. This provides zero context, when I call netavark I know I got an error from netavark. This message should be something like `failed to load network options: {}, err`

src/commands/setup.rs
```diff
@@ -31,7 +31,12 @@ impl Setup {
...
+                    errno: -1,
```
Luap99, line 37: Should this be the exit code? If so it should be greater than 0.

src/commands/teardown.rs
```diff
@@ -24,7 +24,12 @@ impl Teardown {
         debug!("{:?}", "Tearing down..");
         let _network_options = match network::types::NetworkOptions::load(&input_file) {
             Ok(opts) => opts,
-            Err(e) => panic!("{}", e),
+            Err(e) => {
+                return Err(Box::new(network::error::NetavarkError {
+                    error: format!("Netavark Error {}", e),
+                    errno: -1,
```
Luap99, line 30: same here

src/commands/setup.rs
flouthoc: I think we should decide on some common error code it code is not already present. I suggested for `-1` if error code is not known. But i think we can move to something standard.

Luap99:
Also this does not reurn valid json:
I get `Error: NetavarkError { error: "Netavark Error missing field `container_id` at line 1 column 2", errno: -1 }`
It should be `{ error: "Netavark Error missing field `container_id` at line 1 column 2", errno: -1 }`

And the json must be printed on stdout not stderr.

lsm5:
@flouthoc @Luap99 i pushed a change to print a json to stdout, but not sure if that's how it's to be done. PTAL :)

lsm5:
fwiw, it's just printing in place, for returning, I wonder if the return type gotta change like @flouthoc mentioned.

src/commands/setup.rs
```diff
@@ -31,7 +31,19 @@ impl Setup {
 
         let network_options = match network::types::NetworkOptions::load(&input_file) {
```
flouthoc, line 32: This entire block could be simplified into 
```rust
        let network_options = match network::types::NetworkOptions::load(&input_file) {
            Ok(opts) => opts,
            Err(e) => {
                return Err(Box::new(network::error::NetavarkError{ error: format!("{}", e), errno: 1}));
            }
        };
```

src/commands/teardown.rs
```diff
@@ -24,7 +24,20 @@ impl Teardown {
         debug!("{:?}", "Tearing down..");
         let _network_options = match network::types::NetworkOptions::load(&input_file) {
             Ok(opts) => opts,
-            Err(e) => panic!("{}", e),
+            Err(e) => {
```
flouthoc, line 27: this could be simplified into 
```rust
        let network_options = match network::types::NetworkOptions::load(&input_file) {
            Ok(opts) => opts,
            Err(e) => {
                return Err(Box::new(network::error::NetavarkError{ error: format!("{}", e), errno: 1}));
            }
        };
```

src/network/error.rs
```diff
@@ -0,0 +1,16 @@
+use std::error::Error;
+use std::fmt;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct NetavarkError {
+    pub error: String,
+    pub errno: i32,
+}
+
+impl fmt::Display for NetavarkError {
```
flouthoc, line 10: display should do a serde conversion like
```rust
impl fmt::Display for NetavarkError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "{}", serde_json::to_string(&self).unwrap())
    }
}
```

flouthoc:
With other changes `main.rs` must be modified to return a string like
```diff
diff --git a/src/main.rs b/src/main.rs
index a8161a2..498d656 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,4 @@
 use clap::{crate_version, Clap};
-use std::error::Error;
 
 use netavark::commands::setup;
 use netavark::commands::teardown;
@@ -25,14 +24,22 @@ enum SubCommand {
     Teardown(teardown::Teardown),
 }
 
-fn main() -> Result<(), Box<dyn Error>> {
+fn main() {
     env_logger::init();
     let opts = Opts::parse();
 
     let file = opts.file.unwrap_or_else(|| String::from("/dev/stdin"));
     match opts.subcmd {
-        SubCommand::Setup(setup) => setup.exec(file),
-        SubCommand::Teardown(teardown) => teardown.exec(file),
+        SubCommand::Setup(setup) => {
+            if let Err(err) = setup.exec(file) {
+                println!("{}", err.to_string())
+            }
+        }
+        SubCommand::Teardown(teardown) => {
+            if let Err(err) = teardown.exec(file) {
+                println!("{}", err.to_string())
+            }
+        }
     }
 }
 
```

lsm5:
@flouthoc thanks for the diff. PTAL at latest version.

flouthoc:
LGTM. We still need a final call if we want main to return error on stdout.

mheon:
STDOUT is correct. I'll do a final review after lunch.

Luap99 (commented):
I just tried this patch and I still get non json errors, e.g.
`failed configure bridge and veth interface: Custom { kind: Other, error: "Error while configuring network interface failed to create a pair of veth interfaces: Received a netlink error message File exists (os error 17):" }`
I think you have to wrap all errors as NetavarkError

lsm5:
> I just tried this patch and I still get non json errors, e.g. `failed configure bridge and veth interface: Custom { kind: Other, error: "Error while configuring network interface failed to create a pair of veth interfaces: Received a netlink error message File exists (os error 17):" }` I think you have to wrap all errors as NetavarkError

ack, I can work on changing that if everyone agrees. @flouthoc @mheon @baude wdyt?

mheon:
That approach SGTM - as long as we print sensible error strings (and it should) and maintain the JSON error semantics

lsm5:
@Luap99 PTAL

flouthoc:
I am still wondering if why did wrapping failed at final level. It should have worked without wrapping error across the code-base. Let me try @lsm5 's older commit.

mheon:
Curious about this as well, I don't know how maintainable this is as implemented.

lsm5:
@Luap99 @mheon @baude @flouthoc PTAL.

flouthoc:
LGTM

mheon:
LGTM on my end, but would prefer we get confirmation from @Luap99 that this works with Podman pre-merge

Luap99:
The exit code is not correct, netavark will exit with 0 despite the error. Also the error message is horrible, I do not think we should wrap the errors that often or at least use only the underlying message:
`{"error":"failed configure bridge and veth interface: Custom { kind: Other, error: \"Error while configuring network interface from container namespace: Custom { kind: Other, error: \\\"interface eth0 already exists on container namespace\\\" }:\" }","errno":1}`

lsm5:
@Luap99 can you please send me the steps you use for testing this?

flouthoc:
@Luap99 We need to remove `std::io::error` wrapping from below the code. `NetavarkError` cannot control that and we also need to figure out a way to propagate error codes.

Luap99:
@lsm5 Checkout my Podman PR and run `sudo NETAVARK_BINARY=/path/to/bin/netavark   go test -v -run -mod=vendor  ./libpod/network/netavark/`. The `test netavark error` test should pass.

If you just want to check netavark just run setup command twice:
run `sudo unshare -n` in two terminals in the second run `echo $$` to get the pid
in the first terminal run `bin/netavark -f ./src/test/config/setupopts2.test.json teardown /proc/<pid second terminal>/ns/net`
if you run this command twice you should get an error returned.

mheon:
Do we need to propagate error codes?

mheon:
To elaborate - I don't think it matters, as long as errors result in non-zero exit code for errors. We can't really react to these errors in a meaningful way in Podman.

Luap99:
> To elaborate - I don't think it matters, as long as errors result in non-zero exit code for errors. We can't really react to these errors in a meaningful way in Podman.

The only interesting case would be to know if it failed while decoding the json input. In this case I could handle it for teardown so that I do not free the leased ips

lsm5:
Would this approach work:
```
diff --git a/src/network/core.rs b/src/network/core.rs
index dbe4e0a..918173d 100644
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -238,7 +238,7 @@ impl Core {
         per_network_opts: &types::PerNetworkOptions,
         network: &types::Network,
         netns: &str,
-    ) -> Result<(), std::io::Error> {
+    ) -> anyhow::Result<()> {
         let container_veth_name: String = per_network_opts.interface_name.to_owned();
         let _subnets = network.subnets.as_ref().unwrap();

@@ -247,12 +247,7 @@ impl Core {
             container_veth_name
         );

-        if let Err(err) = Core::remove_container_veth(&container_veth_name, netns) {
-            return Err(std::io::Error::new(
-                std::io::ErrorKind::Other,
-                format!("unable to remove container veth: {:?}", err),
-            ));
-        }
+        Core::remove_container_veth(&container_veth_name, netns)?;

         debug!("Container veth removed: {:?}", container_veth_name);
```

Before:
```
$ ./bin/netavark -f ./src/test/config/setupopts2.test.json teardown foo
{"error":"unable to remove container veth: Custom { kind: Other, error: \"failed to open network namespace: No such file or directory (os error 2)\" }","errno":1}
```

After: 
```
$ ./bin/netavark -f ./src/test/config/setupopts2.test.json teardown foo
{"error":"failed to open network namespace: No such file or directory (os error 2)","errno":1}
```

Would this affect maintainability?

@Luap99 I tried running the podman test you mentioned on podman main branch, AFAICT your PR merged, but that command ran 0 tests. Let me know if I missed anything.

Also, running the teardown command twice like you suggested doesn't barf even on main branch. So, I'd suspect teardown itself needs some fixing?

flouthoc:
> Would this approach work:
> 
> ```
> diff --git a/src/network/core.rs b/src/network/core.rs
> index dbe4e0a..918173d 100644
> --- a/src/network/core.rs
> +++ b/src/network/core.rs
> @@ -238,7 +238,7 @@ impl Core {
>          per_network_opts: &types::PerNetworkOptions,
>          network: &types::Network,
>          netns: &str,
> -    ) -> Result<(), std::io::Error> {
> +    ) -> anyhow::Result<()> {
>          let container_veth_name: String = per_network_opts.interface_name.to_owned();
>          let _subnets = network.subnets.as_ref().unwrap();
> 
> @@ -247,12 +247,7 @@ impl Core {
>              container_veth_name
>          );
> 
> -        if let Err(err) = Core::remove_container_veth(&container_veth_name, netns) {
> -            return Err(std::io::Error::new(
> -                std::io::ErrorKind::Other,
> -                format!("unable to remove container veth: {:?}", err),
> -            ));
> -        }
> +        Core::remove_container_veth(&container_veth_name, netns)?;
> 
>          debug!("Container veth removed: {:?}", container_veth_name);
> ```
> 
> Before:
> 
> ```
> $ ./bin/netavark -f ./src/test/config/setupopts2.test.json teardown foo
> {"error":"unable to remove container veth: Custom { kind: Other, error: \"failed to open network namespace: No such file or directory (os error 2)\" }","errno":1}
> ```
> 
> After:
> 
> ```
> $ ./bin/netavark -f ./src/test/config/setupopts2.test.json teardown foo
> {"error":"failed to open network namespace: No such file or directory (os error 2)","errno":1}
> ```
> 
> Would this affect maintainability?
> 
> @Luap99 I tried running the podman test you mentioned on podman main branch, AFAICT your PR merged, but that command ran 0 tests. Let me know if I missed anything.
> 
> Also, running the teardown command twice like you suggested doesn't barf even on main branch. So, I'd suspect teardown itself needs some fixing?

@lsm5 One Problem with this is that we are dropping error context `unable to remove container veth` but anyways I found the actual cause of weird wrapping for error which was use of `{:?}` while formatting. Fixed formatting in https://github.com/containers/netavark/pull/74 and also tested with your PR. So one #74 gets merged we can rebase this one should fix formatting and wrapping issues.

flouthoc:
We should rebase this PR after https://github.com/containers/netavark/pull/74

lsm5:
rebased on main, PTAL.

flouthoc:
LGTM. Ignoring the error code propagation part not sure but i believe we could tackle that later.

Luap99:
Please exit with an exit code > 0 otherwise podman will blow up.
https://doc.rust-lang.org/std/process/fn.exit.html

lsm5:
> Please exit with an exit code > 0 otherwise podman will blow up. https://doc.rust-lang.org/std/process/fn.exit.html

@Luap99 @flouthoc PTAL at the latest change.

src/main.rs
```diff
@@ -25,14 +24,32 @@ enum SubCommand {
...
-        SubCommand::Setup(setup) => setup.exec(file),
-        SubCommand::Teardown(teardown) => teardown.exec(file),
+        SubCommand::Setup(setup) => {
+            if let Err(err) = setup.exec(file) {
+                let er = netavark::error::NetavarkError {
+                    error: format!("{}", err),
+                    errno: 1,
+                };
+                println!("{}", er.to_string());
+                quit::with_code(1);
```
flouthoc, line 40:
```suggestion
                quit::with_code(er.errno);
```

src/main.rs
```diff
@@ -25,14 +24,32 @@ enum SubCommand {
...
+            }
+        }
+        SubCommand::Teardown(teardown) => {
+            if let Err(err) = teardown.exec(file) {
+                let er = netavark::error::NetavarkError {
+                    error: format!("{}", err),
+                    errno: 1,
+                };
+                println!("{}", er.to_string());
+                quit::with_code(1);
```
flouthoc, line 50:
```suggestion
                quit::with_code(er.errno);
```

flouthoc (approved):
LGTM. Just nit.

src/main.rs
```diff
@@ -25,14 +24,32 @@ enum SubCommand {
...
+            }
+        }
+        SubCommand::Teardown(teardown) => {
+            if let Err(err) = teardown.exec(file) {
+                let er = netavark::error::NetavarkError {
+                    error: format!("{}", err),
+                    errno: 1,
+                };
+                println!("{}", er.to_string());
+                quit::with_code(er.errno);
```
Luap99, line 50: Is there any point in using a new dependency for this. Can you please just use `std::process::exit()` directly.

lsm5: sure thing, just that I saw a few recommendations online for quit. I'll change this in a bit. Thanks

lsm5: done, ptal

PR #53, opened 2021-11-01
teardown: implement teardown for removing container interfaces or container `veth`.

flouthoc:
Implements teardown which is responsible for removing container
interfaces and performing cleanups.

Usage
```console
netavark -f <config> teardown <networknamespace>
```

### DIFF
--- a/src/commands/teardown.rs
+++ b/src/commands/teardown.rs
@@ -22,10 +22,16 @@ impl Teardown {
         network::validation::ns_checks(&self.network_namespace_path);
 
         debug!("{:?}", "Tearing down..");
-        let _network_options = match network::types::NetworkOptions::load(&input_file) {
+        let network_options = match network::types::NetworkOptions::load(&input_file) {
             Ok(opts) => opts,
             Err(e) => panic!("{}", e),
         };
+
+        //Remove container interfaces
+        network::core::Core::remove_interface_per_podman_network(
+            &network_options,
+            &self.network_namespace_path,
+        )?;
 
         debug!("{:?}", "Teardown complete");
         Ok(())
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -184,4 +184,74 @@ impl Core {
             }
         }
     }
+
+    pub fn remove_interface_per_podman_network(
+        network_opts: &types::NetworkOptions,
+        netns: &str,
+    ) -> Result<(), std::io::Error> {
+        for (net_name, network) in &network_opts.network_info {
+            // get network name
+            let network_name: String = net_name.to_owned();
+            // get PerNetworkOptions for this network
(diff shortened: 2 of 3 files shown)


flouthoc:
@mheon @Luap99 @baude @lsm5 Adding utility functions which will be used by global teardown. Not being used anywhere as of now.

We would also need functions to clean IpTables `FORWARD_RULES` and `chains` if any.

flouthoc:
I think ordering of cleanup is imp afaik not sure.

mheon:
Our existing cleanup functions will exclusively remove firewall port-forwarding rules (not yet implemented) and the container's veth pair (either interface works, removing one removes the pair). It will not tear down the bridge, or the firewall rules for the bridge. We're going to implement a separate CLI command to do that, which `podman network rm` will invoke.

Please integrate these into the Teardown command as such, there should be no ordering concerns given your code is the only thing that will be running.

flouthoc:
@mheon makes sense. let me rig this into cli

flouthoc:
@Luap99 @mheon @baude @lsm5 This is done.

Working example
```console
# Create links using `setup`
sudo ./bin/netavark -f src/test/config/setupopts2.test.json setup /var/run/netns/netnstest

# Remove links using `teardown`
sudo ./bin/netavark -f src/test/config/setupopts2.test.json teardown /var/run/netns/netnstest
```

flouthoc:
@baude this should also help with easier cleanups.

Luap99 (commented):
LGTM

flouthoc:
@Luap99 @mheon Do we also need to clean anything other than interfaces. I am assuming ip tables but not entirely sure.

mheon:
We only need to do so on complete removal of the network which will be another CLI command.

Luap99:
I still do not think we want a new cli command for this, we should always cleanup if there are no more containers connect to the network.

mheon:
Clean up iptables, sure, I can see that. Removing the bridge entirely might be too far, and I think we want Netavark to do that on network removal.

Luap99:
Why? I know that people are already complaining about this with CNI. Removing the bridge is safe because we have a lock in podman. We can query netlink to only list interfaces connected to the specific bridge, if there are none we do full cleanup.

mheon:
@Luap99 Worried about people potentially doing custom network configuration - I recall a few issues on the Podman repo about doing this in postrun hooks. We'd basically be wiping it out every time the bridge went away.

Luap99:
My main concern is that while podman does know it, it is very expensive to query. We would need to list all running containers and check the networks for each one. With netlink we could use a single call with IFLA_MASTER. This should be much faster.

flouthoc:
Fair point.

flouthoc:
I also agree that `netlink` can do this much faster as compared to iterating over data structures.

PR #57, opened 2021-11-03
core: remove dangling interfaces from host in case of error

flouthoc:
When invoking `setup` with same config twice we get an error that interface
already exists which is expected. But even then we leave two dandling interface
on host.

* First the one we created for host.
* Second which we were unable to move to container namespace.

We must clean dangling interfaces in case of error. So teardown is not a
mess.

This solves the issue when two consecutive setup commands for same configurations
is invoked.

**How to reproduces the problem solved which is being solved in this PR ?**

```console
sudo ./bin/netavark -f src/test/config/setupopts2.test.json setup /var/run/netns/netnstest
sudo ./bin/netavark -f src/test/config/setupopts2.test.json setup /var/run/netns/netnstest
```

Second setup will fail as it is expected but would also leave dangling
veth pairs on host. Which will make future teardown difficult.

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -1,5 +1,6 @@
 use crate::network::{core_utils, types};
 use log::debug;
+use log::warn;
 use nix::sched;
 use rand::Rng;
 use std::fs::File;
@@ -119,10 +120,17 @@ impl Core {
         ) {
             Ok(_) => (),
             Err(err) => {
+                // it seems something went wrong
+                // we must not leave dangling interfaces
+                // otherwise cleanup would become mess
+                // try removing leaking interfaces from host
+                if let Err(er) = core_utils::CoreUtils::remove_interface(&host_veth_name) {
+                    warn!("Failed while cleaning up interfaces: {}", er);
+                }
                 return Err(std::io::Error::new(
                     std::io::ErrorKind::Other,
                     format!("Error while configuring network interface {}:", err),
-                ))
+                ));
             }
         };
 


flouthoc:
@Luap99 @mheon @baude @lsm5 This should fix the issues which we are facing with dangling interfaces.

flouthoc:
Once this merged and all leaking interfaces are cleaned up from host manually. There should be no issues ever of any dangling interfaces.

src/network/core.rs
```diff
@@ -119,10 +119,15 @@ impl Core {
         ) {
             Ok(_) => (),
             Err(err) => {
+                // it seems something went wrong
+                // we must not leave dangling interfaces
+                // otherwise cleanup would become mess
+                // try removing leaking interfaces from host
+                let _ = core_utils::CoreUtils::remove_interface(&host_veth_name);
```
Luap99, line 126: Can you log the error if there is one, maybe as warning.

flouthoc: sure

src/network/core.rs
```diff
@@ -119,10 +120,17 @@ impl Core {
         ) {
             Ok(_) => (),
             Err(err) => {
+                // it seems something went wrong
+                // we must not leave dangling interfaces
+                // otherwise cleanup would become mess
+                // try removing leaking interfaces from host
+                if let Err(er) = core_utils::CoreUtils::remove_interface(&host_veth_name) {
```
baude, line 127: this is sort of a blind remove? should we be logging that?

flouthoc: Its a blind remove cause we created a veth pair on host but whole call failed so we must remove it.
We are loggin if we failed to remove this.

flouthoc: Only problem is when we are unable to remove this and we are warning when we fail to do it.

PR #59, opened 2021-11-03
Return response to caller (Podman)

baude:
Return a map of StatusBlock describing the network set ups.


flouthoc:
LGTM

flouthoc (approved):

PR #60, opened 2021-11-03
core: turn up  `lo` adapter in ctr namespace during setup

flouthoc:
We must turn up loopback while `setup` cause podman will not do that for
us.

### DIFF
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -192,6 +192,27 @@ impl CoreUtils {
         Ok(())
     }
 
+    #[tokio::main]
+    pub async fn turn_up_interface(ifname: &str) -> Result<(), Error> {
+        let (connection, handle, _) = match rtnetlink::new_connection() {
+            Ok((conn, handle, messages)) => (conn, handle, messages),
+            Err(err) => {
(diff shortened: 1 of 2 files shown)


flouthoc:
@Luap99 PTAL

mheon:
LGTM

src/network/core_utils.rs
```diff
@@ -192,6 +192,27 @@ impl CoreUtils {
...
+            Ok((conn, handle, messages)) => (conn, handle, messages),
+            Err(err) => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("failed to connect: {}", err),
+                ))
+            }
+        };
+
+        tokio::spawn(connection);
```
Luap99, line 207: Not for this PR but we should stop doing it. We should only create one netlink connection and use this for all calls. Right now we have way to many unneeded clone syscalls.

Luap99 (commented):
LGTM

src/network/core_utils.rs
flouthoc: Sure we could pass it along the function. I'll refactor this later after making sure nothing breaks.

PR #64, opened 2021-11-04
Refactor bridge setup

Luap99:
We should setup the bridge and firewall rules in the same loop. We also
have to check the given network driver. For now we only support bridge
but we can easily add more in the future.

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -1,8 +1,10 @@
 //! Configures the given network namespace with provided specs
 use crate::firewall;
 use crate::network;
+use crate::network::types;
 use clap::{self, Clap};
 use log::debug;
+use std::collections::HashMap;
 use std::error::Error;
 use sysctl::Sysctl;
 use sysctl::SysctlError;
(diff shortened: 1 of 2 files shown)


Luap99:
@baude @flouthoc @mheon PTAL this is a just small cleanup

src/commands/setup.rs
```diff
@@ -46,18 +48,43 @@ impl Setup {
...
         for (net_name, network) in network_options.network_info {
-            debug!("Setting up network {} firewall rules", net_name);
+            debug!(
+                "Setting up network {} with driver {}",
+                net_name, network.driver
+            );
 
-            // Setup basic firewall rules for each network.
-            firewall_driver.setup_network(network)?;
+            match network.driver.as_str() {
```
mheon, line 60: I'd love to get this behind an interface eventually similar to what I'm doing with the firewall driver

mheon:
LGTM. I have ideas for future cleanups, but this looks like a good improvement

flouthoc:
LGTM and I'd agree we can move all the logic from `src/command/setup.rs` into a new interface maybe `src/podman/setup.rs`.

PR #65, opened 2021-11-04
Fix POSTROUTING rule exists check

Luap99:
There was a typo in the iptables rule exists check, we have to read the NAT table.


flouthoc:
LGTM

PR #66, opened 2021-11-05
netlink add ecmp route

Luap99:
The default route will be added for each connected network, so when you
would try to use two networks it would fail because the default route
was already added. While we could ignore eexist error this is not what
we want.
If we only have one route and disconnect the network with the default
route we will loose internet connectivity. To best way is to have each
network create the default route. This can be done by not setting the
NLM_F_EXCL flag for the netlink request.

Also see this CNI bridge plugin PR which does the same there: https://github.com/containernetworking/plugins/pull/615

### DIFF
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -1,14 +1,21 @@
 use futures::stream::TryStreamExt;
+use futures::StreamExt;
 use ipnetwork::IpNetwork;
 use ipnetwork::Ipv4Network;
 use ipnetwork::Ipv6Network;
 use libc;
 use nix::sched;
 use rtnetlink;
+use rtnetlink::packet::constants::*;
 use rtnetlink::packet::rtnl::link::nlas::Nla;
+use rtnetlink::packet::NetlinkPayload;
+use rtnetlink::packet::RouteMessage;
 use std::fmt::Write;
 use std::fs::File;
 use std::io::Error;
+use std::net::IpAddr;
+use std::net::Ipv4Addr;
+use std::net::Ipv6Addr;
 use std::os::unix::prelude::*;
 use std::process;
 use std::thread;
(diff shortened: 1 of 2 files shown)


Luap99:
@mheon @baude @flouthoc PTAL

src/network/core_utils.rs
```diff
@@ -1,14 +1,21 @@
 use futures::stream::TryStreamExt;
+use futures::StreamExt;
```
flouthoc, line 2: are we using this import anywhere. I am not able to see in the diff.

src/network/core_utils.rs
```diff
@@ -133,45 +140,67 @@ impl CoreUtils {
...
-            Err(err) => Err(std::io::Error::new(
-                std::io::ErrorKind::Other,
-                format!("failed to add route: {}", err),
-            )),
-        }
+            .gateway(*gateway)
+            .message_mut()
+            .to_owned();
+
+        CoreUtils::execute_route_msg(handle, msg).await
```
flouthoc, line 154: I think we need to propagate the error back upwards

src/network/core_utils.rs
```diff
@@ -133,45 +140,67 @@ impl CoreUtils {
...
-            Ok(_) => Ok(()),
-            Err(err) => Err(std::io::Error::new(
-                std::io::ErrorKind::Other,
-                format!("failed to add route: {}", err),
-            )),
+            .gateway(*gateway)
+            .message_mut()
+            .to_owned();
+
+        CoreUtils::execute_route_msg(handle, msg).await
```
flouthoc, line 171: We need to propagate error upwards.

flouthoc:
LGTM. Just couple of nits.

src/network/core_utils.rs
Luap99: Yes I it need for `let mut response = match handle.clone().request(req)`, do not ask me why.
Also the linter fails with unused imports so we should be safe.

src/network/core_utils.rs
Luap99: It does this, this is just the awkward rust syntax with no semicolon == return

src/network/core_utils.rs
flouthoc: ah i see call to request returns a future `Stream`

src/network/core_utils.rs
flouthoc: resolved in above comment.

flouthoc:
LGTM

PR #74, opened 2021-11-15
netavark: fix error wrapping and formatting

flouthoc:
Fix error wrapping and formatting. Remove unnecessary `{:?}` debug which escapes and prints entire `struct` for debugging.
formatter from errors.

Tidy up errors.

Final `Error: Custom` gets removed after https://github.com/containers/netavark/pull/47

Before:
```console
Error: Custom { kind: Other, error: "failed configure bridge and veth interface: Custom { kind: Other, error: \"Error while configuring network interface from container namespace: Custom { kind: Other, error: \\\"interface ethsomething0 already exists on container namespace\\\" }:\" }" }
```
### After

* with https://github.com/containers/netavark/pull/47
```console
{"error":"failed configure bridge and veth interface: failed while configuring network interface from container namespace: interface ethsomething0 already exists on container namespace:","errno":1}
```
* without #47 
```console
Error: Custom { kind: Other, error: "failed configure bridge and veth interface: failed while configuring network interface from container namespace: interface ethsomething0 already exists on container namespace:" }
```

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -108,7 +108,7 @@ impl Core {
             Err(err) => {
                 return Err(std::io::Error::new(
                     std::io::ErrorKind::Other,
-                    format!("failed configure bridge and veth interface: {:?}", err),
+                    format!("failed configure bridge and veth interface: {}", err),
                 ))
             }
         };
@@ -141,7 +141,7 @@ impl Core {
             Err(err) => {
                 return Err(std::io::Error::new(
                     std::io::ErrorKind::Other,
-                    format!("Error while configuring network interface {}:", err),
+                    format!("failed while configuring network interface {}:", err),
                 ))
             }
         };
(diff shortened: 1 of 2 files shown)


lsm5 (approved):
LGTM, thanks!

src/network/core.rs
```diff
@@ -215,12 +215,12 @@ impl Core {
                         Err(err) => Err(err),
                     }
                 });
-                match handle.join() {
-                    Ok(interface_address) => interface_address,
+                match handle.join().unwrap() {
```
Luap99, line 218: Why use `unwrap()`. We want to handle errors properly and return them.

src/network/core.rs
```diff
@@ -281,10 +281,10 @@ impl Core {
 
                     Ok(())
                 });
-                if let Err(err) = handle.join() {
+                if let Err(err) = handle.join().unwrap() {
```
Luap99, line 284: same

src/network/core.rs
flouthoc: @Luap99 If `join` fails at thread spawning level process will `panic` with appropriate message this is a special case. However the error returned by our task needs to  processed correctly which we are doing in `if` block.

Doc for `join` : https://doc.rust-lang.org/std/thread/struct.JoinHandle.html
```markdown
pub fn join(self) -> Result<T>

Waits for the associated thread to finish.

This function will return immediately if the associated thread has already finished.

In terms of atomic memory orderings, the completion of the associated thread synchronizes with this function returning. In other words, all operations performed by that thread happen before all operations that happen after join returns.

If the associated thread panics, Err is returned with the parameter given to panic!.
Panics

This function may panic on some platforms if a thread attempts to join itself or otherwise may create a deadlock with joining threads.
```

flouthoc: nvm resetting this change for readability purpose. `{:?}` is needed for `join` since it returns `Boxed Any` which does not implements `display` but it will not affect wrapping the error.

flouthoc:
@Luap99 PTAL

PR #76, opened 2021-11-16
core: add support for setting up `macvlan` driver from config

flouthoc:
Following PR adds extends `setup` command to allow configuring `macvlan`
if specified in config as a `driver`.

When merged following should work with `macvlan` config.
```console
sudo ./bin/netavark -f <config_with_macvlan> setup /var/run/netns/netnstest
```

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -136,7 +136,7 @@ impl Setup {
                 }
                 "macvlan" => {
                     let per_network_opts =
-                        network_options.networks.get(&net_name).ok_or_else(|| {
+                        network_options.networks.get(net_name).ok_or_else(|| {
                             std::io::Error::new(
                                 std::io::ErrorKind::Other,
                                 format!("network options for network {} not found", net_name),
@@ -145,13 +145,10 @@ impl Setup {
                     //Configure Bridge and veth_pairs
                     let status_block = network::core::Core::macvlan_per_podman_network(
                         per_network_opts,
-                        &network,
+                        network,
                         &self.network_namespace_path,
                     )?;
-                    response.insert(net_name, status_block);
-                    // Setup basic firewall rules for each network.
-                    firewall_driver.setup_network(network)?;
-                    // TODO: Set up port forwarding. How? What network do we point to?
+                    response.insert(net_name.to_owned(), status_block);
                 }
                 // unknown driver
                 _ => {
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -134,6 +134,25 @@ impl Setup {
                         }
                     }
                 }
+                "macvlan" => {
+                    let per_network_opts =
+                        network_options.networks.get(&net_name).ok_or_else(|| {
+                            std::io::Error::new(
+                                std::io::ErrorKind::Other,
+                                format!("network options for network {} not found", net_name),
+                            )
+                        })?;
+                    //Configure Bridge and veth_pairs
+                    let status_block = network::core::Core::macvlan_per_podman_network(
+                        per_network_opts,
+                        &network,
+                        &self.network_namespace_path,
+                    )?;
+                    response.insert(net_name, status_block);
+                    // Setup basic firewall rules for each network.
+                    firewall_driver.setup_network(network)?;
+                    // TODO: Set up port forwarding. How? What network do we point to?
+                }
                 // unknown driver
                 _ => {
                     return Err(std::io::Error::new(
--- a/src/commands/teardown.rs
+++ b/src/commands/teardown.rs
@@ -62,6 +62,21 @@ impl Teardown {
 
                     // TODO teardown firewall if no interfaces connected to bridge!
                 }
+                "macvlan" => {
+                    let per_network_opts =
+                        network_options.networks.get(&net_name).ok_or_else(|| {
+                            std::io::Error::new(
+                                std::io::ErrorKind::Other,
+                                format!("network options for network {} not found", net_name),
+                            )
+                        })?;
+                    //Remove container interfaces
+                    network::core::Core::remove_interface_per_podman_network(
+                        per_network_opts,
+                        &network,
+                        &self.network_namespace_path,
+                    )?;
+                }
                 // unknown driver
                 _ => {
                     return Err(std::io::Error::new(
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -1,5 +1,5 @@
 use crate::network::types::NetAddress;
-use crate::network::{core_utils, types};
+use crate::network::{constants, core_utils, types};
 use ipnet;
 use log::debug;
 use log::warn;
@@ -242,9 +242,21 @@ impl Core {
         let mut response = types::StatusBlock {
             interfaces: Some(HashMap::new()),
         };
-        // Does config have a macvlan mode ? I think not
-        // Important !! Hardcode MACVLAN_MODE to bridge
-        let macvlan_mode: u32 = 4u32;
+        // Default MACVLAN_MODE to bridge or get from driver options
+        let mut macvlan_mode: u32 = constants::MACVLAN_MODE_BRIDGE;
+        if let Some(options_map) = network.options.as_ref() {
+            if let Some(mode) = options_map.get("mode") {
+                match core_utils::CoreUtils::get_macvlan_mode_from_string(mode) {
+                    Ok(mode) => {
+                        macvlan_mode = mode;
+                    }
+                    Err(err) => {
+                        return Err(err);
+                    }
+                }
+            }
+        }
+
         // get master interface name
         let mut master_ifname: String = match network.network_interface.as_ref() {
             None => {
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -232,6 +232,173 @@ impl Core {
         }
     }
 
+    pub fn macvlan_per_podman_network(
+        per_network_opts: &types::PerNetworkOptions,
+        network: &types::Network,
+        netns: &str,
+    ) -> Result<types::StatusBlock, std::io::Error> {
+        //  StatusBlock response
+        let mut response = types::StatusBlock {
+            dns_search_domains: Some(Vec::new()),
+            dns_server_ips: Some(Vec::new()),
+            interfaces: Some(HashMap::new()),
+        };
+        // Does config have a macvlan mode ? I think not
+        // Important !! Hardcode MACVLAN_MODE to bridge
+        let macvlan_mode: u32 = 4u32;
+        // get master interface name
+        let master_ifname: String = network.network_interface.as_ref().unwrap().to_owned();
+        // static ip vector
+        let mut address_vector = Vec::new();
+        // network addresses for response
+        let mut response_net_addresses: Vec<NetAddress> = Vec::new();
+        // interfaces map, but we only ever expect one, for response
+        let mut interfaces: HashMap<String, types::NetInterface> = HashMap::new();
+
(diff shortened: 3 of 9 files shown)


flouthoc:
@Luap99 I have assumed that input config is same as `bridge` and some fields like `gateway` will be unused. But if you could share any actual config for `macvlan` i would like to validate against that.

Luap99 (commented):
I think you have to add the default route, could you also add teardown. This should be very simple since you only have to delete the interface in the container ns.

src/network/core.rs
```diff
@@ -234,6 +234,173 @@ impl Core {
...
+        let mut response = types::StatusBlock {
+            dns_search_domains: Some(Vec::new()),
+            dns_server_ips: Some(Vec::new()),
+            interfaces: Some(HashMap::new()),
+        };
+        // Does config have a macvlan mode ? I think not
+        // Important !! Hardcode MACVLAN_MODE to bridge
+        let macvlan_mode: u32 = 4u32;
+        // get master interface name
+        let master_ifname: String = network.network_interface.as_ref().unwrap().to_owned();
```
Luap99, line 252: I forgot to mention, network_interface is optional because we do not require it for macvlan. If it is empty you should use the default route interface. Here is the cni code https://github.com/containernetworking/plugins/blob/f1f128e3c9220634d3f26b96950271f87c34e424/plugins/main/macvlan/macvlan.go#L63

flouthoc: @Luap99 thanks added a change in new commit.

flouthoc:
@Luap99 teardown for `macvlan` is almost same as `bridge` we just need to remove `macvlan` interfaces and delete function does not cares if its a `macvlan interface or a veth`.

src/commands/setup.rs
```diff
@@ -70,7 +70,7 @@ impl Setup {
             );
 
             match network.driver.as_str() {
-                "bridge" => {
+                "test" => {
```
Luap99, line 73: You have to remove this

src/commands/setup.rs
```diff
@@ -134,6 +134,70 @@ impl Setup {
...
+                            firewall_driver.setup_port_forward(
+                                &network_options.container_id,
+                                i,
+                                container_ips[0],
+                                networks[0].subnet,
+                                net_name,
+                                &id_network_hash.as_str()[0..MAX_HASH_SIZE],
+                            )?;
+                        }
+                    }
```
Luap99, line 199: macvlan does not need firewall setuop, please remove this

src/commands/teardown.rs
```diff
@@ -62,6 +62,25 @@ impl Teardown {
...
+                    //Remove container interfaces
+                    network::core::Core::remove_interface_per_podman_network(
+                        per_network_opts,
+                        &network,
+                        &self.network_namespace_path,
+                    )?;
+                    // Teardown basic firewall port forwarding
+                    // firewall_driver.teardown_port_forward(network)?;
+
+                    // TODO teardown firewall if no interfaces connected to bridge!
```
Luap99, line 82: same, no firewall stuff needed

src/network/core_utils.rs
```diff
@@ -34,6 +34,51 @@ impl CoreUtils {
...
+                                        }
+                                    }
+                                }
+                                Err(_) => continue,
+                                Ok(None) => continue,
+                            };
+                        }
+                    }
+                }
+                Err(_) => continue,
```
baude, line 71: should this be logged?

src/network/core.rs
```diff
@@ -232,6 +232,180 @@ impl Core {
...
+        netns: &str,
+    ) -> Result<types::StatusBlock, std::io::Error> {
+        //  StatusBlock response
+        //  StatusBlock response
+        let mut response = types::StatusBlock {
+            interfaces: Some(HashMap::new()),
+        };
+        // Does config have a macvlan mode ? I think not
+        // Important !! Hardcode MACVLAN_MODE to bridge
+        let macvlan_mode: u32 = 4u32;
```
Luap99, line 247: We support setting macvlan modes, see https://github.com/containers/podman/blob/main/docs/source/markdown/podman-network-create.1.md#--optoption--o

You have to read the `mode` field from the network options and use the correct mode. If it is unset default to bridge.

src/network/core.rs
```diff
@@ -232,6 +232,180 @@ impl Core {
...
+        // get master interface name
+        let mut master_ifname: String = network.network_interface.as_ref().unwrap().to_owned();
```
Luap99, line 249: You cannot use unwrap here. This can be empty. Your code below must handle this if it is None.

src/commands/setup.rs
flouthoc: removed in latest commit

src/network/core.rs
flouthoc: Yes just checking for `empty` string in line after this and getting default `Route` interface if this is empty.

src/network/core.rs
flouthoc: ah sure i'll add support for this probably `enum` map was not aware that we were supporting all types.

src/network/core_utils.rs
flouthoc: @baude this is not needed we are just looking for any `rout-able` interface. I did this intentionally.

src/network/core.rs
flouthoc: ackd. I'll rectify this.

src/network/core.rs
flouthoc: @Luap99 added support for all available `macvlan` modes in latest commit.

src/network/core_utils.rs
```diff
@@ -34,6 +35,65 @@ impl CoreUtils {
         final_slice.join(":")
     }
 
+    pub fn get_macvlan_mode_from_string(mode: &str) -> u32 {
+        // Replace to constant from library once this gets merged.
+        // TODO: use actual constants after https://github.com/little-dude/netlink/pull/200
```
Luap99, line 40: PR is merged, you can change it.

src/network/core_utils.rs
```diff
@@ -34,6 +35,65 @@ impl CoreUtils {
...
+        match mode {
+            "bridge" => constants::MACVLAN_MODE_BRIDGE,
+            "private" => constants::MACVLAN_MODE_PRIVATE,
+            "vepa" => constants::MACVLAN_MODE_VEPA,
+            "passthru" => constants::MACVLAN_MODE_PASSTHRU,
+            "source" => constants::MACVLAN_MODE_SOURCE,
+            // default to bridge
+            _ => constants::MACVLAN_MODE_BRIDGE,
```
Luap99, line 48: This should error. We should only default to bridge if the mode key does not exists in the network options.

src/network/core.rs
```diff
@@ -232,6 +232,196 @@ impl Core {
...
+        if let Some(options_map) = network.options.as_ref() {
+            if let Some(mode) = options_map.get("mode") {
+                macvlan_mode = core_utils::CoreUtils::get_macvlan_mode_from_string(mode);
+            }
+        }
+
+        // get master interface name
+        let mut master_ifname: String = "".to_string();
+
+        match network.network_interface.as_ref() {
```
Luap99, line 256: I think this can this be changed to `let master_ifname = match ...`

src/network/core.rs
```diff
@@ -232,6 +232,196 @@ impl Core {
...
+                    master_ifname = ifname;
+                }
+            }
+            Some(interface) => {
+                master_ifname = interface.to_owned();
+            }
+        }
+
+        // no interface name provided
+        if master_ifname.trim().is_empty() {
```
Luap99, line 268: No need fro trim. If there are extra spaces it is a user error, we should not handle this.

src/network/core.rs
```diff
@@ -232,6 +232,196 @@ impl Core {
...
+            if let Ok(ifname) = core_utils::CoreUtils::get_default_route_interface() {
+                master_ifname = ifname;
+            }
```
Luap99, line 271: If the master interface is still empty we should error here

src/network/core_utils.rs
flouthoc: @Luap99 A version needs to be cut for `netlink-packet-core` or we would need to include it in our `Cargo.toml`. There is a cascading dependency of versioned packages preventing me to use `master`.

Luap99: ok lets keep it for now

flouthoc: I'll try finding a better way to point all dependent package to `master` in a new PR maybe if that works.

Luap99 (commented):
LGTM

Luap99:
@mheon @baude  PTAL

flouthoc:
@mheon Luckily i don't need a rebase PTAL whenever you can.

src/network/core.rs
```diff
@@ -232,6 +232,210 @@ impl Core {
...
+                        "unable to find any valid master interface for macvlan".to_string(),
+                    ));
+                }
+            }
+            Some(interface) => interface.to_owned(),
+        };
+
+        // user could provide empty value as well so handle that
+        if master_ifname.is_empty() {
+            if let Ok(ifname) = core_utils::CoreUtils::get_default_route_interface() {
```
mheon, line 277: Why an if-let here? It seems like we should be returning an error if we can't get a default interface name, because ifname will still be empty at the end of this block and we'll just proceed on and fail later.

src/network/core.rs
```diff
@@ -232,6 +232,210 @@ impl Core {
...
+                    warn!("failed while cleaning up interfaces: {}", er);
+                }
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("failed while configuring macvlan {}:", err),
+                ));
+            }
+        };
+
+        match File::open(&netns) {
```
mheon, line 378: This really needs to turn into a function or macro, I see it a lot in netavark code

src/network/core_utils.rs
```diff
@@ -34,6 +35,68 @@ impl CoreUtils {
         final_slice.join(":")
     }
 
+    pub fn get_macvlan_mode_from_string(mode: &str) -> Result<u32, std::io::Error> {
+        // Replace to constant from library once this gets merged.
+        // TODO: use actual constants after https://github.com/little-dude/netlink/pull/200
+        match mode {
```
mheon, line 41: Can we normalize the input to lowercase to avoid simple user errors?

Luap99: IMO The netavark format should be as strict as possible. I would prefer to normalize in podman if needed.

src/network/core.rs
flouthoc: @mheon Yes that is handled below in  `else` block if result is not `Ok` and another block below handles the case for empty check.

src/network/core.rs
flouthoc: I agree there are multiple blocks I'll fix this in connection refactor PR.

src/network/core_utils.rs
mheon: Ack, I'm OK with this

src/network/core.rs
mheon: What else block? I don't see one

flouthoc: @mheon added a `else` my bad got confused with the one above.

flouthoc: https://github.com/containers/netavark/pull/76/files#diff-dadc3c975feb7c242a4ea465a9dcab3a98600a15ce4c5632ad5a671c0e3d209aR285

PR #80, opened 2021-11-17
Enabled portfw'ing for iptables

baude:
when using the iptables firewall driver, we can no forward ports from
the container host to the container.

i.e. curl http://localhost:8080

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -1,6 +1,7 @@
 //! Configures the given network namespace with provided specs
 use crate::error::NetavarkError;
 use crate::firewall;
+use crate::firewall::iptables::MAX_HASH_SIZE;
 use crate::network;
 use crate::network::types;
 use clap::{self, Clap};
@@ -62,7 +63,7 @@ impl Setup {
         let mut response: HashMap<String, types::StatusBlock> = HashMap::new();
 
         // Perform per-network setup
-        for (net_name, network) in network_options.network_info {
+        for (net_name, network) in network_options.network_info.iter() {
             debug!(
                 "Setting up network {} with driver {}",
                 net_name, network.driver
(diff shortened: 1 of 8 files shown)


src/commands/setup.rs
```diff
@@ -80,13 +81,57 @@ impl Setup {
...
+                        Some(i) => {
+                            let localnet_path = format!("net.ipv4.conf.{}.route_localnet", i);
+                            let sysctl_localnet = get_sysctl_value(localnet_path.as_str())?;
+                            if sysctl_localnet != *"1" {
+                                set_sysctl_value(localnet_path.as_str(), "1")?;
+                            }
+                        }
+                    }
+                    firewall_driver.setup_network(network.clone(), id_network_hash.clone())?;
+                    let port_bindings = network_options.port_mappings.clone();
```
flouthoc, line 108: nit: Not a blocker: Would it be possible to move this inside `firewall` create as a new function and call and single function in setup.

flouthoc (commented):
@baude Could you share a tryout example as well. It would be helpful to test this.

src/commands/setup.rs
```diff
@@ -62,7 +63,7 @@ impl Setup {
         let mut response: HashMap<String, types::StatusBlock> = HashMap::new();
 
         // Perform per-network setup
-        for (net_name, network) in network_options.network_info {
+        for (net_name, network) in network_options.network_info.iter() {
```
flouthoc, line 66: Is `iter` needed here since i cant the change in `network_info`

src/commands/setup.rs
```diff
@@ -80,13 +81,57 @@ impl Setup {
                     //Configure Bridge and veth_pairs
                     let status_block = network::core::Core::bridge_per_podman_network(
                         per_network_opts,
-                        &network,
+                        network,
```
flouthoc, line 84: Same there is a type change i guess because of iter.

src/commands/setup.rs
baude: yes, it is needed

src/commands/setup.rs
baude: yes

src/commands/setup.rs
baude: i think this is a good idea, just want to confirm with @mheon

flouthoc:
LGTM. @Luap99 @mheon PTAL . I haven't tried running yet but code looks good.

PR #82, opened 2021-11-18
Initial checkin for teardown iptables

baude:
When a container is cleaned up, podman will call netavark to teardown
the network information.  We need to clean up any container specific
iptables rules.

Once we have a fn to detect if the container is the last on the network,
then we can also teardown network specific iptables rules.

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -3,13 +3,12 @@ use crate::error::NetavarkError;
 use crate::firewall;
 use crate::firewall::iptables::MAX_HASH_SIZE;
 use crate::network;
-use crate::network::types;
+use crate::network::core_utils::CoreUtils;
+use crate::network::{core_utils, types};
 use clap::{self, Clap};
 use log::debug;
 use std::collections::HashMap;
 use std::error::Error;
-use sysctl::Sysctl;
-use sysctl::SysctlError;
 
 const IPV4_FORWARD: &str = "net.ipv4.ip_forward";
 
@@ -35,7 +34,6 @@ impl Setup {
                 bail!("invalid namespace path: {}", e);
             }
         }
-
         debug!("{:?}", "Setting up...");
 
         let network_options = match network::types::NetworkOptions::load(&input_file) {
@@ -54,11 +52,8 @@ impl Setup {
         };
 
         // Sysctl setup
-        // set ip forwarding to 1 if not already
-        let sysctl_ipv4 = get_sysctl_value(IPV4_FORWARD)?;
-        if sysctl_ipv4 != *"1" {
-            set_sysctl_value(IPV4_FORWARD, "1")?;
-        }
+        // set ip forwarding to 1
+        core_utils::CoreUtils::apply_sysctl_value(IPV4_FORWARD, "1")?;
 
         let mut response: HashMap<String, types::StatusBlock> = HashMap::new();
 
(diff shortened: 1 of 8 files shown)


src/commands/setup.rs
```diff
@@ -55,9 +53,9 @@ impl Setup {
 
         // Sysctl setup
         // set ip forwarding to 1 if not already
-        let sysctl_ipv4 = get_sysctl_value(IPV4_FORWARD)?;
+        let sysctl_ipv4 = core_utils::CoreUtils::lookup_sysctl_value(IPV4_FORWARD)?;
         if sysctl_ipv4 != *"1" {
-            set_sysctl_value(IPV4_FORWARD, "1")?;
+            core_utils::CoreUtils::apply_sysctl_value(IPV4_FORWARD, "1")?;
```
Luap99, line 58: Just wondering, what is the point of reading this value when we could just always set to 1.

src/firewall/firewalld.rs
```diff
@@ -67,21 +66,24 @@ impl firewall::FirewallDriver for FirewallD {
...
 
     fn teardown_port_forward(
         &self,
+        _network: Network,
         _container_id: &str,
         _port_mappings: Vec<types::PortMapping>,
-        _container_ip: &str,
+        _network_name: &str,
+        _id_network_hash: &str,
+        _options: &PerNetworkOptions,
```
Luap99, line 86: Please remove the leading underscore from the variables so that the compiler will complain if they are unused.

src/firewall/iptables.rs
```diff
@@ -109,13 +109,34 @@ impl firewall::FirewallDriver for IptablesDriver {
...
+        // through localhost to containers
+        let network_interface = network.network_interface;
+        match network_interface {
+            None => {}
+            Some(i) => {
+                let localnet_path = format!("net.ipv4.conf.{}.route_localnet", i);
+                let sysctl_localnet = CoreUtils::lookup_sysctl_value(localnet_path.as_str())?;
+                if sysctl_localnet != *"1" {
+                    CoreUtils::apply_sysctl_value(localnet_path.as_str(), "1")?;
+                }
```
Luap99, line 129: same here, what is the point of reading this value when we could just always set to 1.

src/network/core_utils.rs
```diff
@@ -887,4 +889,18 @@ impl CoreUtils {
         let response = &hash_string[0..length];
         response.to_string()
     }
+
+    // get a sysctl value by the value's namespace
+    pub fn lookup_sysctl_value(ns_value: &str) -> Result<String, SysctlError> {
```
flouthoc, line 894: I thought all the `sysctl` utils were in firewall crate. But this is also fine.

src/firewall/iptables.rs
```diff
@@ -285,6 +395,21 @@ fn chain_exists(driver: &IPTables, table: &str, chain: &str) -> Result<bool, Box
     serde::__private::Result::Ok(false)
 }
 
+fn remove_if_exists(
```
flouthoc, line 398: function name could be more explicit `remove_rule_if_exists` I think would help with readability.

src/firewall/iptables.rs
```diff
@@ -233,11 +255,99 @@ impl firewall::FirewallDriver for IptablesDriver {
...
+        );
+        let network_chain_name = CONTAINER_CHAIN.to_owned() + id_network_hash;
+        let container_ip = container_ips[0];
+        // First delete any container specific rules
+        // POSTROUTING
+        let comment_network_cid = format!(
+            "-m comment --comment 'name: {} id: {}'",
+            network_name, container_id
+        );
+        remove_if_exists(
```
mheon, line 285: What chains are you actually removing here? I think you missed the network's NAT chain

src/commands/setup.rs
baude: i was preventing the case where if it didnt exist, it would not blow up.  but maybe the api calls handle that and then in fact it would be better ... i'll double check that

src/network/core_utils.rs
baude: i was asked to move it

src/firewall/iptables.rs
baude: that code is in a different PR as stated in the commit message

lsm5 (commented):
LGTM

src/firewall/iptables.rs
```diff
@@ -231,13 +304,95 @@ impl firewall::FirewallDriver for IptablesDriver {
...
     }
 
-    fn teardown_port_forward(
-        &self,
-        _container_id: &str,
-        _port_mappings: Vec<types::PortMapping>,
-        _container_ip: &str,
-    ) -> Result<(), Box<dyn Error>> {
-        todo!();
+    fn teardown_port_forward(&self, tear: TeardownPortForward) -> Result<(), Box<dyn Error>> {
```
Luap99, line 307: Why does teardown_port_forward accept a struct but setup_port_forward doe snot. Can you please make this consistent.

baude: I wanted to defer those changes ... would you be comfy if we merge this and then I do that immediately? This PR is already big enough and sweeping changes.

Luap99: sure

PR #83, opened 2021-11-18
Do not serialize the errno field

Luap99:
We already get the exit code when the program exits so there is no need
to include it in the json. Podman will reject unknown fields to prevent
reading unexpected stuff.


mheon:
LGTM

PR #84, opened 2021-11-18
core: add `bridge_count_connected_interfaces` to check if bridge has any connected interface

flouthoc:
add helper to check if bridge has any connected interface

Usage
```rust
    //count interfaces on bridge
    if let Ok(vector) =  bridge_count_connected_interfaces("cni-podman0") {
        println!("{:?}", vector.len());
    }

```

### DIFF
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -34,6 +34,99 @@ impl CoreUtils {
         final_slice.join(":")
     }
 
+    //count interfaces on bridge
+    #[tokio::main]
+    #[allow(irrefutable_let_patterns)]
+    pub async fn bridge_count_connected_interfaces(
+        ifname: &str,
+    ) -> Result<Vec<u32>, std::io::Error> {
+        let mut connected_veth: Vec<u32> = Vec::new();
+        let (_connection, handle, _) = match rtnetlink::new_connection() {
+            Ok((conn, handle, messages)) => (conn, handle, messages),
+            Err(err) => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("failed to connect: {}", err),
+                ))
+            }
+        };
+
+        tokio::spawn(_connection);
+
+        let mut master_id: u32 = 0;
+        let mut master_index: u32 = 0;
+
+        let mut links = handle
+            .link()
+            .get()
+            .set_name_filter(ifname.to_string())
+            .set_filter_mask(AF_BRIDGE as u8, RTEXT_FILTER_BRVLAN)
+            .execute();
+        while let msg = links.try_next().await {
+            match msg {
+                Ok(Some(msg)) => {
+                    for nla in msg.nlas.into_iter() {
+                        if let Nla::Master(data) = nla {
+                            master_id = data;
+                            master_index = msg.header.index;
+                            continue;
+                        }
+                    }
+                }
+                Err(err) => {
+                    return Err(std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        format!(
+                            "Unable to resolve physical address for interface {}: {}",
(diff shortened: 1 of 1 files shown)


flouthoc (commented):
PR has poor error handling, please wait for me to correct error.

flouthoc:
@Luap99 @mheon @baude This is good to merge. I handled the error now.

src/network/core_utils.rs
```diff
@@ -34,6 +34,99 @@ impl CoreUtils {
         final_slice.join(":")
     }
 
+    //count interfaces on bridge
+    #[tokio::main]
+    #[allow(irrefutable_let_patterns)]
```
Luap99, line 39: Why is this needed?

src/network/core_utils.rs
```diff
@@ -34,6 +34,99 @@ impl CoreUtils {
...
+    pub async fn bridge_count_connected_interfaces(
+        ifname: &str,
+    ) -> Result<Vec<u32>, std::io::Error> {
```
Luap99, line 42: Why do we return an vector, it should be enough to return a bool.

src/network/core_utils.rs
```diff
@@ -34,6 +34,99 @@ impl CoreUtils {
...
+        while let msg = links.try_next().await {
+            match msg {
+                Ok(Some(msg)) => {
+                    for nla in msg.nlas.into_iter() {
+                        if let Nla::Master(data) = nla {
+                            if data == master_id {
+                                if msg.header.index == master_index {
+                                    continue;
+                                }
+                                connected_veth.push(msg.header.index);
```
Luap99, line 106: just return true here if we have a match, see other comment

src/network/core_utils.rs
```diff
@@ -34,6 +34,99 @@ impl CoreUtils {
...
+        tokio::spawn(_connection);
+
+        let mut master_id: u32 = 0;
+        let mut master_index: u32 = 0;
+
+        let mut links = handle
+            .link()
+            .get()
+            .set_name_filter(ifname.to_string())
+            .set_filter_mask(AF_BRIDGE as u8, RTEXT_FILTER_BRVLAN)
```
Luap99, line 63: Why do you need this filter here?

src/network/core_utils.rs
```diff
@@ -34,6 +34,99 @@ impl CoreUtils {
...
+            .execute();
+        while let msg = links.try_next().await {
```
Luap99, line 65: Why is this a loop, in the other code parts you use to get a single interface
```
let mut links = handle
            .link()
            .get()
            .set_name_filter(macvlan_tmp_ifname.to_string())
            .execute();
        match links.try_next().await {
...
```

src/network/core_utils.rs
flouthoc: I just thought in future if need to query connected `veth`'s  then function should support so. So it returns a list of  id's of connected veth's

src/network/core_utils.rs
flouthoc: This is not needed on the first block i can remove this.

src/network/core_utils.rs
flouthoc: Yes the first loop is not needed. I amend this.

src/network/core_utils.rs
flouthoc: reasoning in the above comment.

src/network/core_utils.rs
Luap99: I do not see why we need this and this will prevent us from an early exit when we found one interface.

flouthoc: no issues I'll convert this to `bool` if any interface is found.

PR #86, opened 2021-11-19
Do not log the timestamp

Luap99:
There is no value on logging the timestamp IMO. When I make the netavark
interface log to syslog in podman I get 3 timestamps. First the syslog
itself, then the logrus in podman and then netavark env logger.
I think we can just remove this one without breaking anything.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -25,7 +25,7 @@ enum SubCommand {
 }
 
 fn main() {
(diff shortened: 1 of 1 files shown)


Luap99:
@mheon @baude @flouthoc @lsm5 PTAL

flouthoc (approved):

PR #87, opened 2021-11-22
[WIP] define error codes

lsm5:

PTAL. If this looks ok, I can add codes for other parts.

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -1,11 +1,12 @@
 //! Configures the given network namespace with provided specs
-use crate::error::NetavarkError;
+use crate::error::{NetavarkError, NetavarkErrorCode};
 use crate::firewall;
 use crate::firewall::iptables::MAX_HASH_SIZE;
 use crate::network;
 use crate::network::core_utils::CoreUtils;
 use crate::network::internal_types::{SetupNetwork, SetupPortForward};
 use crate::network::{core_utils, types};
+use anyhow::anyhow;
 use clap::{self, Clap};
 use log::debug;
 use std::collections::HashMap;
@@ -69,10 +70,9 @@ impl Setup {
                 "bridge" => {
                     let per_network_opts =
                         network_options.networks.get(net_name).ok_or_else(|| {
-                            std::io::Error::new(
-                                std::io::ErrorKind::Other,
-                                format!("network options for network {} not found", net_name),
-                            )
+                            anyhow!(NetavarkErrorCode::ErrNoNetworkOptions {
+                                network_name: net_name.to_string()
+                            })
                         })?;
                     //Configure Bridge and veth_pairs
                     let status_block = network::core::Core::bridge_per_podman_network(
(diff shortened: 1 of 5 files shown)


lsm5:
can we get some eyes on this please? I can add the same for other parts if this is going in the right direction.

Luap99 (commented):
I think the idea was to remove the `std::io::Error::new( std::io::ErrorKind::Other,...` wrapping and instead use out own struct for everything

mheon:
It doesn't look like this is capable of wrapping errors. Is there a preexisting rust library that allows error wrapping?

lsm5:
> I think the idea was to remove the `std::io::Error::new( std::io::ErrorKind::Other,...` wrapping and instead use out own struct for everything

ack, will do. thanks.

> It doesn't look like this is capable of wrapping errors. Is there a preexisting rust library that allows error wrapping?

`thiserror` which I've already begun using or `error_chain`.

flouthoc:
> I think the idea was to remove the `std::io::Error::new( std::io::ErrorKind::Other,...` wrapping and instead use out own struct for everything

Yes I think we need a `wrap` function which can take any error and wrap it into a common format. I think original `NetavarkError` should be able to do this as well but I haven't tried it yet.

lsm5:
PTAL, let me know if current version looks any better.

src/commands/setup.rs
```diff
@@ -69,10 +70,9 @@ impl Setup {
                 "bridge" => {
                     let per_network_opts =
                         network_options.networks.get(net_name).ok_or_else(|| {
-                            std::io::Error::new(
-                                std::io::ErrorKind::Other,
-                                format!("network options for network {} not found", net_name),
-                            )
+                            anyhow!(NetavarkErrorCode::ErrNoNetworkOptions {
```
flouthoc, line 73: Is there any specific case for `anyhow!`. I am still not sure but I would also request others to pitch in since i was ok with wrapping in the first place.

something like 
```rust
return NetavarkError::Error(err, "my custom wrap message".to_string())
```

flouthoc: I thought this again  I think we just need a new function something like

```rust
return NetavarkError::Error(err, "my custom wrap message".to_string(), ERR_CODE);
```
and NetavarkError::Error returns a valid `NetavarkError` 

so we could remove
```rust
std::io::Error::new( std::io::ErrorKind::Other,...
```

@baude @Luap99 @mheon please correct me if i am wrong.

mheon: You're correct, but I would be fine with using an existing library for wrapping errors (I imagine one does exist already) instead of implementing this ourselves. I don't actually see `anyhow!()` wrapping, though, it accepts an error but not a string to annotate said error with?

lsm5: @mheon @flouthoc Do we need separate annotations if it's mentioned in the enum itself along with the error code? I can update this to get rid of the anyhow!() and use a function to return error code + custom msg if rqeuired. Do you see any issues with the `thiserror` enum created? 

A bunch of crates like `error_chain` and `failure` are now deprecated. anyhow is recommended now instead of failure and thiserror is recommended instead of error_chain.

Let me know.

lsm5:
@flouthoc is this worth resuming or has something else been committed to account for this in the past month?

flouthoc:
@lsm5 I don't think this is still handled. However we are getting relevant error messages from `netavark` in podman's trace output but this can be still nice addition.  Maybe we can discuss this in post-scrum to see what others think.

Luap99:
This was fixed by @mheon so we can close this one

PR #90, opened 2021-11-22
Use structs for input

baude:
When calling setup_network, teardown_network, setup_port_forward, and
teardown_port_forward we now use defined structs for input.

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -4,6 +4,7 @@ use crate::firewall;
 use crate::firewall::iptables::MAX_HASH_SIZE;
 use crate::network;
 use crate::network::core_utils::CoreUtils;
+use crate::network::internal_types::{SetupNetwork, SetupPortForward};
 use crate::network::{core_utils, types};
 use clap::{self, Clap};
 use log::debug;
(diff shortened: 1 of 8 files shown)


src/network/types.rs
```diff
@@ -225,3 +225,36 @@ pub struct TeardownPortForward {
     // remove network related information
     pub complete_teardown: bool,
 }
+
+//  SetupNetwork contains options for setting up a container
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
Luap99, line 230: remove Serialize Deserialize from the structs, we do not want to marshal/unmarshal them

Luap99 (commented):
Can you move the "internal" structs to a new file. I would like to separate internal structs from the "stable" podman interface structs.

src/commands/setup.rs
```diff
@@ -82,21 +83,25 @@ impl Setup {
...
-                                network.clone(),
-                                &network_options.container_id,
-                                i,
-                                net_name,
-                                &id_network_hash,
-                                per_network_opts,
-                                // &id_network_hash.as_str()[0..MAX_HASH_SIZE],
-                            )?;
+                            let spf = SetupPortForward {
+                                net: (*network).clone(),
```
Luap99, line 97: Is there a reason you have to clone the values?

baude: yeah, the network value cannot be moved and the syntax is a result of preventing dereferencing

baude:
fixed up

src/firewall/firewalld.rs
```diff
@@ -1,6 +1,6 @@
 use crate::firewall;
-use crate::network::types;
-use crate::network::types::{Network, PerNetworkOptions, TeardownPortForward};
+use crate::network::internal_types::{SetupPortForward, TearDownNetwork, TeardownPortForward};
+use crate::network::{internal_types, types};
```
mheon, line 3: This seems redundant, you're importing everything in internal_types - and then some specific bits of it as well, above?

PR #91, opened 2021-11-22
Cirrus: Switch CI onto VMs from containers

cevich:
Prior to this commit, CI ran entirely in a container environment.  This
was very much less than ideal, since actual network/firewall
manipulations could not be exercised.  With this commit, everything will
execute on the same VM images and image-build workflow used by all the
other containers-org project automations.

Note: Given all other projects generally use golang, the VM images here
are *NOT* tailored to a rust project.  Unfortunately that means
performing runtime updates/installs and using cache to reduce the dnf
repo. Metadata/package downloading.  These aspects can be removed once
dedicated netavark VM images are realized and implemented in a future
commit.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,26 +1,141 @@
 ---
 
-test_task:
-  alias: "build_and_test"
-  name: "Build and Test"
-  container:
-    image: quay.io/libpod/netavark-devel
-  install_script: make all
-  test_script: make test
+# Format Ref: https://cirrus-ci.org/guide/writing-tasks/
 
-test_task:
-  alias: "validate_test"
-  name: "Validate Code"
-  container:
-    image: quay.io/libpod/netavark-devel
-  test_script: make validate
+# Main collection of env. vars to set for all tasks and scripts.
+env:
+    # Actual|intended branch for this run
+    DEST_BRANCH: "main"
+    # The default is 'sh' if unspecified
+    CIRRUS_SHELL: "/bin/bash"
+    # Location where source repo. will be cloned
+    CIRRUS_WORKING_DIR: "/var/tmp/netavark"
+    # Rust package cache also lives here
+    CARGO_HOME: "/usr/local/cargo"
+    # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
+    SCRIPT_BASE: "./contrib/cirrus"
+    FEDORA_NAME: "fedora-35"
+    IMAGE_SUFFIX: "c6226133906620416"
+    FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
+
+
+gcp_credentials: ENCRYPTED[d6efdb7d6d4c61e3831df2193ca6348bb02f26cd931695f69d41930b1965f7dab72a838ca0902f6ed8cde66c7deddae2]
+
+
+# Default VM to use unless set or modified by task
+gce_instance:
+    image_project: "libpod-218412"
+    zone: "us-central1-c"
+    cpu: 2
+    memory: "4Gb"
+    disk: 200  # GB, do not set <200 per gcloud warning re: I/O performance
+    image_name: "${FEDORA_CACHE_IMAGE_NAME}"
+
+
+build_task:
+  alias: "build"
+  pkg_cache: &pkg_cache # TODO: Remove when packages included in static VM images
+    # This cache is intended to avoid constantly re-downloading repo. metadata
+    # (~150mb) and necessary RPMs (~300mb) every time CI runs.  It will likely
+    # flap often, as the metadata and package DBs change on access.  This is
+    # okay, as the cache still provides a benefit vs constantly hitting the
+    # RPM repos.  Use of this cache depends on dnf arguments passed in setup.sh
+    folder: "/var/cache/dnf"
+    # This is significant, Cirrus-CI will automatically store separate
+    # caches for branches & PRs.  We use the branch-name here mainly to
+    # distinguish PR-level caches in order to properly support backport-PRs
+    # to release branches.  Otherwise all PRs & branches will share caches
+    # with other PRs and branches (for a given $DEST_BRANCH value).
+    fingerprint_key: "pkg_${DEST_BRANCH}"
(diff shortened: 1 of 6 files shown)


cevich:
(Failure is expected for now)

cevich:
Okay, I think this is ready for review at least.  There's certainly room for more optimization but this should be good enough at least for the near-term.

Luap99:
I cannot really review this but I would like to get this merged, @baude @mheon @edsantiago PTAL

cevich:
> WIP Add netavark integration tests #94

Oh you've got some tests ready to go.  Let me rebased that PR into here, so we can see them run in this environment...

cevich:
> I cannot really review this but I would like to get this merged, @baude @mheon @edsantiago PTAL

Heads up, I've got a rebase/force-push incoming along with some mostly cosmetic changes (better commit message, new comments, and renaming 'test' -> 'unit' task).

Luap99:
@cevich It looks like `dbus-daemon` need to be installed in the VM. The other test errors are expected

cevich:
> @cevich It looks like dbus-daemon needs to be installed in the VM. The other test errors are expected

Thanks, that helps.  Trying again...

cevich:
Note: I'm not so happy with the 10minute build-task runtime, I was hoping caching `$CARGO_HOME` would help more than it appears to be.  It's certainly an improvement vs completely cold-cache, but still longer than I'd like.

cevich:
(Rebased/force pushed w/ #94 + moved display of toolchain/env. var details into `runner.sh` for better visibility)

.cirrus.yml
```diff
@@ -1,26 +1,153 @@
...
+# Main collection of env. vars to set for all tasks and scripts.
+env:
+    # Actual|intended branch for this run
+    DEST_BRANCH: "main"
+    # The default is 'sh' if unspecified
+    CIRRUS_SHELL: "/bin/bash"
+    # Location where source repo. will be cloned
+    CIRRUS_WORKING_DIR: "/var/tmp/netavark"
+    # Rust package cache also lives here
+    CARGO_HOME: "/usr/local/cargo"
```
Luap99, line 14: Are you sure you are caching the right directory?

It uses $HOME/.cargo AFAIK, so /root/.cargo if you run as root.

cevich: According my my reading of the cargo (command) docs, setting `$CARGO_HOME` should override this.   The update I just made should allow us to confirm this env. var. is really being set at built-time vs only setup-time.

cevich: Update: Okay, it's certainly being [set at build-time (log)](https://cirrus-ci.com/task/6629489451991040?logs=main#L7).  Hmm.

cevich: I may need to just go in an check this out manually, since this PR adds a working `hack/get_ci_vm.sh`.

cevich: Oh, speaking of which.  Let me make sure you've got access to use that.

Luap99: You are right `CARGO_HOME` env should work but it should compile fast if the cache is already there. If I look at the log I can that it downloads the packages which should not happen with an working cache.

Luap99: I think I understand now. The packages are cached there but this is just for the downloads. The compiled package blobs are stored in the project directory under `bin/release`.

cevich: @Luap99 you've now got access to use the hack script.  So it appears like the `$CARGO_HOME` cache is working from one task to the next.  The latest build ran in ~7m and [validate took only ~2m (log)](https://cirrus-ci.com/task/4588795870838784?logs=main#L110).  So it's only the PR-run to PR-run cache that's flapping for some reason.  I think I'll leave it as-is for now, and dig deeper next week (after US Holiday break).

cevich:
Seems we're missing all the `firewalld` stuff as well, and there's a `Makefile` bug that needs fixing...

Luap99:
@cevich Is this a fedora 35 VM? I thought firewalld is shipped by default.

cevich:
> Is this a fedora 35 VM? I thought firewalld is shipped by default.

It is, the cloud images are different.  It's typical to not run a local firewall and pass this responsibility off to the cloud-provider.  Speaking of which, there's going to be a bit of a likely chicken-vs-egg conflict with the PR/CI workflow here:

* Cirrus-CI depends on local networking (and firewall) allowing outbound traffic to stream output from CI and execute things.
* If the tests break either networking or the local firewall, this fact will not be observable.
* Cirrus-CI will automatically re-try (once) any task where communication to the VM was lost.
* The *only* signal you'll get this is happening is in the Cirrus-CI WebUI, it will say (in red): CI Agent Not Responding.

In the case of a non-GCP privileged user's PR, they will need to be assisted with debugging, possibly with manually adding their ssh key to a VM from `hack/get_ci_vm.sh` - in other words, it could be an ugly, frustrating situation to deal with :confused:

Luap99:
@cevich These test do not touch the host network at all. Each test run in their own network namespace to prevent any leaking iptables or network interfaces.

cevich:
> These test do not touch the host network at all.

Good thinking.  Still, kernel bugs or test-goofs happen.  The frustrating thing is you'll get *zero* clue this has occured: Cirrus-CI output will just "stop", the VM will go down (destroying the logs), and the task will restart (once).  I know from past-experience this is a very (VERY) frustrating situation to deal with.  So after this merges, please make sure you can successfully use `hack/get_ci_vm.sh`, assuming the problem is reproducible, that will be your only hope.

.cirrus.yml
cevich: Just checked a VM running the build task.  Indeed there's no `$HOME/.cargo` only `/usr/local/cargo`.  So this must be a Cirrus-CI Caching problem - it's not storing the content that I expect, or something is causing flapping.  In any case, I'm leaving this debugging up to the future.

cevich:
Okay, last update I'm going to make for now: I commented out execution of the integration tests, but left all the bits in there.  I think this PR is ready for review and merge - Unless there are horrible problems with my `bash` (mostly copy/pasted from other projects).

Luap99 (commented):
LGTM

PR #94, opened 2021-11-23
Add netavark integration tests

Luap99:
Add bats tests for netavark. The test run netavark in completely
separate network namespaces to not leak any information on the host. The
test runs with the iptables driver but also with the firewalld driver
see test/020-firewalld.bats. This is important to cover all firewall
driver code paths.

You can run the tests with `sudo bats test/` or `podman unshare
--rootless-netns bats test/`. Note that the `--rootless-netns` flag is
called `--rootless-cni` for podman < v4.0.

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -1,0 +1,590 @@
+# -*- bash -*-
+
+# Netavark binary to run
+NETAVARK=${NETAVARK:-./bin/netavark}
+
+TESTSDIR=${TESTSDIR:-$(dirname ${BASH_SOURCE})}
+
+# export RUST_BACKTRACE so that we get a helpful stack trace
+export RUST_BACKTRACE=full
+
+# this will cause tests to fail because stdou/stderr are not separate
+# export RUST_LOG=netavark=debug
+
+HOST_NS_PID=
+CONTAINER_NS_PID=
+
+function basic_setup() {
+    HOST_NS_PID=$(create_netns)
+    CONTAINER_NS_PID=$(create_netns)
+
+    # make sure to set DBUS_SYSTEM_BUS_ADDRESS to an empty value
+    # netavark will try to use firewalld connection when possible
+    # because we run in a separate netns we cannot use firewalld
+    # firewalld run in the host netns and not our custom netns
+    # thus the firewall rules end up in the wrong netns
+    # unsetting does not work, it would use the default address
+    export DBUS_SYSTEM_BUS_ADDRESS=
+
+    NETAVARK_TMPDIR=$(mktemp -d --tmpdir=${BATS_TMPDIR:-/tmp} netavark_bats.XXXXXX)
+
+    run_in_host_netns ip link set lo up
+}
+
+function basic_teardown() {
+    kill -9 $HOST_NS_PID
+    kill -9 $CONTAINER_NS_PID
+
+    rm -rf "$NETAVARK_TMPDIR"
+}
+
+# Provide the above as default methods.
+function setup() {
+    basic_setup
+}
+
+function teardown() {
+    basic_teardown
+}
+
+function create_netns() {
+    # create a new netns and mountns and run a sleep process to keep it alive
+    # we have to redirect stdout/err to /dev/null otherwise bats will hang
+    unshare -n sleep inf &>/dev/null &
+    echo $!
+}
+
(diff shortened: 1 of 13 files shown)


Luap99:
@edsantiago Could you please look at this (not urgent).

test/helpers.bash
```diff
@@ -0,0 +1,236 @@
...
+        printf "#|         > '%s'\n" "$line" >&2
+    done
+    printf "#\\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n" >&2
+    false
+}
+
+function is_json() {
+    JSON="$1"
+    run_helper jq -r "$2" <<<"$JSON"
+    is "$output" "$3" "$4"
```
edsantiago, line 235: Why refactor `$JSON` but not the other positional args?

test/helpers.bash
```diff
@@ -0,0 +1,236 @@
...
+#########
+function die() {
+    # FIXME: handle multi-line output
+    echo "#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv" >&2
+    echo "#| FAIL: $*" >&2
+    echo "#\\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^" >&2
+    false
+}
+
+########
```
edsantiago, line 173: Would you consider using `assert` and `expect_output` from [buildah:helpers.bash](https://github.com/containers/buildah/blob/68cd60292259fa6b9fe09b9eee06d391eaf8228a/tests/helpers.bash#L252) instead of from podman? Podman tests came first; when I worked on Buildah I had learned a lot from my mistakes. Those are better. (And yes, I always meant to backport, but no, it's probably too late to do so).

test/helpers.bash
```diff
@@ -0,0 +1,236 @@
...
+
+function get_container_netns_path() {
+    echo /proc/$CONTAINER_NS_PID/ns/net
+}
+
+################
+#  run_netavark  #  Invoke $NETAVARK, with timeout, using BATS 'run'
+################
+#
+# This is the preferred mechanism for invoking podman: first, it
```
edsantiago, line 65: You may want to search for other copy-paste uses of `podman`

test/firewalld-dbus.conf
```diff
@@ -0,0 +1,65 @@
+<!DOCTYPE busconfig PUBLIC "-//freedesktop//DTD D-Bus Bus Configuration 1.0//EN" "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
```
edsantiago, line 1: Wouldn't this file fit better under `testfiles`?

edsantiago (commented):
Very quick first pass. Nice work! I will try to make a second pass tomorrow.

test/helpers.bash
Luap99: Is there any point in using `expect_output`, it looks like assert is enough?

test/firewalld-dbus.conf
Luap99: yes

test/helpers.bash
edsantiago: It's a minor shortcut for what is by far the most common expression.
```bash
    assert "$output" = "foo bar" "test failed"
    expect_output "foo bar" "test failed"
```
There's just a tiny bit less chance of screwing it up using the second form; and fewer quotation marks makes it slightly more readable. But it's personal preference, really, and I can see a benefit to using only `assert`.

Luap99:
@mheon @baude @lsm5 @flouthoc PTAL
This is still WIP because I would like to merge https://github.com/containers/netavark/pull/91 first and then rebase but you already can use this for local testing if needed.

flouthoc:
Thanks for adding this i was really having trouble running tests on my `host`. I'll try it out and will report back in case of any issues.

Luap99:
This is ready to merge, PTAL

edsantiago:
@Luap99 in case you did a fire-and-forget: we're not ignoring you, we're waiting for you to fix the CI failure.

edsantiago:
LGTM

flouthoc (approved):

PR #95, opened 2021-11-24
Fix make: 'test' is up to date.

cevich:
Without this, running 'make test' will fail to execute if the `test` directory timestamp is current.


lsm5 (approved):
LGTM

PR #96, opened 2021-11-24
initial ipv6 implementation

baude:
enable ipv6 connection handling

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -12,6 +12,7 @@ use std::collections::HashMap;
 use std::error::Error;
 
 const IPV4_FORWARD: &str = "net.ipv4.ip_forward";
+const IPV6_FORWARD: &str = "/proc/sys/net/ipv6/conf/all/forwarding";
 
 #[derive(Clap, Debug)]
 pub struct Setup {
@@ -53,7 +54,7 @@ impl Setup {
         };
 
         // Sysctl setup
-        // set ip forwarding to 1
+        // set ipv4 forwarding to 1
         core_utils::CoreUtils::apply_sysctl_value(IPV4_FORWARD, "1")?;
 
         let mut response: HashMap<String, types::StatusBlock> = HashMap::new();
@@ -64,7 +65,10 @@ impl Setup {
                 "Setting up network {} with driver {}",
                 net_name, network.driver
             );
-
+            // set ipv6 forwarding to 1
+            if network.ipv6_enabled {
+                core_utils::CoreUtils::apply_sysctl_value(IPV6_FORWARD, "1")?;
+            }
             match network.driver.as_str() {
                 "bridge" => {
                     let per_network_opts =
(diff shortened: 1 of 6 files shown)


src/commands/setup.rs
```diff
@@ -77,29 +81,36 @@ impl Setup {
...
                             let spf = SetupPortForward {
-                                net: (*network).clone(),
+                                net: network.clone(),
                                 container_id: network_options.container_id.clone(),
                                 port_mappings: i,
-                                network_name: (*net_name).clone(),
+                                network_name: net_name,
                                 network_hash_name: id_network_hash,
-                                options: (*per_network_opts).clone(),
+                                container_ip: container_ips[idx],
```
Luap99, line 113: The loop order does not make sense! You use the index from the networks loop but the static_ips are per network, len(static_ips) == len(subnets). What you have to do for your port forwarding logic is loop over the static ips and take the first ipv4 (if there is one) and the first ipv6 (if there is one). If there are only ipv4 you only forward to the first ipv4 not all ips. The same for ipv6.

The same for the teardown logic

src/firewall/iptables.rs
```diff
@@ -189,113 +192,101 @@ impl firewall::FirewallDriver for IptablesDriver {
...
-                container_ip.to_string(),
+                setup_portfw.container_ip.to_string(),
                 comment_network_cid
             ),
         )?;
 
         // FOR EACH PORT
-        for i in setup_portfw.port_mappings {
+        for i in setup_portfw.port_mappings.clone() {
+            error!("ipv6 is {}", is_ipv6);
```
Luap99, line 243: This should be at least debug

baude: of course, this is all just here until we confirm it works.  then i will do more code comments and cleanups

Luap99:
For ipv6 we have to set `/proc/sys/net/ipv6/conf/<brname>/accept_dad` and `/proc/sys/net/ipv6/conf/<brname>/accept_ra` to 0. I also think the same systcls are needed for the host veth interface.

baude:
please review and merge

src/network/core_utils.rs
```diff
@@ -715,7 +715,18 @@ impl CoreUtils {
                 ))
             }
         }
-
+        // TODO, this should only be done for IPv6
```
flouthoc, line 718: Why is this inside `configure_bridge`.

src/network/core_utils.rs
```diff
@@ -911,7 +922,16 @@ impl CoreUtils {
                 ))
             }
         }
-
+        let k = format!("/proc/sys/net/ipv6/conf/{}/enhanced_dad", &host_veth);
```
flouthoc, line 925: Same I dont think we should do a `sysctl` while configuring `veth`

src/network/core_utils.rs
flouthoc: I think we should have a different function for this and call `sysctl` in `core.rs`

src/network/core_utils.rs
flouthoc: I think we should have a different function for this and call `sysctl` in `core.rs`

flouthoc:
@baude Above two point looks like a blocker to me. Netlink functions shouldn't be configuring `sysctls` internally.

flouthoc:
Other than these two points LGTM

src/network/core_utils.rs
flouthoc: Actually I'd also like to request why is this not possible in `setup.rs` or with firewall where other `sysctls` are made if `driver` is `bridge`.

src/network/core_utils.rs
flouthoc: After checking this , we already have `host_veth` at higher level so this better in `core.rs` where we are calling function to configure bridge.

src/network/core_utils.rs
Luap99: Yes the systctl are part of the bridge setup, I am not sure if need to set them for other drivers. Also it is important the the systctl is set before the interface is set up/get the ip assigned.

src/network/core_utils.rs
Luap99, line 925: This is the wrong sysctl, you need to set `accept_dad` not `enhanced_dad`

Luap99: Also you have set both sysctls for the bridge interface as well not only the veth

PR #100, opened 2021-11-29
Cargo.toml: exclude unnecessary files from crate

lsm5:
Fedora's Rust packaging guidelines recommend excluding unnecessary files
from Cargo.toml itself.

https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_excluding_unnecessary_files

PTAL

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.0.1"
 edition = "2018"
 authors = ["github.com/containers"]
 description = "A container network stack"
+exclude = ["/.cirrus.yml", "/.github/*", "/hack/*"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]


mheon:
LGTM

Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.0.1"
 edition = "2018"
 authors = ["github.com/containers"]
 description = "A container network stack"
+exclude = ["/.cirrus.yml", "/.github/dependabot.yml"]
```
Luap99, line 7: Can you exclude the full .github directory in case someone adds a new file in the future.

lsm5: Done, added hack dir as well.

lsm5:
added entire `.github` dir and also added `hack` dir. Will the `hack` dir ever have anything useful to end users?

Luap99 (commented):
LGTM

PR #101, opened 2021-11-29
core: generate host veth name early in the stack

flouthoc:
get host mac early for usage

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -95,11 +95,15 @@ impl Core {
         debug!("IP address for veth vector: {:?}", address_vector);
         debug!("Gateway ip address vector: {:?}", gw_ipaddr_vector);
 
+        // get random name for host veth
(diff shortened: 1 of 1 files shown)


baude:
LGTM

src/network/core.rs
```diff
@@ -110,6 +114,16 @@ impl Core {
...
+        let host_veth_mac = match core_utils::CoreUtils::get_interface_address(&host_veth_name) {
+            Ok(addr) => addr,
+            Err(err) => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("failed to get host veth mac: {}", err),
+                ))
+            }
+        };
+        debug!("Host veth mac: {:?}", host_veth_mac);
```
Luap99, line 126: Why do we need the host veth mac?

flouthoc: @Luap99 I think @baude needs it for some of his experiments.

PR #103, opened 2021-11-29
Cirrus: Use extra CPU/Mem for build task

cevich:
This is needed to improve the initial build performance, which despite
caching `$CARGO_HOME` still takes ~10min to complete initially.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,7 +23,7 @@ gcp_credentials: ENCRYPTED[d6efdb7d6d4c61e3831df2193ca6348bb02f26cd931695f69d419
 
 
 # Default VM to use unless set or modified by task
-gce_instance:
+gce_instance:  &standard_vm
     image_project: "libpod-218412"
     zone: "us-central1-c"
     cpu: 2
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -34,6 +34,11 @@ gce_instance:
 
 build_task:
   alias: "build"
+  # Compiling is very CPU intensive, make it chooch quicker for this task only
+  gce_instance:
+    <<: *standard_vm
+    cpu: 8
+    memory: "16Gb"
```
Luap99, line 41: I do not think we need the extra memory, see https://cirrus-ci.com/task/6358768268083200. It was at 1.57 GB max for this task.

cevich: Okay, so it took 01:56 like this.  I dropped it back to 4gig, let's see if the duration changes at all.  Less memory is slightly cheaper if we don't need it.  It's the CPUs that are the big expense...IIUC we're charged by the hour even if the task takes a minute.

cevich: aaaand it looks like 8gig is the minimum if we request 8 CPUs.

cevich: It's showing 2-minutes with 8 + 8, so let's run with that.

Luap99 (commented):
LGTM

lsm5 (approved):
/lgtm

PR #109, opened 2021-11-30
Improve Makefile + Caching in CI

cevich:
This is fairly complex unfortunately, but should reduce the amount of re-compiling we do in CI.  It also makes the `make` + `cargo` situation clearer WRT all the `.PHONY` targets, and disambiguation of the var, args, and term `target`.

Lastly, the "total success" aggregation task now uploads the compiled + tested `netavark` and `netavark.debug` binaries.  The intention is to make use of these in downstream CI (like podman).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -11,7 +11,9 @@ env:
     # Location where source repo. will be cloned
     CIRRUS_WORKING_DIR: "/var/tmp/netavark"
     # Rust package cache also lives here
-    CARGO_HOME: "/usr/local/cargo"
+    CARGO_HOME: "/var/cache/cargo"
+    # Rust compiler output lives here (see Makefile)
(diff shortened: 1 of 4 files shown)


cevich:
***Note:*** It still seems the `unit` task compiles way too many things, despite having applied the `make build_tests` target and (cached the output) during build-task :confused: Leaving this up to someone else to fix.

cevich:
This is ready for review.

cevich:
Rebased and force-pushed.

flouthoc (approved):

cevich:
Rebased and force-pushed (some minor comment-changes made)

Luap99 (commented):
I cannot really review this. My makefile/cirrus yaml knowledge is almost non existent.
But from what I understand LGTM

Luap99:
@baude @mheon PTAL

PR #110, opened 2021-12-01
core: configure required kernel parameters `accept_ra` and `accept_dad` for `ipv6` not for `ipv4`

flouthoc:
Explicitly required kernel parameters like `accept_ra` and `accept_dad` are
only needed for interface when we are using `ipv6` not `ipv4`.

Standard netavark configuration has a field `ipv6_enabled` which should be enough
for us to propagate configuration to lower utility functions in the
stack.

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -53,6 +53,8 @@ impl Core {
                 }
             }
         }
+        // is ipv6 enabled, we need to propogate this to lower stack
+        let ipv6_enabled = network.ipv6_enabled;
 
         let container_veth_name: String = per_network_opts.interface_name.to_owned();
         let static_ips: &Vec<IpAddr> = per_network_opts.static_ips.as_ref().unwrap();
@@ -123,6 +125,7 @@ impl Core {
             &host_veth_name,
             netns,
             mtu_config,
+            ipv6_enabled,
         ) {
             Ok(addr) => addr,
             Err(err) => {
@@ -143,6 +146,7 @@ impl Core {
         Ok(response)
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub fn add_bridge_and_veth(
         br_name: &str,
         netns_ipaddr: Vec<ipnet::IpNet>,
@@ -151,6 +155,7 @@ impl Core {
         host_veth_name: &str,
         netns: &str,
         mtu_config: u32,
+        ipv6_enabled: bool,
     ) -> Result<String, std::io::Error> {
         //copy subnet masks and gateway ips since we are going to use it later
         let mut gw_ipaddr_clone = Vec::new();
@@ -158,8 +163,12 @@ impl Core {
             gw_ipaddr_clone.push(*gw_ip)
         }
         //call configure bridge
-        let _ = match core_utils::CoreUtils::configure_bridge_async(br_name, gw_ipaddr, mtu_config)
-        {
+        let _ = match core_utils::CoreUtils::configure_bridge_async(
+            br_name,
+            gw_ipaddr,
+            mtu_config,
+            ipv6_enabled,
+        ) {
             Ok(_) => (),
             Err(err) => {
                 return Err(std::io::Error::new(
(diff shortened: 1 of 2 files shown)


flouthoc:
@Luap99 @baude This is a follow up PR after `ipv6` support. PTAL

src/network/core.rs
```diff
@@ -36,6 +36,8 @@ impl Core {
         let mut response_net_addresses: Vec<NetAddress> = Vec::new();
         // interfaces map, but we only ever expect one, for response
         let mut interfaces: HashMap<String, types::NetInterface> = HashMap::new();
+        // is ipv6 enabled, we need to propogate this to lower stack
+        let ipv6_enabled = network.ipv6_enabled;
```
flouthoc, line 40: Note: We are good here since this is a native `bool` not an `Option<bool>`.

Luap99 (commented):
I am not sure why we need this, `ipv6_enabled` is mostly used for user input when creating networks. I do not think we can guarantee that this is always correctly set. There is absolutely zero harm to set this sysctl when there is no ipv6 ip address.

flouthoc:
@Luap99 Following check is used to set forwarding for `IPv6` as well https://github.com/containers/netavark/blob/main/src/commands/setup.rs#L69 and kernels with disabled `ipv6` will fail with `ENOENT` for any path containing `/proc/sys/net/ipv6/` which shouldn't happen for `ipv4` setup.

But i understand your concern in worst case even if we don't set `ipv6_enabled` from podman serializer will automatically set it to `false`

Luap99:
ack, LGTM

flouthoc:
@baude @mheon @lsm5 PTAL

baude:
> I am not sure why we need this, `ipv6_enabled` is mostly used for user input when creating networks. I do not think we can guarantee that this is always correctly set. There is absolutely zero harm to set this sysctl when there is no ipv6 ip address.

It is a perf hit to netavark.  if a simple code conditional can avoid doing it, we should as well.  also, the is_ipv6 variable is not useful for dual stacks, so I have always used an ip address of some sort to determine if ipv6

src/network/core.rs
baude: see my other comment, this does not work for dual stacks.

baude:
make certain the network the interface in question is actually ipv6 and then LGTM

src/network/core.rs
flouthoc: yes added check for `dual-stack` configurations.

flouthoc:
> > I am not sure why we need this, `ipv6_enabled` is mostly used for user input when creating networks. I do not think we can guarantee that this is always correctly set. There is absolutely zero harm to set this sysctl when there is no ipv6 ip address.
> 
> It is a perf hit to netavark. if a simple code conditional can avoid doing it, we should as well. also, the is_ipv6 variable is not useful for dual stacks, so I have always used an ip address of some sort to determine if ipv6

Added a `ipv6_enabled` modifier for `dual-stack` configurations in a new commit where i am explicitly checking if any of the ips is `v6`

flouthoc:
@Luap99 @baude PTAL made changes.

baude:
LGTM now

flouthoc:
@Luap99 @baude I have one doubt for checking dual stack. Could it be possible that no static `ipv6` is specified and config had `ipv6_enabled` as false but gateway was `ipv6` ?

I am assuming that is a human-error and no manager will generate a config like that but just want to ask.

flouthoc:
It shoud be okay if `podman` sets `ipv6_enabled` when gateway is also `ipv6`.

flouthoc:
@Luap99 Ping any thoughts on this ?

Luap99:
That is a config error. I do not think we have to care for me the moment.

flouthoc:
> That is a config error. I do not think we have to care for me the moment.

Fair enough. Seems like `user` config issue, lets leave it for another day.

flouthoc:
I'll add it in code comments or documentation though whenever we are going to have documentation.

Issue #107, opened 2021-11-30
Set MTU


baude:
netavark needs to support the ability to set the MTU

flouthoc:
Picking this one up.

flouthoc:
above PR should close this.

PR #111, opened 2021-12-01
core: allow users to configure `MTU` for `bridge` and `veth` interfaces from config

flouthoc:
Netavark config accepts `MTU` as network option so allow users to
configure the same for both `bridge` and `veth` interfaces if defined.

Closes: https://github.com/containers/netavark/issues/107

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -36,6 +36,23 @@ impl Core {
         let mut response_net_addresses: Vec<NetAddress> = Vec::new();
         // interfaces map, but we only ever expect one, for response
         let mut interfaces: HashMap<String, types::NetInterface> = HashMap::new();
+        // mtu to configure, 0 means it was not set do nothing.
+        let mut mtu_config: u32 = 0;
+        if let Some(options_map) = network.options.as_ref() {
+            if let Some(mtu) = options_map.get("mtu") {
+                match mtu.parse() {
+                    Ok(mtu) => {
+                        mtu_config = mtu;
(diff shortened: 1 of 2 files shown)


flouthoc:
I want to rebase this PR after: https://github.com/containers/netavark/pull/110

src/network/core.rs
```diff
@@ -36,6 +36,23 @@ impl Core {
         let mut response_net_addresses: Vec<NetAddress> = Vec::new();
         // interfaces map, but we only ever expect one, for response
         let mut interfaces: HashMap<String, types::NetInterface> = HashMap::new();
+        // mtu to configure, 0 means it was not set do nothing.
```
flouthoc, line 39: We should not set `mtu` to 0. Although `kernel` allows it but its a bug https://serverfault.com/a/652604

So netavark will `NOOP` in case of `mtu: 0`

Luap99: I agree 0 means default so we should do nothing

Luap99:
LGTM

flouthoc:
whoops I really wanted to get this merged first #110 no issues i'll rebase the other one.

src/network/core_utils.rs
```diff
@@ -750,6 +789,12 @@ impl CoreUtils {
             }
         }
 
+        if mtu != 0 {
```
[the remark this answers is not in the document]
flouthoc: Sure i'll fix this.

PR #112, opened 2021-12-01
Downgrade nix to what exists in Fedora

lsm5:
It can be painful to bump the rust-nix package in Fedora.

Dependabot upgrades of nix should be blocked until the version is
packaged in Fedora.

@mheon @baude @Luap99 @flouthoc need to get this merged to make Fedora packaging happy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -560,7 +560,7 @@ dependencies = [
  "libc",
  "log",
  "netlink-packet-route",
- "nix 0.23.0",
+ "nix 0.22.0",
  "rand",
  "rtnetlink",
  "serde",
@@ -670,19 +670,6 @@ name = "nix"
 version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf1e25ee6b412c2a1e3fcb6a4499a5c1bfe7f43e014bdce9a6b6666e5aa2d187"
-dependencies = [
- "bitflags",
- "cc",
- "cfg-if 1.0.0",
- "libc",
- "memoffset",
-]
-
-[[package]]
-name = "nix"
-version = "0.23.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f305c2c2e4c39a82f7bf0bf65fb557f9070ce06781d4f2454295cc34b1c43188"
 dependencies = [
  "bitflags",
  "cc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ url = "2.1.0"
 zbus = "1.9.1"
 rtnetlink = { git = "https://github.com/little-dude/netlink", branch = "master" }
 futures = "0.3.18"
-nix = "0.23.0"
+nix = "0.22.0" # DO NOT BUMP UNTIL ALREADY PACKAGED IN FEDORA
 rand = "0.8.3"
 faccess = "0.2.3"
 tokio = { version = "1.14.0", features = ["full"] }


flouthoc:
LGTM. 

@lsm5  Just a small doubt I am wondering why `fedora` has a objection with the version we are using since its going to be local to this crate and specified version for `nix` is already released. For example we already use a custom version for `rtnetlink = { git = "https://github.com/little-dude/netlink", branch = "master" }`

lsm5:
@flouthoc Quoting Fabio Valentini in #fedora-rust IRC channel:
```
nix is a bit of a special case, because it breaks compatibility quite often, and we don't want to use compat packages for every 0.17, 0.18, 0.19, 0.20, 0.21, 0.22, and 0.23 release ... :(
but you could try temporarily downgrading to nix 0.22. unless you use any new APIs that should work, it is what some other crates already do.
```

lsm5:
> @flouthoc Quoting Fabio Valentini in #fedora-rust IRC channel:
> 
> ```
> nix is a bit of a special case, because it breaks compatibility quite often, and we don't want to use compat packages for every 0.17, 0.18, 0.19, 0.20, 0.21, 0.22, and 0.23 release ... :(
> but you could try temporarily downgrading to nix 0.22. unless you use any new APIs that should work, it is what some other crates already do.
> ```

I can add this to the commit log itself if that's preferable

flouthoc:
@Luap99 I think this edge case is only for `nix` other packages could be of any version.

flouthoc:
@lsm5 Please correct me if i am wrong.

flouthoc (approved):

flouthoc:
LGTM other than @Luap99 's concern.

lsm5:
@Luap99 @flouthoc let me check with the fedora-rust people

lsm5:
> @lsm5 I am fine with downgrading but as @flouthoc points out we use the master branch from the netlink lib, will this cause trouble for packaging as well? I am pretty sure we need this for macvlan.

@Luap99 yes, if the master branch of rtnetlink is not already present in fedora, it can cause issues with packaging the latest netavark tags in fedora, so it would be really good to either use fedora packaged versions / latest tags or I could try patching the requirement in the netavark rpm sources, but I imagine that could be a painful process.

We could also nudge rtnetlink people to release faster if that's a worthwhile option.

lsm5:
@Luap99 @flouthoc the fedora rust packaging people probably have some automation in place, but I usually find the latest published crates are already available on fedora, with some exceptions like nix of course. So, getting rtnetlink folks to release faster should be good for everyone I think.

lsm5:
@cevich ping, is it possible to create a CI job that'll check if a crate version is already packaged in fedora?

lsm5:
also, bundling *could* be an option in case we need the latest and greatest, but unless the bundling happens in the upstream netavark source itself, it would mean lot of PITA for packaging / automation of rpms etc.

cevich:
@lsm5 this is a much more nuanced issue, as the ultimate practical requirements are specified in Cargo.toml which has no connection to if/what RPMs are actually installed.  Meaning, anybody can alter the file and immediately cause a disconnect in CI vs user/developer experience.  You know, the environments we want to keep as close as possible :cry:  This is the issue I had in mind the other day when I asked about netavark packaging.  We should probably discuss it as a team, since it could potentially affect everybody.

lsm5:
> @lsm5 this is a much more nuanced issue, as the ultimate practical requirements are specified in Cargo.toml which has no connection to if/what RPMs are actually installed. Meaning, anybody can alter the file and immediately cause a disconnect in CI vs user/developer experience. You know, the environments we want to keep as close as possible cry This is the issue I had in mind the other day when I asked about netavark packaging. We should probably discuss it as a team, since it could potentially affect everybody.

ack, it's only after the call yesterday that I realized the pain points with rust dependencies and Fedora.

lsm5:
@cevich I'm closing this. We can first explore the cargo vendor route like @flouthoc suggested. I can patch the nix requirement in the rpm itself for now.

PR #113, opened 2021-12-02
Fix UDP and network stuff

baude:
Add support for better error messages and also things to help integration test pass UDP and IPv6.

### DIFF
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -832,7 +832,7 @@ impl CoreUtils {
         if let Err(err) = handle
             .link()
             .add()
-            .veth(host_veth.to_string(), container_veth.to_string())
+            .veth(container_veth.to_string(), host_veth.to_string())
             .execute()
             .await
         {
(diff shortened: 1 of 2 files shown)


Luap99 (commented):
LGTM and fixes our weird ipv6 issue

flouthoc (approved):

src/network/core_utils.rs
```diff
@@ -832,7 +832,7 @@ impl CoreUtils {
         if let Err(err) = handle
             .link()
             .add()
-            .veth(host_veth.to_string(), container_veth.to_string())
+            .veth(container_veth.to_string(), host_veth.to_string())
```
flouthoc, line 835: @baude Sorry i think i merged it. Just one doubt I understand that swapping `veth` order is fixing the issue but not sure how ? Is it a `kernel` bug ?

Luap99: I am not sure tbh. I just tried until it works. I have a script with only `ip` commands to reproduce the bridge setup. However with `ip link add` the order did not matter so it is not only the kernel. I think the `ip link` command adds some netlink attributes and our rust code does not.

PR #117, opened 2021-12-03
include %H%M%S in the release tag

lsm5:
In the case of multiple PR merges and thus multiple COPR builds
everyday, it would be good to also include HHMMSS in the release tag.

@flouthoc @baude @mheon PTAL

### DIFF
--- a/.autocopr/netavark.spec
+++ b/.autocopr/netavark.spec
@@ -3,14 +3,16 @@
 Name: netavark
 Epoch: 100
 Version: 0
-%define build_timestamp %{lua: print(os.date("%Y%m%d"))}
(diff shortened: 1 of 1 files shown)


lsm5:
@edsantiago PTAL here as well.

flouthoc (approved):

edsantiago:
LGTM

lsm5:
@mheon @baude PTAL.

lsm5:
gotta check if centos 9 stream builds successfully too

lsm5:
removed `rust-packaging` dependency and added cargo explicitly so that centos doesn't complain.

PR #118, opened 2021-12-03
Cirrus: Dedicated VM images for CI

cevich:
Image comes from https://github.com/containers/automation_images/pull/101

The primary benefits to using a dedicated netavark VM image are:

* Predictable environment changes.  Underlying packages are only updated when we decide to build + incorporate an updated image.
* Improved Speed/efficiency of CI.  Remove the 1-2minutes spent in each task updating/installing required packages, since they're pre-installed in the image.

Total runtime in this PR: 11-13min vs 18-20min in #109 

***Depends on:*** https://github.com/containers/netavark/pull/109

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,8 +17,8 @@ env:
     # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
     SCRIPT_BASE: "./contrib/cirrus"
     FEDORA_NAME: "fedora-35"
-    IMAGE_SUFFIX: "c6226133906620416"
-    FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
+    IMAGE_SUFFIX: "c4559025915297792"
+    FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
 
 
 gcp_credentials: ENCRYPTED[d6efdb7d6d4c61e3831df2193ca6348bb02f26cd931695f69d41930b1965f7dab72a838ca0902f6ed8cde66c7deddae2]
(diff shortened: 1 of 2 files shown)


cevich:
(I dunno why Cirrus-CI tests aren't starting - I'll investigate this next week.)

Luap99:
@cevich I noticed in my integration tests that the images doesn't have `kernel-modules` installed. Can you add this package. This is needed for the sctp protocol, we have to load it with `modprobe sctp`.

cevich:
> I noticed in my integration tests that the images doesn't have kernel-modules installed. Can you add this package. This is needed for the sctp protocol, we have to load it with modprobe sctp.

Certainly can.  I think there's even a file I can stick it into so it's loaded at boot (if you know, tell me).

Luap99:
@cevich There is no need to load it automatically, the tests that need it will load it. If the module is not found the test will be skipped. Mainly because this module is also not loaded on my local system.

cevich:
Okay, this is fundamentally working with the new dedicated VM image.

@Luap99 do you have any other special requests for the image before I build again to include `kernel-modules` and force-loading of `sctp` on boot?

Ref: [package selection is here.](https://github.com/containers/automation_images/pull/101/files#diff-94f63acfc03dca806a430fa27d497636071f97c051bedd9af0925656bce8024aR21-R50)

Luap99:
I think the package `nftables` would be needed at some point in the future if we implement the nftables backend.

cevich:
> I think the package nftables would be needed at some point in the future

K., got it.

cevich:
***Note*** This includes commits from #109

Luap99:
@cevich Can we merge or do you want to rebase after #109?

Luap99 (commented):
LGTM

cevich:
Nope, no rebase needed, merge away.

cevich:
hmmm...actually.  It still shows 3-commits here.  Let me rebase so we don't dupe them...

PR #119, opened 2021-12-06
core: disable duplication address detection inside container only if its `IPv6`

flouthoc:
Disable duplication address detection inside container namespace only if
its `ipv6` not `ipv4`

This is similar to : https://github.com/containers/netavark/pull/110
After: https://github.com/containers/netavark/pull/113


flouthoc:
@Luap99 @baude PTAL

flouthoc:
@baude @mheon PTAL

PR #125, opened 2021-12-09
Refactor iptables

baude:
Switch to a model where we use a rule generator approach for setup and
teardown.  This eliminates a significant amount duplicate code.

### DIFF
--- a/src/commands/teardown.rs
+++ b/src/commands/teardown.rs
@@ -1,7 +1,9 @@
 use crate::error::NetavarkError;
 use crate::firewall::iptables::MAX_HASH_SIZE;
 use crate::network::core_utils::CoreUtils;
-use crate::network::internal_types::{TearDownNetwork, TeardownPortForward};
+use crate::network::internal_types::{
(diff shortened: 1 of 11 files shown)


src/commands/teardown.rs
```diff
@@ -104,23 +106,30 @@ impl Teardown {
                                         "no network address provided",
                                     )
                                 })?;
-                                let td = TeardownPortForward {
-                                    network: network.clone(),
+                                let spf = SetupPortForward {
```
mheon, line 109: If we're refactoring already, maybe rename this so SetupPortForward isn't used to tear down port forwarding. `PortForwardConfig` maybe?

test/helpers.bash
```diff
@@ -484,7 +484,7 @@ EOF
     done
 
     ## FIXME cleanup is broken atm
```
Luap99, line 486: Cab you also remove this comment here

src/commands/teardown.rs
Luap99: +1

baude:
Fixed up local test failures and resolved comments.

PR #126, opened 2021-12-09
Support internal networks

baude:
An internal network should have no nat or port mappings as it is
supposed to a "closed" network.

Fixes #: 105

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -69,6 +69,8 @@ impl Setup {
             if network.ipv6_enabled {
                 core_utils::CoreUtils::apply_sysctl_value(IPV6_FORWARD, "1")?;
             }
+            // If the network is internal, we override the global setting and disabled forwarding
+            // on a per interface instance
             match network.driver.as_str() {
                 "bridge" => {
                     let per_network_opts =
@@ -85,7 +87,25 @@ impl Setup {
                         &self.network_namespace_path,
                     )?;
                     response.insert(net_name.to_owned(), status_block);
-
+                    if network.internal {
+                        match &network.network_interface {
+                            None => {}
+                            Some(i) => {
+                                core_utils::CoreUtils::apply_sysctl_value(
+                                    format!("/proc/sys/net/ipv4/conf/{}/forwarding", i).as_str(),
+                                    "0",
+                                )?;
+                                if network.ipv6_enabled {
+                                    core_utils::CoreUtils::apply_sysctl_value(
+                                        format!("/proc/sys/net/ipv6/conf/{}/forwarding", i)
+                                            .as_str(),
+                                        "0",
+                                    )?;
+                                }
+                            }
+                        };
+                        continue;
+                    }
                     let id_network_hash = CoreUtils::create_network_hash(net_name, MAX_HASH_SIZE);
                     let sn = SetupNetwork {
                         net: network.clone(),
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -90,7 +90,6 @@ impl Core {
         //we have the bridge name but we must iterate for all the available gateways
         for (idx, subnet) in network.subnets.iter().flatten().enumerate() {
             let subnet_mask_cidr = subnet.subnet.prefix_len();
-
             if let Some(gw) = subnet.gateway {
                 let gw_net = match gw {
                     IpAddr::V4(gw4) => match ipnet::Ipv4Net::new(gw4, subnet_mask_cidr) {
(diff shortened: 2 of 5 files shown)


Luap99 (commented):
Just not setting firewall rules in not enough. We have to make sure that the container ips cannot reach a host ip. I think you can achieve this by setting the `ipv4/conf/<bridge_name>/forwarding` to 0. Just try to ping your hosts ip address from the container, it should fail. 

Also please add a test.

baude:
> Just not setting firewall rules in not enough. We have to make sure that the container ips cannot reach a host ip. I think you can achieve this by setting the `ipv4/conf/<bridge_name>/forwarding` to 0. Just try to ping your hosts ip address from the container, it should fail.

The veth is not getting an IP address and is therefore not able to route when using podman.  You think we are doing something special in podman ?

Luap99:
Are you talking about the bridge interface?

We have to make sure that containers on an internal network cannot reach anything else.
CNI does this by not setting the gateway ip address on the bridge, however as @mheon and I talked we cannot do this since we need an ip address on the bridge in order for the dns server to work with internal networks.

So now we have to options left, set iptables rules to block traffic. This is bad since a firewalld reload will wipe the rules and allo communication. 
The other options is do disable routing for this interface so that the kernel will not route the traffic to the outside. I have not tested if this works but I hope so.

Luap99:
OK I just checked setting the above mentioned sysctl works as expected, so we can have use the bridge ip for the dns server but still block routing.

Btw I think you should also skip the default route in the netlink code

baude:
ok, all reworked and test added.

src/network/core.rs
```diff
@@ -74,36 +74,37 @@ impl Core {
...
-                                    gw6, subnet_mask_cidr, err
-                                ),
-                            ))
-                        }
-                    },
-                };
-
-                gw_ipaddr_vector.push(gw_net)
+            if !network.internal {
+                if let Some(gw) = subnet.gateway {
```
Luap99, line 78: with that you no longer assign the gateway ip to the bridge interface.
Actually we might have to do this otherwise an internal container could access services from the host which listen on 0.0.0.0 but how can we support dns with that? @mheon WDYT?

test/100-bridge-iptables.bats
```diff
@@ -12,8 +12,28 @@ fw_driver=iptables
     assert "${lines[0]}" "==" "[INFO  netavark::firewall] Using iptables firewall driver" "iptables driver is in use"
 }
 
+@test "$fw_driver - internal network" {
+
+# grab the iptables info before running the test
```
Luap99, line 17: please indent the lines properly

src/network/core.rs
baude: We could "cross that bridge" when we have dns ready to implement?

test/100-bridge-iptables.bats
```diff
@@ -12,8 +12,26 @@ fw_driver=iptables
...
+   assert "default" "!~" "$output" "No default route for internal networks"
+
+   expected_rc=1 run_in_container_netns ping -c 1 10.88.0.1
+
+   run_netavark --file ${TESTSDIR}/testfiles/internal.json teardown $(get_container_netns_path)
+}
+
 @test "$fw_driver - simple bridge" {
-    run_netavark --file ${TESTSDIR}/testfiles/simplebridge.json setup $(get_container_netns_path)
+run_netavark --file ${TESTSDIR}/testfiles/simplebridge.json setup $(get_container_netns_path)
```
Luap99, line 34: also indent here

Luap99 (commented):
LGTM, one small nit
@mheon PTAL

src/commands/setup.rs
```diff
@@ -136,7 +156,7 @@ impl Setup {
                                 firewall_driver.setup_port_forward(spf)?;
                             }
                         }
-                    }
+                    } // end
```
mheon, line 159: Why the comment?

src/network/core.rs
mheon: If we're disabling forwarding already why go the extra step and also not assign an IP?

Is there an expectation that Docker internal networks are unable to access services on the host? If so, this makes sense. If not, we should leave the IP.

Luap99: I just checked what docker is doing. There are just blocking the access via iptables. You can still connect to interfaces on the hosts.
In this case we definitely still need the gateway ip on the bridge in order to communicate with the dns server.

mheon: Alright, let's leave the IP on, then.

baude:
nits cleaned

PR #128, opened 2021-12-12
Remove use of unwraps

baude:
remove use of unwraps where it might lead to a panic.


baude:
there are a couple of ways to do this, I generally used matches.  If reviewers would prefer a different approach, feel free to suggest it.

src/network/core.rs
```diff
@@ -27,7 +27,15 @@ impl Core {
...
         // get bridge name
-        let bridge_name: String = network.network_interface.as_ref().unwrap().to_owned();
+        let bridge_name = match network.network_interface.as_ref() {
+            None => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "no bridge provided".to_string(),
+                ))
+            }
+            Some(i) => (*i).clone(),
```
flouthoc, line 37: LGTM. 

Just one nit and its non-blocking: changing above match to `match &network.network_interface` should let us remove this use of custom `de-referencing` at (*i)

baude:
merge me

flouthoc (approved):

flouthoc:
@baude Merged. Thanks

PR #132, opened 2021-12-14
Use `AsRef<str>` for sysctl bits

cgwalters:
I was randomly looking at another PR, and noticed that in about
half the cases, we heap allocate a `String`, but then use `as_str()`
to borrow it briefly.  In other places we directly use `&str`.
It maximizes flexibility for the caller to use `AsRef<str>`; then
we don't need a `.as_str()` dance, we can just directly pass ownership
of the string into the function.

To pick one other random real world example of this, see:
https://github.com/rust-lang/rust/blob/9bb77da74dac4768489127d21e32db19b59ada5b/compiler/rustc_errors/src/diagnostic_builder.rs#L198

### DIFF
--- a/src/firewall/iptables.rs
+++ b/src/firewall/iptables.rs
@@ -106,7 +106,7 @@ impl firewall::FirewallDriver for IptablesDriver {
             None => {}
             Some(i) => {
                 let localnet_path = format!("net.ipv4.conf.{}.route_localnet", i);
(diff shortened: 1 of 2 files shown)


mheon:
Compile test is complaining, code otherwise LGTM

cgwalters:
Fixed, I think I made that typo accidentally when I went to re-commit to do the signoff

flouthoc (approved):
LGTM

PR #138, opened 2021-12-20
Add port ranges

baude:
Use the ability of ranges with iptables instead of creating multiple
rules for each port.

### DIFF
--- a/test/100-bridge-iptables.bats
+++ b/test/100-bridge-iptables.bats
@@ -12,24 +12,6 @@ fw_driver=iptables
     assert "${lines[0]}" "==" "[INFO  netavark::firewall] Using iptables firewall driver" "iptables driver is in use"
 }
 
-@test "$fw_driver - internal network" {
-   run_in_host_netns iptables -t nat -nvL
-   before="$output"
-
-   run_netavark --file ${TESTSDIR}/testfiles/internal.json setup $(get_container_netns_path)
-
-   run_in_host_netns iptables -t nat -nvL
-   after="$output"
-   assert "$before" == "$after" "make sure tables have not changed"
-
-   run_in_container_netns ip route show
-   assert "default" "!~" "$output" "No default route for internal networks"
-
-   run_in_container_netns ping -c 1 10.88.0.1
-
-   run_netavark --file ${TESTSDIR}/testfiles/internal.json teardown $(get_container_netns_path)
-}
-
 @test "$fw_driver - simple bridge" {
     run_netavark --file ${TESTSDIR}/testfiles/simplebridge.json setup $(get_container_netns_path)
     result="$output"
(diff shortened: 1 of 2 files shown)


test/100-bridge-iptables.bats
```diff
@@ -12,24 +12,6 @@ fw_driver=iptables
     assert "${lines[0]}" "==" "[INFO  netavark::firewall] Using iptables firewall driver" "iptables driver is in use"
 }
 
-@test "$fw_driver - internal network" {
```
Luap99, line 15: Why is this removed?

test/100-bridge-iptables.bats
```diff
@@ -145,3 +127,19 @@ fw_driver=iptables
     setup_sctp_kernel_module
     test_port_fw ip=dual proto=sctp
 }
+
+@test "$fw_driver - port range forwarding ipv4 - tcp" {
+    test_port_fwrange=3
```
Luap99, line 132:
```suggestion
    test_port_fw range=3
```

test/100-bridge-iptables.bats
```diff
@@ -145,3 +127,19 @@ fw_driver=iptables
...
+}
+
+@test "$fw_driver - port range forwarding ipv6 - tcp" {
+    test_port_fw ip=6 range=3
+}
```
Luap99, line 137: Can you also add a test with ip=dual

Luap99 (commented):
The tests are failing, it looks like always the second ncat fails with connection refused.

test/100-bridge-iptables.bats
flouthoc: @baude This is not yet fixed

flouthoc:
@baude I think there is something up with the code. Cause the example you were trying with `-p 8080-8082:80-82` works with `main` branch without this PR. So podman is for sure splitting port forward into seperate entries rather than using `range` field.

I wonder if we only need this PR for `-p 8080-8082:80-82` then it works with `firewall` disabled from the current `main` without any addition.

For `range` part i think there is something up with the code cause tests looked fine to me.

Luap99:
@flouthoc You can run with `--log-level trace`, this will show you the exact config for netavark. I am pretty sure `-p 8080-8082:80-82` creates a range, if not than this is a bug

flouthoc:
@Luap99 I'll try that but could you also confirm if this works for you from `main` branch. `run -it -p 8080:8082:80-82 quay.io/libpod/alpine_nginx` and make sure you switch the backend to `netavark`.

Luap99:
I am on PTO. I have no access to my workstation so I cannot test netavark

flouthoc (changes requested):
@baude There is a tiny logical error or typo in `DNAT` rule `--dport` must accept `host-range`. Following diff should make it work and CI should also pass.

Could you please amend PR with following diff.

```diff
diff --git a/src/firewall/varktables/types.rs b/src/firewall/varktables/types.rs
index 05df4e1..c6884ae 100644
--- a/src/firewall/varktables/types.rs
+++ b/src/firewall/varktables/types.rs
@@ -418,7 +418,7 @@ pub fn get_port_forwarding_chains<'a>(
                 i.protocol,
                 container_ip_value,
                 container_port,
-                i.host_port.to_string(),
+                &host_port
             ),
             None,
         ));
```

flouthoc:
@baude cargo fmt is needed.

flouthoc (approved):

Luap99:
Please restore the internal test and add the ip=dual case as noted in the review e ts

PR #147, opened 2022-01-04
Update to zbus 2.0.0/zvariant 3.0.0

mheon:
The new zbus/zvariant are async, but otherwise very similar in API terms. For now, just use the block_on wrapper to retain prior blocking behavior - I'm unconvinced there's a good reason to yield the thread in this case, anyways.


mheon:
@baude @Luap99 @flouthoc PTAL

Luap99 (commented):
LGTM

PR #149, opened 2022-01-06
Change to making only a single call to setup/tear down

mheon:
This is necessary to simplify operation for port forwarding via firewalld. Instead of requiring two separate calls to add and remove port forward rules for v4 and v6, both are now carried in the struct, which means I don't have to worry about one of my DBus calls messing up the other. For iptables it's mostly a wash, the code gets a slight bit more complex, but not significantly.


mheon:
@baude @Luap99 PTAL

mheon:
@baude @Luap99 @flouthoc PTAL, this is necessary for the firewalld work to continue

baude:
i could make the argument that you should create regression tests for the new checking you do, but lets get this in asap.

LGTM

flouthoc (approved):

flouthoc:
@mheon merged.

PR #155, opened 2022-01-10
Add setup/teardown for port forwarding via firewalld

mheon:
Not quite working yet, but firewalld accepts the config, and that's good enough for me to push this.


mheon:
Stripping WIP, this is working, remaining work is on the firewall side

src/firewall/firewalld.rs
```diff
@@ -60,19 +69,216 @@ impl firewall::FirewallDriver for FirewallD {
...
 
-    fn setup_port_forward(&self, _setup_portfw: PortForwardConfig) -> Result<(), Box<dyn Error>> {
-        todo!();
+    fn setup_port_forward(&self, setup_portfw: PortForwardConfig) -> Result<(), Box<dyn Error>> {
+        // NOTE: There is a serious TOCTOU risk in this function if netavark
+        // is either run in parallel, or is not the only thing to edit this
+        // policy.
+        // Because of Podman's locking, this should be safe in the typical
+        // case.
+        // I don't think there's a safer way, unfortunately.
```
Luap99, line 99: Did you open a issue with firewalld to fix this API. I really do not like the TOCTOU risk here and it complicates the code versus just adding a new port.

Luap99: Also do you have to rewrite this to iptables now?

Luap99:
validate is also not happy

src/firewall/firewalld.rs
mheon: Have not asked yet, but I will.

I'm going to leave this code intact but adjust the existing iptables code to allow iptables to also add appropriate rules to firewalld, if it exists, to allow traffic; then we can hardcode iptables-only for RHEL.

Luap99:
@mheon What is the state here, you have lint issues.

mheon:
Mostly focused on other PRs (getting iptables working w/ firewalld in the picture), but this can still go in despite that. I'll clean up the lint issues and get this ready for merging this afternoon.

mheon:
Now allows iptables backend on systems with firewalld, and makes iptables the default unless explicitly overidden

mheon:
If tests go green, this is ready

mheon:
Tests are complaining because of a string compare. Let me fix...

mheon:
Repushed. @Luap99 @baude PTAL

mheon:
This is ready

Luap99:
Will this work with firewalld 1.1? E.g. should include a version check for firewalld and use the firewalld code when >= 1.1?

mheon:
It should, but 1.1.0 has not been released yet, and I don't want to add a check for something that doesn't exist yet and thus can't be tested.

src/firewall/firewalld.rs
```diff
@@ -60,19 +69,216 @@ impl firewall::FirewallDriver for FirewallD {
...
+                _ => bail!("Port forwarding rule that was not a structure encountered"),
+            }
+        }
+
+        let new_rules = Value::new(port_forwarding_rules);
+        let mut new_policy_config = HashMap::<&str, &Value>::new();
+        new_policy_config.insert("forward_ports", &new_rules);
+
+        // Send the updated configuration back to firewalld.
+        match block_on(self.conn.call_method(
```
flouthoc, line 262: Is there any specific reason to block if api is `async` in nature ? I don't think its a blocker though. Just want to  confirm.

flouthoc:
@mheon There are few `block_on` I am fine if its intentional but if `netlink` part is not dependent on `firewall` stuff then we should have any reasons to block but if its vice-versa then I think its good.

src/firewall/firewalld.rs
mheon: There's no reason for us *to* be async. Everything has to happen in sequence, else we'll be interacting with firewalld objects that don't exist yet. Blocking requests make the most sense for this.

flouthoc: Fair.

flouthoc:
> @mheon There are few `block_on` I am fine if its intentional but if `netlink` part is not dependent on `firewall` stuff then we should have any reasons to block but if its vice-versa then I think its good.

This is already explained by @mheon in above comment.

mheon:
Anyone want to merge?

flouthoc:
LGTM @baude PTAL

flouthoc (approved):

PR #161, opened 2022-01-14
core: configure `DNS nameservers` for containers if needed.

flouthoc:
Netavark does not configures name servers if they are on non-default
networks. As a result they don't resolv queries to IP of gateways and
reach to default IP address.

Following PR is needed for containers which are not on default `podman`
network so `podman can configure appropriate resolver.

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -24,6 +24,7 @@ impl Core {
     ) -> Result<types::StatusBlock, std::io::Error> {
         //  StatusBlock response
         let mut response = types::StatusBlock {
+            dns_server_ips: Some(Vec::<IpAddr>::new()),
             interfaces: Some(HashMap::new()),
         };
         // get bridge name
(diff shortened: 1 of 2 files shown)


flouthoc:
@Luap99 @baude @mheon PTAL. I would need this PR so podman can do the right thing for `containers` on a `network`.

src/network/constants.rs
```diff
@@ -1,5 +1,10 @@
 //Following module contains all the network constants
 
+// If network name is equivalent to default, netavark knows that this is the default network
+// netavark chooses to do certain things for default network differently
+// E.g Does not configures DNS nameservers for default network
+pub static DEFAULT_NETWORK_NAME: &str = "podman";
```
mheon, line 6: I don't think we want to continue this with the Netavark/Aardvark stack. Even the default network should do DNS.

Luap99: Yeah this is wrong. DNS should work the same for everything. We have a dns_enabled field which says if we have to do dns.

flouthoc: ah fair i'll use `dns_enabled` no worries.

flouthoc:
@Luap99 @mheon PTAL

src/network/core.rs
```diff
@@ -379,6 +388,8 @@ impl Core {
         let mut response_net_addresses: Vec<NetAddress> = Vec::new();
         // interfaces map, but we only ever expect one, for response
         let mut interfaces: HashMap<String, types::NetInterface> = HashMap::new();
+        // nameservers which can be configured for this container
+        let mut nameservers: Vec<IpAddr> = Vec::new();
```
Luap99, line 392: We do not support dns for macvlan. The gateway address will in most cases not be a host address so we cannot use that.

PR #162, opened 2022-01-17
dns: Add support for `DNS` with containers via `aardvark`.

flouthoc:
Allows netavark to invoke and create `aardvark` entries on `setup` and
`teardown`. Following PR enables communication with containers on the
same network.


flouthoc:
Allows `netavark` to also consume optional `tmpfs` path via `--config` if nothing is specified it tries to `read/write` config from `/tmp/aardvark`.

src/main.rs
```diff
@@ -29,9 +32,11 @@ fn main() {
     let opts = Opts::parse();
 
     let file = opts.file.unwrap_or_else(|| String::from("/dev/stdin"));
+    // aardvark config directory must be supplied by parent or it defaults to /tmp/aardvark
+    let config = opts.config.unwrap_or_else(|| String::from("/tmp/aardvark"));
```
Luap99, line 36: Why do we have a default? Just keep it as an option and when a network has dns enabled and the config dir is not set return an error.

src/dns/aardvark.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    pub fn delete_entry(self, container_id: String, network_name: String) -> Result<()> {
+        let path = Path::new(&self.config).join(network_name);
+        let mut file = File::open(&path)?;
+        let reader = BufReader::new(&mut file);
+
+        let lines: Vec<_> = reader
+            .lines()
+            .map(|l| l.expect("Couldn't read a line"))
+            .collect();
```
Luap99, line 219: I think this reads the full file into memory, I think we should only read line by line.

```
while read line
  if not contains ID
    write line to new file
```

src/dns/aardvark.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub fn delete_from_netavark_entries(
+        self,
+        network_options: types::NetworkOptions,
+    ) -> Result<()> {
+        let mut modified = false;
+        let container_id = network_options.container_id;
+        for (key, network) in network_options.network_info {
+            if network.dns_enabled {
+                modified = true;
+                self.clone().delete_entry(container_id.clone(), key)?;
```
Luap99, line 248: Why do we have to clone self?

src/main.rs
flouthoc: I just added it till we add support in upstream `podman` for `config` otherwise returning `error` will break `podman`. So this is a placeholder we can get rid of it later.

src/dns/aardvark.rs
flouthoc: Removing entries have lot of trade-offs. We already do lot of `i/o` while writing, I think doing `i/o` for each line while reading could become we can avoid it since in most of the cases `lines` would be few so memory might not become a bottle-neck.

But if it becomes bottleneck we can optimize this later.

src/dns/aardvark.rs
flouthoc: `self` is used again later in this iteration and `self` does not implements `copy` trait.

src/dns/aardvark.rs
```diff
@@ -0,0 +1,256 @@
...
+        let path = Path::new(&self.config).join(entry.network_name);
+        // check if this is the first container in this network
+        if !path.exists() {
+            // create file
+            // and write first line as gateway ip
+            let create = File::create(&path)?; // return error if fails
+            let data: String;
+            if !entry.network_gateway_v4.is_empty() && !entry.network_gateway_v6.is_empty() {
+                data = format!(
+                    "{},{}\n\n",
```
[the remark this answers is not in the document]
flouthoc: This is part of the `spec` sheet I believe.

```
<ip>,<ip>

id   <ip>, <ip> name,alias,alias
```

src/dns/aardvark.rs
```diff
@@ -0,0 +1,256 @@
...
+        } else {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "Invalid pid to notify",
+            ));
+        }
+
+        Ok(())
+    }
+    pub fn commit_entries(self, entries: Vec<AardvarkEntry>) -> Result<()> {
```
baude, line 70: We need to be able to handle "updates" here as well.  For example, network connect, where an alias might be added.

src/main.rs
Luap99: ok

src/dns/aardvark.rs
Luap99: But your code already reads line by line into an vector. Please do not read everything into memory the amount of reads will still be the same.

src/dns/aardvark.rs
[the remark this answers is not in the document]
Luap99: @mheon PTAL

src/dns/aardvark.rs
Luap99: This is not needed. A network connect/disconnect will always change the full network membership never a single alias.

flouthoc: @baude This should be handled by `Setup(` already.

src/dns/aardvark.rs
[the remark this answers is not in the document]
flouthoc: No double break lines is not needed my bad.

src/dns/aardvark.rs
flouthoc: fixed with `&mut`

src/dns/aardvark.rs
flouthoc: fixed now reading file at once.

flouthoc:
@Luap99 @mheon @baude @lsm5 I have tested this PR if `aardvark-binary` exists at `/usr/libexec/podman` dns will work without doing anything extra. PTanotherL

src/dns/aardvark.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        pid
+    }
+
+    pub fn start_aardvark_server_if_not_running(&mut self) -> Result<()> {
+        let aardvark_pid = self.get_aardvark_pid();
+        if aardvark_pid != -1 {
+            // check if pid is running
+            let system = sysinfo::System::new_all();
+            match system.process(sysinfo::Pid::from(aardvark_pid)) {
```
Luap99, line 76: Can we just send `0` as signal to check if the process is alive? `sysinfo` is a huge dependency that we really do not need just to check for a single process

src/dns/aardvark.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        if !Path::new(&self.config).exists() {
+            // silently try to create empty config dir if its not there
+            let _ = fs::create_dir(&self.config);
+        }
+
+        log::debug!("Spawning aardvark server");
+        Command::new(AARDVARK_BINARY[0])
+            .args(["--config", &self.config, "-p", "53", "run"])
+            .spawn()?;
```
Luap99, line 96: I think we want to use systemd-run when available, otherwise we run into the same issue as dnsname/dnsmasq that it will get killed by systemd.
/cc @mheon

src/dns/aardvark.rs
flouthoc: Nice idea `0` is not supported so using `SIGWINCH`

src/dns/aardvark.rs
flouthoc: I am currently running it with `system-d` never seen any failures till now. Does system-d kills in uncertain manner ? Maybe I'll create a followup up PR to fix this then.

mheon: Yes, we want to use systemd-run. Systemd really likes killing processes when user sessions end, on on a number of other conditions, and that could easily take out Aardvark but leave the user's containers running.

flouthoc: @mheon @Luap99 Fixed in latest commit.

src/dns/aardvark.rs
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        pid
+    }
+
+    pub fn start_aardvark_server_if_not_running(&mut self) -> Result<()> {
+        let aardvark_pid = self.get_aardvark_pid();
+        if aardvark_pid != -1 {
+            // check if pid is running
+            match signal::kill(Pid::from_raw(aardvark_pid), Signal::SIGWINCH) {
```
Luap99, line 74: Why SIGWINCH? Just send `0` to check if the process is alive?

Luap99:
nooo to late

src/dns/aardvark.rs
flouthoc: @Luap99 Library prevents me sending `0` it expects `enum` and `0` belongs to `SIGHUP`. https://docs.rs/signal/latest/signal/enum.Signal.html . This must be fixed upstream but `SIGWINCH` works for now.

src/dns/aardvark.rs
```diff
@@ -0,0 +1,307 @@
...
+            }
+        }
+
+        if !Path::new(&self.config).exists() {
+            // silently try to create empty config dir if its not there
+            let _ = fs::create_dir(&self.config);
+        }
+
+        log::debug!("Spawning aardvark server");
+        Command::new("systemd-run")
```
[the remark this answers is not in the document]
flouthoc: hmm let me open a PR for non `systemd` users.

src/dns/aardvark.rs
Luap99: How about only trying to start the server at the end where you send SIGHUP and when it fails start the server. I assume the server reads the config on start anyway.

flouthoc: Yes we can merge `notify` and `startup`  that should work as well. `notify` if already running or `start` the server.

Luap99: yes let's do this

PR #168, opened 2022-01-19
dns: Enables name resolution for `rootless` containers

flouthoc:
* Spawns aardvark server for `rootless` containers using `systemd-run`
within same `scope` and `--user`.

* Also use regular child process for environments where `systemd-run` is
not present.

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -13,6 +13,7 @@ use log::debug;
 use std::collections::HashMap;
 use std::error::Error;
 use std::net::IpAddr;
+use std::path::Path;
 
 const IPV4_FORWARD: &str = "net.ipv4.ip_forward";
 const IPV6_FORWARD: &str = "/proc/sys/net/ipv6/conf/all/forwarding";
@@ -32,7 +33,12 @@ impl Setup {
         }
     }
 
-    pub fn exec(&self, input_file: String, config_dir: String) -> Result<(), Box<dyn Error>> {
+    pub fn exec(
+        &self,
+        input_file: String,
+        config_dir: String,
+        rootless: bool,
+    ) -> Result<(), Box<dyn Error>> {
         match network::validation::ns_checks(&self.network_namespace_path) {
             Ok(_) => (),
             Err(e) => {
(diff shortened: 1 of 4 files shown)


flouthoc:
Wait till https://github.com/containers/common/pull/891 is vendor-ed into podman.

src/main.rs
```diff
@@ -12,6 +12,9 @@ struct Opts {
     /// config directory for aardvark, usually path to a tmpfs.
     #[clap(short, long)]
     config: Option<String>,
+    /// Tells if current netavark invocation is for rootless container.
+    #[clap(short, long)]
+    rootless: Option<bool>,
```
Luap99, line 17: Does this allow just `--rootless`? It should work similar to the cobra go library, e.g. if you have a bool option you can use `--rootless` and also `--rootless=true|false` but not `--rootless true|false`

src/dns/aardvark.rs
```diff
@@ -89,17 +103,41 @@ impl Aardvark {
...
+                    .args([
+                        "-q",
+                        "--user",
+                        "--scope",
+                        AARDVARK_BINARY[0],
+                        "--config",
+                        &self.config,
+                        "-p",
+                        "53",
+                        "run",
```
Luap99, line 119: Please share the basic arguments instead of duplicating them for all calls

src/main.rs
flouthoc: Yes it should support both.

src/commands/teardown.rs
```diff
@@ -41,7 +47,9 @@ impl Teardown {
 
         if Aardvark::check_aardvark_support() {
             // stop dns server first before netavark clears the interface
-            let mut aardvark_interface = Aardvark::new(config_dir);
+            let path = Path::new(&config_dir).join("aardvark-dns".to_string());
+            let mut aardvark_interface =
+                Aardvark::new(path.into_os_string().into_string().unwrap(), rootless);
```
mheon, line 52: `.unwrap()` is unsafe, please avoid.

src/commands/setup.rs
```diff
@@ -199,7 +205,14 @@ impl Setup {
...
         if Aardvark::check_aardvark_support() {
-            let mut aardvark_interface = Aardvark::new(config_dir);
+            let path = Path::new(&config_dir).join("aardvark-dns".to_string());
+            debug!(
+                "aardvark args {} {}",
+                path.clone().into_os_string().into_string().unwrap(),
+                rootless
+            );
+            let mut aardvark_interface =
+                Aardvark::new(path.into_os_string().into_string().unwrap(), rootless);
```
mheon, line 215: Unwrap is unsafe, please avoid

src/dns/aardvark.rs
```diff
@@ -89,17 +103,37 @@ impl Aardvark {
         }
 
         log::debug!("Spawning aardvark server");
-        Command::new("systemd-run")
-            .args([
-                "-q",
+
+        if Aardvark::is_executable_in_path("systemd-run") {
```
mheon, line 107: We could avoid this in favor of using systemd's dbus API - saves us a fork-exec and a lot of problems dealing with their CLI. I won't force you to do it here but we should do it before we release.

src/commands/teardown.rs
flouthoc: fixed.

src/commands/setup.rs
flouthoc: fixed

src/dns/aardvark.rs
flouthoc: Yes. I have added a comment `TODO` here. I'll check if we have api for this then i'll modify this.

mheon:
LGTM

PR #171, opened 2022-01-20
dns: double `fork` aardvark server to preserve its nature like a `daemon` although it will be killed by `podman`.

flouthoc:
* Its important that nature of aardvark server is more like a daemon
so following block ensures that aardvark server keeps on running even
if parent is killed

* `setsid()` ensures that there is no controlling terminal on the child process

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -1,6 +1,10 @@
 use crate::network::types;
 use nix::sys::signal::{self, Signal};
 use nix::unistd::Pid;
+use nix::{
+    sys::wait::waitpid,
+    unistd::{fork, ForkResult},
+};
 use std::collections::HashMap;
 use std::fs;
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


flouthoc:
This PR is not needed for `bats` as `bats` has a different issues. But we should still `double-fork` and keep using `--scope`.

flouthoc:
I think we still need double forking as stated in the PR description.

cc @giuseppe for suggesting the idea of `setsid(` and `double-forking` due to the nature for aardvark as a long server.

giuseppe (approved):
LGTM

Luap99 (commented):
LGTM

flouthoc:
I'm not sure how quickly I can add proper logging in a followup PR so to get it into working state I added @Luap99's patch

flouthoc:
@Luap99 nvm I think we should I get `/dev/null` first I'll do some more testing add a loggin as well. Lets merge this later.

Luap99:
This is no longer relevant

PR #173, opened 2022-01-21
Makefile: install target independent of build

lsm5:
You will need to run `make` and `make install` separately.

We will not need any `-nobuild` targets which should make life a little
easier for packagers.

@flouthoc @baude @mheon @cevich @Luap99 PTAL.


mheon:
LGTM

PR #178, opened 2022-01-24
set stdout/err to /dev/null for aardvark

Luap99:
First refactor the aardvark spawn code a bit to remove duplication.
Second, aardvark should not use stdout/err from netavark. Since aardvark
is a long running process the output streams could be closed when it
tries to write something.

TODO We should setup a log file when we are on debug logging level.


mheon:
LGTM

Luap99:
@flouthoc @baude PTAL

PR #179, opened 2022-01-24
Makefile: fix docs install target

lsm5:

@baude @mheon @flouthoc @Luap99 PTAL


mheon:
LGTM

Luap99:
/lgtm
but we should fix the docs before we ship this

PR #182, opened 2022-01-25
Remove dependency on faccess crate

lsm5:
faccess crate hasn't been updated in over a year and a half now.
The current version available on crates.io fails to build for all arches
and the patch to fix it hasn't been merged yet.

Best to get rid of faccess and use stdlib itself.

@flouthoc @baude @mheon @Luap99 PTAL


lsm5:
@mheon l'd prefer we cut a tag after this is merged, much easier to get through fedora review.

Luap99:
Approach looks good to me but I think you have to tweak a test error message.

Luap99:
OK nervermind, the new code still returns the same error so we are good. LGTM

lsm5:
@Luap99 PTAL again. I negated that entire expression because I realized readonly returns false for writable files. Let me know..

src/network/validation.rs
```diff
@@ -1,9 +1,9 @@
-use faccess::{AccessMode, PathExt};
 use log::debug;
-use std::path::Path;
+use std::fs::File;
 
 pub fn ns_checks(file: &str) -> std::io::Result<()> {
     debug!("{:?}", "Checking network namespace permissions...");
     // TODO check for FS_MAGIC
-    Path::new(&file).access(AccessMode::EXISTS & AccessMode::WRITE)
+    let _ = !File::open(&file)?.metadata()?.permissions().readonly();
```
Luap99, line 7: permisssion.readonly doesn't matter you never use the returned value
The check is performed only by `File::open(&file)?.metadata()?`, The question mark returns the error for you

lsm5: so does it suffice to just stop at getting metadata? Because IIRC, we also needed to check for writability and error otherwise?

Luap99: We can check for writeable but then you have to use the actual return value here and return a error if it is read only

lsm5: ack, let me spend some time to include that. Thanks

Luap99: actually no we cannot check for write permissions, the netns file can be read only it only needs to be a valid fd for the setns call so I think you can stop after metadata

lsm5:
ready for review. PTAL.

src/network/validation.rs
lsm5: ah, I noticed this only now lol. Let me revert my changes.

lsm5: Done, tweaked the debug message as well. PTAL.

lsm5:
ready for review.

Luap99 (commented):
/lgtm

PR #183, opened 2022-01-25
Do not remove all iptables rules on teardown

baude:
We should not be removing the HOSTPORT_MASQ rule from the POSTROUTING table as this can effectively ruin port mapping on other running containers.


mheon:
LGTM

PR #184, opened 2022-01-26
dns: support for container `aliases` if specified from container managers.

flouthoc:
Podman can pass network aliases via `--network-alias` so netavark should read it and pass
relevent config to aardvark

Supports DNS resolution with network aliases.

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -212,11 +212,13 @@ impl Aardvark {
         &mut self,
         container_name: String,
         container_id: String,
+        per_network_opts: HashMap<String, types::PerNetworkOptions>,
         netavark_res: HashMap<String, types::StatusBlock>,
     ) -> Result<()> {
         let entries = Aardvark::netavark_response_to_aardvark_entries(
             container_name,
             container_id,
+            per_network_opts,
             netavark_res,
         );
         self.commit_entries(entries)?;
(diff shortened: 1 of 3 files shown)


flouthoc:
@baude @mheon @Luap99 @lsm5 Should make alias feature work with aardvark.

src/dns/aardvark.rs
```diff
@@ -265,14 +268,27 @@ impl Aardvark {
                                                 container_ip_v6 = container_ip.to_string();
                                             }
 
+                                            // unwrap is safe since we already have statusblock
+                                            // for this network
+                                            let mut name_vector =
+                                                Vec::from([container_name.clone()]);
+                                            if let Some(alias) = &per_network_opts
+                                                .get(&network_name)
```
baude, line 276: remind me, what happens with this call if there is NO aliases key?

flouthoc: @baude if key is not defined `if` would not satisfy so nothing would happen.

Btw I also test this after: https://github.com/containers/aardvark-dns/pull/34 and https://github.com/containers/netavark/pull/188 it is working as expected.

src/dns/aardvark.rs
```diff
@@ -265,14 +268,27 @@ impl Aardvark {
...
+                                                .unwrap()
```
mheon, line 277: Probably safer to `unwrap_or_else`

flouthoc: @mheon We generate status block via per_network_opts for which are already doing a check. So its not likely that we will get `None` here, i also added a comment https://github.com/containers/netavark/pull/184/files#diff-6307ba066c1817023f2a53b2c7c6a64d35bde6cf7f443cf5b07f34b9f163513bR271

But I am okay if you still like a handling here.

mheon: Seems safest, in case the rest of the code changes around us

flouthoc: removed unwrap() in latest commit.

Issue #186, opened 2022-01-26
check buildability for all fedora arches in CI


lsm5:
We recently had netavark build fail on non-intel arches because of an issue in the faccess crate (recently gotten rid of). 

Going forward, every PR should be checked for buildability for all the arches supported by Fedora. I imagine something like altbuild (or crossbuild) tasks for podman could be replicated here.

/cc @cevich

PR #187, opened 2022-01-26
[WIP] Cirrus: Add cross-compile task

cevich:
Fixes https://github.com/containers/netavark/issues/186


mheon:
Your new test is red. I think it's an issue with the test?

cevich:
> Your new test is red. I think it's an issue with the test?

It is, somehow.  The `cross` tool is suppose to use podman to do the build, but apparently needs rustup as well :confounded: 

IIRC (need to check) rustup isn't available in Fedora or there was a good reason I didn't want to install it.

cevich:
[Oof](https://cirrus-ci.com/task/5480194914910208?logs=main#L100) `error[E0463]: can't find crate for 'std'`

@lsm5 Buddy! Pal! :grin: I'm googling around and seeing lots of complaints about "Can't rust cross-compile on Fedora..." and then without `rustup` (no package), I can't use [the `cross` tool](https://github.com/cross-rs/cross) either :confused:  [According to these docs](https://rustrepo.com/repo/japaric-rust-cross-rust-embedded#requirements) the main limiting factor for Fedora seems to be lack of available cross-compiled `rust-std` RPMs(?).  I dunno. It's all very confusing to a `bash` guy :grinning: 

Or maybe does @Luap99 or @mheon have any suggestions?

lsm5:
@cevich a bunch of the vendored dependencies in `vendor` have a `$DEP/ci/miri.sh` which seems to curl rustup components.

cevich:
Thanks for checking and for the suggestion.  I'll probably go that way when I have time to work on this next week.

contrib/cirrus/setup.sh
```diff
@@ -18,3 +18,22 @@ msg "************************************************************"
...
+    cd $CARGO_HOME/bin/
+    curl --fail --silent --location -O --url \
+        "https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init"
+    chmod +x rustup-init
+    # complete_setup() will take care of updating $PATH properly
+    showrun ./rustup-init -y \
+        --default-toolchain "$rust_version" \
+        --no-modify-path \
+        --target $(tr ' ' ','<<<"$CROSS_TARGETS")
+fi
```
cevich, line 39: @lsm5 and @Luap99 PTAL, does this seem really fragile to you?  (it does to me)

WDYT if instead of using the Fedora RPMs at all, I installed upstream rust into the VMs at image build time?

I think as long as we pick some static rust toolchain version equal-or-less than the current Fedora version, it's reasonably safe?  And we gain access to `rustup` (unavailable in Fedora).

A down-side could be...it will probably make building netavark/aardvark RPMs and SRPMs harder (since we're using non-RPM toolchain and dependencies).  Possibly making development harder (need to install custom toolchain).  OTOH, this is ***upstream*** testing, so perhaps staying in sync with the distro. is less important?

Opinions?

cevich:
```/usr/bin/ld: ...cut...: error adding symbols: file in wrong format```

Anyone know what I'm missing?  Is there a special cross-compile linker I need to install?

File: contrib/cirrus/runner.sh [hunk 2]
```diff
@@ -45,6 +45,15 @@ _run_verify_vendor() {
     fi
 }
 
+_run_cross() {
+    local cross_target
+    req_env_vars CROSS_TARGETS
```
flouthoc, line 50: @cevich You'd also need to remove `.cargo` directory

flouthoc:
@cevich otherwise host would need something like `arm-linux-gnueabihf-gcc` ,  https://github.com/cross-rs/cross

cevich:
@flouthoc Ugh...another PITA with using `cross` ([log](https://cirrus-ci.com/task/6480011359158272)):  The images come from the docker hub (which is rate-limited for anonymous pulls).  I'll have to setup the host to use the `gcr.io` proxy (not always reliable)...

cevich:
...I got cross-rs working manually with `hack/get_ci_vm.sh`, but with some major caveats:  Editing the default `/etc/containers/registries.conf` is needed to avoid short-name resolution problems (all the default/necessary containers are listed by short-names).  Or (and we may have to do this) we need a [custom `Cross.toml` that specifies fully-qualified images names for each target](https://github.com/cross-rs/cross#custom-docker-images) (see below).

The other headache is we can't use the `mirror.gcr.io` proxy for the docker hub (it's rate-limited) because it can't handle non-x86_64 manifest-lists (which these images appear to be?).  So we'd need to manually mirror each of the images over under Quay somewhere, and use `Cross.toml` to point at them. Ugh.

As an alternative to pulling container images every time, I could cache them in the VM image at build time, but that will further complicate the situation by adding another "touch-point" for maintenance :confounded: 

Lastly, whatever we do needs to be easily maintainable by someone other than me.  It seems like using cross-rs means, adding/changing a target will involve:
1. Mirror the new container image under Quay.
2. Update `.cirrus.yml` to include the new target name.
3. Add new target and image reference to `Cross.toml`.

In short, @flouthoc I don't think cross-rs is a workable solution, we need something simpler.

> otherwise host would need something like arm-linux-gnueabihf-gcc , https://github.com/cross-rs/cross also supports podman so it should be easy to setup.

@lsm5 I previously found/installed a bunch of `gcc-*-<arch>` and `binutils-*-<arch>` packages, but `cargo build --target <aarch64>` still fails with [same/similar linker failure.](https://github.com/containers/netavark/pull/187#issuecomment-1030305229).  Do you have any idea what `arm-linux-gnueabihf-gcc` is and if/where/why/how we can get it for Fedora?

At this point, the cross-task is munging up the VM so thoroughly at runtime, I'm not opposed to just downloading a binaries from somewhere.  Certainly pre-caching them at VM Image build time will save runtime, and avoid needing to mess with hand-editing files and coordinating that with manually mirrored container images :nauseated_face:

cevich:
***Rolling back the git-clock to performing the cross-compile directly on the host VM***

cevich:
[I think this](https://github.com/cross-rs/cross/blob/master/docker/Dockerfile.aarch64-unknown-linux-gnu#L27-L32) (taken from the aarch64 cross-rs Dockerfile) is key:

```
ENV CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc \
    CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER="/linux-runner aarch64" \
    CC_aarch64_unknown_linux_gnu=aarch64-linux-gnu-gcc \
    CXX_aarch64_unknown_linux_gnu=aarch64-linux-gnu-g++ \
    QEMU_LD_PREFIX=/usr/aarch64-linux-gnu \
    RUST_TEST_THREADS=1
```

cevich:
@flouthoc this is the PR for https://issues.redhat.com/browse/RUN-1423  I think I've got the automation part mostly correct and working here.  Based on what I found digging in the cross-rs Dockerfiles, I think all that might be needed is setting up some arch-specific env. vars or configuration values.

Otherwise, if you really want to go the cross-rs way, I gave it a try in [2b30854](https://github.com/containers/netavark/commit/2b30854a7872e91d5a9d71d335f08d188706f1ed) ([cirrus logs](https://github.com/containers/netavark/runs/5146058458)) and made [some notes above](https://github.com/containers/netavark/pull/187#issuecomment-1035420054).  We'll need to customize the short-name config., and also copy the needed docker hub images manually over to quay (I can give you access).  IIRC, there are env. vars. that can be set at runtime to tell cross-rs the FQIN to use (i.e. from quay).  Then someone needs to maintain our quay copies of the images until the end of time :cry:

flouthoc:
@cevich ackd. I think we can build all images needed by `cross` on CI and use it locally (https://github.com/cross-rs/cross/blob/master/build-docker-image.sh) .I think in `netavark/aardvark` case CI trigger limit might not be too much so we might not get rate-limited but I sure we can move images to `quay.io` for a better guarantee later on.

I have opened a new PR and will try out some stuff there lets see how it goes

cevich:
The docker hub pull limit is tracked by egress from the GCE network gateway, not the VMs.  My experience is the limit gets hit ***MUCH** faster than you'd guess :disappointed:   Since the `cross` job is success-blocking, hitting the limit even once will drive developers crazy.

Building images is an innovative idea, though duplicating it constantly at test-time presents lots of opportunities for flakes.  Maybe it could be done ahead of time, say in [c/automation_images](https://github.com/containers/automation_images) and pushed to quay from there?  Or simply mirroring the docker-hub images could be done there as well.  Another idea could be pulling the required images into [the VM at build time](https://github.com/containers/automation_images/blob/main/cache_images/fedora-netavark_packaging.sh)(script), so they're ready to be used on boot.

Anyway...poke me if you get stuck or need any more pointers.

cevich:
Ref: https://github.com/containers/netavark/pull/252

PR #188, opened 2022-01-26
core: populate default search domain in `/etc/resolv.conf`

flouthoc:
Podman setup expects a default search domain in `/etc/resolv.conf` i.e
`dns.podman` lets supports that.

### DIFF
--- a/src/network/constants.rs
+++ b/src/network/constants.rs
@@ -1,5 +1,8 @@
 //Following module contains all the network constants
 
+// default search domain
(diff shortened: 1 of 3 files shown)


flouthoc:
@Luap99 Populates `/etc/resolv.conf` with default `search dns.podman` if dns is enabled or podman is connected to network.

flouthoc:
Lets wait for https://github.com/containers/aardvark-dns/pull/34 before we merge this, so all upstream are always in working state.

PR #193, opened 2022-01-27
Fix lint problems with rust v1.58

Luap99:
Unnecessary use of to_string() throws an lint error since v1.58 so we
should fix that.


lsm5 (approved):
LGTM

PR #194, opened 2022-01-27
Bump to v1.0.0-rc1

mheon:
As the title says

### DIFF
--- a/test/001-basic.bats
+++ b/test/001-basic.bats
@@ -7,11 +7,11 @@ load helpers
 
 @test "netavark version" {
     run_netavark --version
-    assert "netavark 0.0.1" "expected version"
+    assert "netavark 1.0.0-dev" "expected version"
 
     run_netavark version
     json="$output"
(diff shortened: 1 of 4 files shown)


flouthoc:
@mheon There are two commits but first one overrides the second. Is this intentional ?

lsm5 (approved):
LGTM

mheon:
@flouthoc Yes. The first commit is tagged as v1.0.0-rc1, and I'll push the tag once merged. The second bumps us back to `-dev` for continued development work.

test/001-basic.bats
```diff
@@ -7,7 +7,7 @@ load helpers
 
 @test "netavark version" {
     run_netavark --version
-    assert "netavark 0.0.1" "expected version"
+    assert "netavark 1.0.0-dev" "expected version"
```
Luap99, line 10: There is another version check below which has to be changed

PR #198, opened 2022-01-28
Makefile: allow installation of debug binaries

lsm5:
Distro packaging needs unstripped binaries to create its debuginfo
packages.

@flouthoc @mheon @baude @Luap99 PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ docs: ## build the docs on the host
 
 .PHONY: install
 install:
-	install ${SELINUXOPT} -D -m0755 bin/netavark $(DESTDIR)/$(LIBEXECPODMAN)/netavark
+	install ${SELINUXOPT} -D -m0755 bin/netavark$(if $(debug),.debug,) $(DESTDIR)/$(LIBEXECPODMAN)/netavark
 	$(MAKE) -C docs install
 
 .PHONY: uninstall


lsm5:
@cevich will this cause any trouble in CI?

Luap99:
The release binary is still contains all the debug info AFAIK. If I run strip on the release binary it shrinks by almost 50% so I would think it is not striped by rust. Do you have any source why you would need the debug binary?

cevich:
> will this cause any trouble in CI?

I don't think so, and especially not if you're talking about RPMs.  RN all operations are using builds from the CI system - given how much churn there is here and in aardvark.  We can/should move downstream CI to use packages only after the churn dies down, and packages have been released in "the wild" for a little bit (few weeks).  Otherwise downstream testing will suffer from the delay in waiting for packages to build (it's also another point of potential failure/flakes).

More you your question though...CI is currently archiving both the "normal" binary and a separate `.debug` binary.  So if the "normal" binary is stripped, it's probably fine.  If downstream debugging is needed, there's a way it can be told to switch to using the `.debug` binary.  So we're all set there I think, so long as both are produced.

HTH

lsm5:
> The release binary is still contains all the debug info AFAIK. If I run strip on the release binary it shrinks by almost 50% so I would think it is not striped by rust. Do you have any source why you would need the debug binary?

Ack. rpmbuild refused to generate debuginfo package for me with `--release`, while it worked with `debug`. Anyway, let me get back on this. There's another lingering issue that's causing koji builds to fail though it works locally.

flouthoc:
@lsm5 LGTM but i guess we are waiting for you to confirm @Luap99 's  doubt right.  Same goes for : https://github.com/containers/aardvark-dns/pull/44

lsm5:
Thanks all. I tried with and without debug flag. Only thing that comes to mind is that rpmbuild isn't able to strip the release binary. And, stripping the debug binary only reduces size from 91M to 16M which isn't right for shipping in rpms anyway. I'm gonna close this and the aardvark PR.

PR #199, opened 2022-01-28
start aardvark only after we write config

Luap99:
This saves startup time since aardvark does not have to read the same
config twice. Also improve the error handling. We should not ignore errors
such as EPERM, etc...


flouthoc:
@Luap99 Could you please do cargo fmt.

baude:
Being replace by #200

PR #209, opened 2022-02-03
no more panics

Luap99:
We should never ever panic. Podman should always present a human
readable error string. If netavark panics this will not be possible.


Luap99:
@mheon @baude @flouthoc PTAL

baude:
yup, LGTM

flouthoc:
LGTM
/lgtm

PR #211, opened 2022-02-04
bumping vendered dependancies

baude:
updating versions of vendored code as identified by dependabot


contrib/cirrus/runner.sh
```diff
@@ -46,10 +48,12 @@ _run_verify_vendor() {
 }
 
 _run_unit() {
+	cargo vendor
```
Luap99, line 51: Why do you want to run this  for unit and integration tests?

Cargo.toml
```diff
@@ -21,22 +21,22 @@ ipnet = { version = "2", features = ["serde"] }
...
-zvariant = "3.1.1"
-sha2 = "0.10.0"
+tokio = { version = "1.16.1", features = ["full"] }
+zvariant = "3.1.2"
+sha2 = "0.10.1"
 netlink-packet-route = "0.8.0"
 
 [build-dependencies]
-vergen = { version = "=6.0.0", default-features = false, features = ["build", "rustc", "git"] }
+vergen = { version = "=6.0.2", default-features = false, features = ["build", "rustc", "git"] }
```
Luap99, line 41: Remove the equal sign, I had to lock the version because the other version would not compile on rust v1.57. This should be fixed with 6.0.2.

Cargo.toml
```diff
@@ -21,22 +21,22 @@ ipnet = { version = "2", features = ["serde"] }
 iptables = "0.4.3"
```
Luap99, line 21: Can you also bump iptables to 0.5

contrib/cirrus/runner.sh
baude: it looks like there is some vendoring difference when comping with Fedora's rust toolchain vs upstream.  For whatever reason, cargo vendor puts things in order.

baude: i have one more idea to try ...  hang tight

baude:
good to go @Luap99 , thanks for the review

mheon:
LGTM

PR #214, opened 2022-02-08
cleanup: simplify code

Luap99:
Use .map() to iterate through mac address.
Get macvlan master_interface in a single match block.
Remove unneeded .to_string() calls.


baude:
LGTM

PR #215, opened 2022-02-08
Update rtnetlink dependency

Luap99:
Update to the latest version instead of using main.
There are some breaking changes which needed adjustment,
set_name_filter() was moved to match_name(). This changed the behaviour
when the interface is not found, it now returns ENODEV so we have to
handle this instead of OK(None).
The good thing about this is that we only query a single interface.


mheon:
LGTM

PR #216, opened 2022-02-08
fix bridge teardown logic

Luap99:
I think the upgrade to rtnetlink broke the connected interfaces detection.
I added a test to ensure it works now correctly.


baude:
lgtm

PR #220, opened 2022-02-11
fix aardvark output logic

Luap99:
When we have debug level we want stderr connected. If this is not the
case we should use /dev/null.

Also do not connect stdin and use Stdio::inherit() instead of the unsafe
call.


Luap99:
@flouthoc @mheon @baude @rhatdan @vrothberg PTAL This should unblock podman CI.

PR #224, opened 2022-02-14
Update serde_json and rand

baude:


flouthoc (approved):
LGTM

flouthoc:
@baude @mheon @Luap99 @lsm5 Should we have `podman` tests in this CI as well. Cause i believe there can be instances where vendoring changes passes `netavark` CI but it can cause hiccups with `podman`.

Although in this PR none of the `vendor` seems breaking to me. But some components like `netlink/tokio/iptables` in `netavark` and `tokio`,`trust-dns` in `aardvark`. They seem to pass `netavark/aardvark` CI but they can cause issues with `podman`.

baude:
yes, we need to discuss this as a team however.

PR #226, opened 2022-02-16
do not remove NETAVARK_FORWARD rule on teardown 

Luap99:
When we have two containers on different networks running and stop one
of them we still need the NETAVARK_FORWARD rule in the FORWARD chain.

If this rule is removed the containers lose connections if the default
FORWARD policy is DROP.

test: add iptables rule check

Add tests to check the exact iptables rules.


Luap99:
@baude @mheon PTAL

mheon:
LGTM

PR #242, opened 2022-02-22
Remove vendor directory

baude:
From now on, the vendor directory will no longer be part of the github
repository.  For distribution maintainers, we will publish the vendor
directory in the form of a tar.gz file with the release.

For the Fedora spec, something like this will to be done at a minimum:

+Source1: %{url}/releases/download/v%{version}/%{name}-%{version}-vendor.tar.gz


baude:
review and merge please.

@lsm5 PTAL

lsm5:
@baude how will the vendored tarball be generated? Does github publish that automatically from `make vendor` or will that be a manual step?

baude:
@lsm, make vendor and tar that up ... we will do it with releases.

lsm5 (approved):
LGTM

contrib/cirrus/runner.sh
```diff
@@ -19,6 +19,7 @@ _run_noarg() {
 _run_build() {
     # Assume we're on a fast VM, compile everything needed by the
     # rest of CI since subsequent tasks may have limited resources.
+	make vendor
```
mheon, line 22: Do we need these? Shouldn't Cargo automatically fetch deps?

baude: they were needed or the sources were not present.

PR #243, opened 2022-02-24
use local cargo dependencies

Luap99:
We should not try to build with deps from the vendor directory locally.
This will force me to run make vendor manually if I pull in new deps.

When distros want to build with bundeld deps they should a) download
the linked vendor tarball which we should provide for the releases and
b) create the cargo config since to let cargo use the vendor dir.
```
tar xvf %{SOURCE1}
mkdir -p .cargo

cat >.cargo/config << EOF
[source.crates-io]
replace-with = "vendored-sources"

[source.vendored-sources]
directory = "vendor"
EOF
```

see https://src.fedoraproject.org/rpms/rust-coreos-installer/blob/rawhide/f/rust-coreos-installer.spec#_86

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,0 @@
-# Use vendor for everything
-[source.crates-io]
-replace-with = "vendored-sources"
-
-[source.vendored-sources]
-directory = "vendor"
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,5 @@ target/
 targets/
 *.swp
 netavark.1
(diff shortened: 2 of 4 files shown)


Luap99:
@baude @mheon @flouthoc @lsm5 PTAL

mheon:
LGTM

Luap99:
force pushed with changes to DISTRO_PACKAGE.md

flouthoc:
@Luap99 I think we need this at aardvark as well, would you like to open a PR there as well ?

PR #246, opened 2022-02-25
macvlan: support `--internal` `macvlan` network with no default gateway in routes

flouthoc:
Higher level container managers like `podman` can choose to create `macvlan` networks
with no default gateway by setting `--internal` as `true` in netavark
spec.

This can allow other tools and users to create macvlan networks without
default gateway routes but container manager like `podman` must allow
`macvlan` with `--internal` flag.

Netavark Implementation for: https://github.com/containers/podman/issues/13319

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -398,35 +398,38 @@ impl Core {
         // prepare a vector of static aps with appropriate cidr
         for (idx, subnet) in network.subnets.iter().flatten().enumerate() {
             let subnet_mask_cidr = subnet.subnet.prefix_len();
-            if let Some(gw) = subnet.gateway {
-                let gw_net = match gw {
-                    IpAddr::V4(gw4) => match ipnet::Ipv4Net::new(gw4, subnet_mask_cidr) {
-                        Ok(dest) => ipnet::IpNet::from(dest),
-                        Err(err) => {
-                            return Err(std::io::Error::new(
-                                std::io::ErrorKind::Other,
-                                format!(
-                                    "failed to parse address {}/{}: {}",
-                                    gw4, subnet_mask_cidr, err
-                                ),
-                            ))
-                        }
-                    },
-                    IpAddr::V6(gw6) => match ipnet::Ipv6Net::new(gw6, subnet_mask_cidr) {
-                        Ok(dest) => ipnet::IpNet::from(dest),
-                        Err(err) => {
-                            return Err(std::io::Error::new(
-                                std::io::ErrorKind::Other,
-                                format!(
-                                    "failed to parse address {}/{}: {}",
-                                    gw6, subnet_mask_cidr, err
-                                ),
-                            ))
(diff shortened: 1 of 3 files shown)


flouthoc:
@Luap99  @mheon I think this can go in irrespective of podman, we should only need to remove validation for podman.

src/network/core.rs
```diff
@@ -520,6 +528,12 @@ impl Core {
                 for ip in &netns_ipaddr {
                     netns_ipaddr_clone.push(*ip)
                 }
+
+                let mut macvlan_gateway = gw_ipaddr;
+                if internal {
+                    // force empty gateway if macvlan is marked as internal
+                    macvlan_gateway = Vec::new();
+                }
```
Luap99, line 536: This should just be added to line 406 - 435, no reason to overwrite here.

flouthoc: yes lets do it there.

Luap99 (commented):
LGTM

flouthoc:
This is good to go.

src/network/core.rs
```diff
@@ -398,35 +398,38 @@ impl Core {
...
+            if !network.internal {
+                if let Some(gw) = subnet.gateway {
+                    let gw_net = match gw {
+                        IpAddr::V4(gw4) => match ipnet::Ipv4Net::new(gw4, subnet_mask_cidr) {
+                            Ok(dest) => ipnet::IpNet::from(dest),
+                            Err(err) => {
+                                return Err(std::io::Error::new(
+                                    std::io::ErrorKind::Other,
+                                    format!(
+                                        "failed to parse address {}/{}: {}",
```
baude, line 411: can you add something to denote that the address is macvlan so associating the error is easier

src/network/core.rs
```diff
@@ -398,35 +398,38 @@ impl Core {
...
+                                ))
+                            }
+                        },
+                        IpAddr::V6(gw6) => match ipnet::Ipv6Net::new(gw6, subnet_mask_cidr) {
+                            Ok(dest) => ipnet::IpNet::from(dest),
+                            Err(err) => {
+                                return Err(std::io::Error::new(
+                                    std::io::ErrorKind::Other,
+                                    format!(
+                                        "failed to parse address {}/{}: {}",
```
baude, line 423: same as ipv4 comment

src/network/core.rs
```diff
@@ -398,35 +398,38 @@ impl Core {
...
-                                ),
-                            ))
-                        }
-                    },
-                };
-
-                gw_ipaddr_vector.push(gw_net)
+            // Only add gateway to route if macvlan is not marked as an internal network
+            if !network.internal {
+                if let Some(gw) = subnet.gateway {
```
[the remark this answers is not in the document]
flouthoc: I don't think we can combine another `not` let statement here.

PR #252, opened 2022-03-03
build: run build sanity for various architecture

flouthoc:
Performs builds across various archs to make sure build sanity is
maintained across archs

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,7 +17,7 @@ env:
     # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
     SCRIPT_BASE: "./contrib/cirrus"
     FEDORA_NAME: "fedora-35"
-    IMAGE_SUFFIX: "c4560539387953152"
+    IMAGE_SUFFIX: "c6118659411148800"
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
 
 
@@ -113,6 +113,15 @@ unit_task:
   setup_script: *setup
   main_script: *main
 
+build_cross_task:
+  alias: "build_cross"
+  depends_on:
+    - "build"
+  cargo_cache: *ro_cargo_cache
+  targets_cache: *ro_targets_cache
+  bin_cache: *ro_bin_cache
+  setup_script: *setup
+  main_script: *main
 
 integration_task:
   alias: "integration"
@@ -152,6 +161,7 @@ success_task:
   alias: success
   depends_on:
     - "build"
+    - "build_cross"
     - "validate"
     - "verify_vendor"
     - "unit"
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,15 @@ test: unit integration
 build_unit: $(CARGO_TARGET_DIR)
 	cargo test --no-run
 
+# Test build cross-architecture
+.PHONY: build_cross
+build_cross: $(CARGO_TARGET_DIR)
+	cargo install cross
+	rustup target add aarch64-unknown-linux-gnu
+	rustup target add arm-unknown-linux-gnueabi
+	cross build --target aarch64-unknown-linux-gnu
+	cross build --target arm-unknown-linux-gnueabi
+
 .PHONY: unit
 unit: $(CARGO_TARGET_DIR)
 	cargo test
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -35,7 +35,13 @@ EOF
 }
 
 _run_validate() {
+    rustup install stable
+    rustup default stable
     make validate
+}
+
+_run_build_cross() {
+    make build_cross
 }
 
 _run_verify_vendor() {
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -14,7 +14,17 @@ if [[ -r "/etc/ci_environment" ]]; then
 fi
 trap "complete_setup" EXIT
 
+
 msg "************************************************************"
 msg "Setting up runtime environment"
 msg "************************************************************"
+
+dnf -y install make automake gcc gcc-c++ kernel-devel
+
 show_env_vars
+
+if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
+	# Setup short-name for rustembedded/cross
+	# TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
+	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
+fi
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -29,10 +29,9 @@ SECRET_ENV_RE='(ACCOUNT)|(GC[EP]..+)|(SSH)|(PASSWORD)|(TOKEN)'
 if [[ -r "/etc/ci_environment" ]]; then
     source /etc/ci_environment
 else  # set default values - see make_cienv() below
-    # Install rust packages globally instead of per-user
-    CARGO_HOME="${CARGO_HOME:-/usr/local/cargo}"
-    # Ensure cargo packages can be executed
-    PATH="$PATH:$CARGO_HOME/bin"
+    # VM Images are built with this setup
+    CARGO_HOME="${CARGO_HOME:-/var/cache/cargo}"
+    source $CARGO_HOME/env
 fi
 
 # END Global export of all variables
--- a/src/commands/teardown.rs
+++ b/src/commands/teardown.rs
@@ -48,7 +48,7 @@ impl Teardown {
 
         if Path::new(&aardvark_bin).exists() {
             // stop dns server first before netavark clears the interface
-            let path = Path::new(&config_dir).join("aardvark-dns".to_string());
+            let path = Path::new(&config_dir).join("aardvark-dns");
             if let Ok(path_string) = path.into_os_string().into_string() {
                 let mut aardvark_interface = Aardvark::new(path_string, rootless, aardvark_bin);
                 if let Err(er) =
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -45,21 +45,18 @@ impl Aardvark {
 
     // On success retuns aardvark server's pid or returns -1;
     fn get_aardvark_pid(&mut self) -> i32 {
-        let pid: i32;
-        let path = Path::new(&self.config).join("aardvark.pid".to_string());
-        match fs::read_to_string(&path) {
-            Ok(content) => {
-                pid = match content.parse::<i32>() {
-                    Ok(val) => val,
-                    Err(_) => {
-                        return -1;
-                    }
+        let path = Path::new(&self.config).join("aardvark.pid");
+        let pid: i32 = match fs::read_to_string(&path) {
+            Ok(content) => match content.parse::<i32>() {
+                Ok(val) => val,
+                Err(_) => {
+                    return -1;
                 }
-            }
+            },
             Err(_) => {
                 return -1;
             }
-        }
+        };
 
         pid
     }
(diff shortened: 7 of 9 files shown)


flouthoc:
@cevich @baude @lsm5  PTAL. This is not final yet. But I think this should be portable enough that anybody can run `build_cross` on local as well. 

If we ever reach to a point where we are making more than `100` pulls in six hours we can start local builds from `cross-rs` and re-use that or push `arch` images to quay.io but seems highly unlikely for netavark/aardvark.

.cirrus.yml
```diff
@@ -113,6 +113,24 @@ unit_task:
...
+    - "build"
+  env:
+    # Space-separated list of targets to test
+    CROSS_TARGETS: >-
+        x86_64-unknown-linux-gnu
+        aarch64-unknown-linux-gnu
+        powerpc64le-unknown-linux-gnu
+        s390x-unknown-linux-gnu
+        aarch64-unknown-linux-gnu
+        arm-unknown-linux-gnueabi
```
cevich, line 128: IIUC, that's 32bit ARM-7 which will never be supported by podman.

.cirrus.yml
```diff
@@ -113,6 +113,24 @@ unit_task:
...
+  depends_on:
+    - "build"
+  env:
+    # Space-separated list of targets to test
+    CROSS_TARGETS: >-
+        x86_64-unknown-linux-gnu
+        aarch64-unknown-linux-gnu
+        powerpc64le-unknown-linux-gnu
+        s390x-unknown-linux-gnu
+        aarch64-unknown-linux-gnu
```
cevich, line 127: duplicate

Makefile
```diff
@@ -75,6 +75,15 @@ test: unit integration
...
 	cargo test --no-run
 
+# Test build cross-architecture
+.PHONY: build_cross
+build_cross: $(CARGO_TARGET_DIR)
+	cargo install cross
+	rustup target add aarch64-unknown-linux-gnu
+	rustup target add arm-unknown-linux-gnueabi
+	cross build --target aarch64-unknown-linux-gnu
+	cross build --target arm-unknown-linux-gnueabi
```
cevich, line 85: Shouldn't this be using `$CROSS_TARGETS`?

contrib/cirrus/runner.sh
```diff
@@ -35,9 +35,15 @@ EOF
 }
 
 _run_validate() {
+    rustup install stable
+    rustup default stable
```
cevich, line 39: `rustup` is non-standard in the Fedora distribution of rust.  I don't think the `setup.sh` code-path leading to this function includes the upstream-rust install due to `"$CIRRUS_TASK_NAME" != "build_cross"`

contrib/cirrus/setup.sh
```diff
@@ -17,4 +17,26 @@ trap "complete_setup" EXIT
 msg "************************************************************"
 msg "Setting up runtime environment"
 msg "************************************************************"
+
 show_env_vars
+
+if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
+	msg "Installing podman"
+	dnf -y install podman
```
cevich, line 25: I'm pretty sure `podman` is already installed in the `c4801636756357120` image.

contrib/cirrus/setup.sh
```diff
@@ -17,4 +17,26 @@ trap "complete_setup" EXIT
...
+	# Setup short-name for rustembedded/cross
+	# TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
+	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
+	msg "Installing gcc cross-compile packages"
+	err_retry 8 1000 "" dnf install -y \
+	    qemu-user-static \
+	    'gcc-*-linux-gnu' \
+	    'gcc-c++-*-linux-gnu'
```
cevich, line 33: IIUC, this isn't necessary if you're using the cross-rs containers.

contrib/cirrus/setup.sh
```diff
@@ -17,4 +17,26 @@ trap "complete_setup" EXIT
...
+	# This is required by rustup-init (it will complain if any tools are in $PATH)
+	rust_version=$(rpm -q --qf '%{v}' rust)
+	msg "Replacing Fedora native rust toolchain with upstream '$rust_version'"
+	dnf erase -y cargo rust rust-std-static
+	msg "Initializing upstream rust environment."
+	mkdir -p $CARGO_HOME/bin/
+	cd $CARGO_HOME/bin/
+	curl https://sh.rustup.rs -sSf | sh -s -- -y
```
cevich, line 41: FYI: I'm not against abandoning the Fedora-rust toolchain completely, and simply installing upstream rust at VM image build time.

cevich (changes requested):
Good work so far.

contrib/cirrus/runner.sh
cevich: Huh, I really don't understand how this is working.  Maybe I'm wrong and Fedora does include `rustup` now?

flouthoc: @cevich Correct `fedora` does not ships rust with `rustup` I dunno why ? Because of it our toolchain is really old but its ok i guess. We can do it by ourselves :)

contrib/cirrus/setup.sh
flouthoc: @cevich uh I ran it without `podman` first and `cross` was unable to detect any container engine. But `dnf` should skip if its already there i guess.

Makefile
flouthoc: Removed `$CROSS_TARGETS` and keeping all targets in `Make` so testing on local is also easy.

contrib/cirrus/setup.sh
cevich: hmmm....odd....maybe there's a newer image with it.  It's def. in the package list in `automation_images`...

contrib/cirrus/setup.sh
flouthoc: I think this is being discussed in above comment.

cevich: Let's just do this, it will make life much easier.

contrib/cirrus/setup.sh
cevich: Let's do this: https://github.com/containers/automation_images/pull/116

cevich:
Building new VM images w/ podman and upstream rust pre-installed: https://github.com/containers/automation_images/pull/116
With those, you can get rid of the custom rust package juggling in `setup.sh`...

flouthoc:
@cevich I tested this with new image you shared, we need `curl https://sh.rustup.rs -sSf | sh -s -- -y` so it configures proper `sysroot` for `rustc`, If you can do that in image it would be good otherwise I think we are also fine do it for `build_cross` in `setup.sh`

Once https://github.com/containers/automation_images/pull/116 gets merged following PR can be merged.

cevich:
Curl Piping into a shell at runtime really sends shivers down my spine.  Not so much for the security-related issues, but that it means 'something is changing' every time CI runs.   Let's definitely do this at VM build time...it's a simple change...

cevich:
@flouthoc okay, give `c6724066122924032` a try.  I think you can remove [all the rust setup steps for cross now](https://github.com/containers/netavark/pull/252/files#diff-dcdc10b43f652991a881c6d816518d2f1b52fe22f34abb9103f90528dff35a68R27-R33).

contrib/cirrus/setup.sh
```diff
@@ -17,4 +17,18 @@ trap "complete_setup" EXIT
...
+	# Setup short-name for rustembedded/cross
+	# TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
+	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
+	rust_version=$(rpm -q --qf '%{v}' rust)
+	msg "Replacing Fedora native rust toolchain with upstream '$rust_version'"
+	dnf erase -y cargo rust rust-std-static
+	msg "Initializing upstream rust environment."
+	mkdir -p $CARGO_HOME/bin/
+	cd $CARGO_HOME/bin/
+	curl https://sh.rustup.rs -sSf | sh -s -- -y
```
cevich, line 33: All these steps are now built-in to the VM image.

cevich (approved):
LGTM.  Nice work.

cevich:
That's strange, `error: invalid toolchain name: 'usr'`, [I have no idea what that means (log)](https://cirrus-ci.com/task/5542951391592448) :confused:

cevich:
Oh! Woops, I forgot to remove the `cargo` package from the VM image.  Also, nowhere do we set `$RUSTUP_HOME`, does that need to be made part of `make_cienv()` in `lib.sh`?

cevich:
[Building new images.](https://github.com/containers/automation_images/pull/116#issuecomment-1059461111)

flouthoc:
> Oh! Woops, I forgot to remove the `cargo` package from the VM image. Also, nowhere do we set `$RUSTUP_HOME`, does that need to be made part of `make_cienv()` in `lib.sh`?

@cevich I think it should already be set once we remove fedora's default `rust` and install from upstream. We can just `source /var/cache/cargo/env` that should work I think

cevich:
> source /var/cache/cargo/env that should work I think

Since the new toolchain is used for all tasks, all necessary env. vars. need to be set (somehow) from `lib.sh` and written into `/etc/ci_environment` otherwise `hack/get_ci_env.sh` won't work.  That's what the `make_cienv()` + `complete_setup()` functions ensure.  I think we need to do something like:

```diff
diff --git a/contrib/cirrus/lib.sh b/contrib/cirrus/lib.sh
index cc78060..a9cee96 100644
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -29,10 +29,9 @@ SECRET_ENV_RE='(ACCOUNT)|(GC[EP]..+)|(SSH)|(PASSWORD)|(TOKEN)'
 if [[ -r "/etc/ci_environment" ]]; then
     source /etc/ci_environment
 else  # set default values - see make_cienv() below
-    # Install rust packages globally instead of per-user
-    CARGO_HOME="${CARGO_HOME:-/usr/local/cargo}"
-    # Ensure cargo packages can be executed
-    PATH="$PATH:$CARGO_HOME/bin"
+    # VM Images are built with this setup
+    CARGO_HOME="${CARGO_HOME:-/var/cache/cargo}"
+    source $CARGO_HOME/env
 fi

 # END Global export of all variables
@@ -47,7 +46,7 @@ make_cienv(){
     local envname
     local envval
     local SETUP_ENVIRONMENT=1
-    for envname in CARGO_HOME PATH CIRRUS_WORKING_DIR SETUP_ENVIRONMENT; do
+    for envname in CARGO_HOME RUSTUP_HOME PATH CIRRUS_WORKING_DIR SETUP_ENVIRONMENT; do
         envval="${!envname}"
         # Properly escape values to prevent injection
         printf -- "$envname=%q\n" "$envval"
~
```

flouthoc:
@cevich Yes once new image is up, i'll make these changes, I think we don't need `RUSTUP_HOME` though, cause `rustup` init does not sets it up bydefault

flouthoc:
@cevich New image is working perfectly fine, its just missing `build-essentials` I had to install them manually but if you can add these to image as well I'll remove it from `setup.sh` -> `make automake gcc gcc-c++ kernel-devel`. Everything else looks good to me.

cevich:
Looks like the new task takes longer than all the other tasks combined, but that's also understandable, and there's nothing we can do about it either :grin: For reference, this is what the dependency graph looks like:
cirrus

Overall LGTM, thanks for all your efforts on this.

flouthoc:
Oh then this PR must be good to go.

cevich:
Yep, I think it's plenty good enough.  Might be worth having someone who knows rust, check out those specific changes though.

flouthoc:
@Luap99 @baude Could you please review the rust changes, its only to make `rust fmt` and `clippy` happy.

Luap99 (commented):
LGTM

contrib/cirrus/setup.sh
```diff
@@ -14,7 +14,17 @@ if [[ -r "/etc/ci_environment" ]]; then
 fi
 trap "complete_setup" EXIT
 
+
 msg "************************************************************"
 msg "Setting up runtime environment"
 msg "************************************************************"
+
+dnf -y install make automake gcc gcc-c++ kernel-devel
```
cevich, line 22: Ref: Added into package list after this PR merged, will get picked up in next image update: https://github.com/containers/automation_images/pull/117

Issue #260, opened 2022-03-15
use optimistic_dad instead of disabling DAD


Luap99:
Thanks to sbrivio_ on IRC for the pointer

flouthoc:
@Luap99 Could you please add more context to this issue. Like why is there is need to enable `DAD`.

Few cons an

* This might be still slow if something is wrong in our setup.
* [Not a big issue] but older kernels do not support `optimistic_dad` 
* Do we need to turn on router advertisement as well ?

Luap99:
RHEL8 will support this, it was added some time in 3.X

rhatdan:
Great.

PR #261, opened 2022-03-16
core, ipv6: enable `optimistic_dad` instead of completely disabling Duplicate Address Detection

flouthoc:
We are disabling duplicate address detection for `ipv6` completely which
might not be needed so instead of setting `accept_dad` to `0` we can
still do `optimistic_dad` to `1` for required interface.

Some cons of using `optimistic_dad`
* Might be still slower than disabling DAD.
* Not good support for older kernels since it was added in later
  released of `3.x`

Closes: https://github.com/containers/netavark/issues/260


src/network/core_utils.rs
```diff
@@ -799,11 +799,11 @@ impl CoreUtils {
                 }
 
                 if ipv6_enabled {
-                    // Disable duplicate address detection if ipv6 enabled
+                    // Enable optimistic_dad on interface if ipv6 is enabled
                     // Do not accept Router Advertisements if ipv6 is enabled
-                    let br_accept_dad = format!("/proc/sys/net/ipv6/conf/{}/accept_dad", ifname);
+                    let br_optimistic_dad = format!("/proc/sys/net/ipv6/conf/{}/optimistic_dad", ifname);
                     let br_accept_ra = format!("net/ipv6/conf/{}/accept_ra", ifname);
```
flouthoc, line 805: Question not sure about this but do we need to enable `Router Advertisements` i.e `accept_ra` ?

flouthoc:
Caution: This is a draft PR and needs to be experimented on before this can be merged. Follow discussion here: https://github.com/containers/netavark/issues/260

src/network/core_utils.rs
Luap99: I don't know. I think this was copied from cni.

Luap99:
I don't think we need to support such old kernels.

The test failures are exactly the reason why we disabled dad. I guess optimistic dad does not work as hoped.

flouthoc:
@Luap99 Yeah I think it will be still slow and I am still missing the reason why do we need to enable DAD inside container is it to avoid conflicts for the new device addresses ? But nvm I'll check a few things and if this does not works out I'll close the issue.

Luap99:
@flouthoc Can you try setting `use_optimistic` as well.

sbrivio-rh:
Hi,

> @Luap99 Yeah I think it will be still slow and I am still missing the reason why do we need to enable DAD inside container is it to avoid conflicts for the new device addresses ? But nvm I'll check a few things and if this does not works out I'll close the issue.

There are two reasons I suggested optimistic DAD should be used instead of disabling DAD completely:
* if the container's MAC address changes, forced neighbour advertisements are allowed if DAD is enabled, and that will clear/updates neighbour caches. See also https://github.com/moby/moby/pull/42746
* if you have a networking model where NDP crosses the container boundary (I'm not sure netavark supports something like that at the moment, or if it ever will), disabling DAD might lead to a catastrophic collision, albeit in extremely rare cases

I see the netcat test is still failing. I don't think that's because of the delay from optimistic DAD (that's really small, and comparable with address assignment timing), but I can try that out in a simplified scenario with a simple network namespace if it helps.

flouthoc:
Hi @sbrivio-rh, Thank you so much for the explanation. Sure could you please try to reprod this in a smaller env. That would be great help. Meanwhile I'm also trying to look at this.

Thank you for the help :D

PR #262, opened 2022-03-18
[v1.0.1-rhel] Backport "Drop vergen dependency"

Luap99:
Backport commit bdaef37da836.

We only need the version, build time, rust target triple, and git commit
for the version command. These values can easily be generated with little
code so we do not need an extra dependency for this.

The main reason for this change is that vergen will fail if no git
commit is found. On distro build systems this is often the case since
they build from a source tar without the git repo. Instead of erroring
we should just show an empty commit in the version output.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@
 # Main collection of env. vars to set for all tasks and scripts.
 env:
     # Actual|intended branch for this run
-    DEST_BRANCH: "main"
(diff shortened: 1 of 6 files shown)


TomSweeneyRedHat:
Exceptions Approved:   https://bugzilla.redhat.com/show_bug.cgi?id=2065706 and https://bugzilla.redhat.com/show_bug.cgi?id=2065707

PR #265, opened 2022-03-23
aardvark: only use systemd-run when systemd is available

Luap99:
According to sd_booted(3) checking if /run/systemd/system exists is good
enough to see if systemd is booted.
While this check is not perfect it will at least allow us to use it
inside a container where systemd binaries are installed but the
container is run without systemd, see
https://github.com/containers/podman/issues/13533

Another option would be to use the systemd dbus API. I gave it a try but
couldn't figure out the exact format we have to send and also had
trouble connecting to the rootless session.

For now this should help to get aardvark working inside a container.

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -13,6 +13,9 @@ use std::net::Ipv4Addr;
 use std::path::Path;
 use std::process::{Command, Stdio};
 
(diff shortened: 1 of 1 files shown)


Luap99:
@mheon @baude @flouthoc PTAL

flouthoc (approved):
LGTM

flouthoc:
GH is down :|

PR #268, opened 2022-03-25
bidge/macvlan: simplify option parsing

Luap99:
We should try to avoid mutable vars when not needed.
Also add tests for all macvlan modes.


flouthoc (approved):
LGTM, just `cargo fmt` is needed.

PR #269, opened 2022-03-25
bridge/macvlan: use IpNet::new() to create subnet

Luap99:
Simplify the logic a bit by using IpNet:new() which works for ipv4 and
ipv6 instead of having to use a match.


flouthoc:
@Luap99 validate fails with latest toolchain, could you PTAL.

flouthoc (approved):
LGTM

PR #271, opened 2022-03-28
add ipam driver none support

Luap99:
Allow creating interfaces only with no ip addresses assigned.
Also combine some duplicated code for bridge/macvlan.

Ref https://github.com/containers/common/pull/967

### DIFF
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -1,4 +1,4 @@
-use crate::network::constants;
+use crate::network::{constants, internal_types, types};
 use futures::stream::TryStreamExt;
 use futures::StreamExt;
 use libc;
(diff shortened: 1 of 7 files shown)


src/network/core_utils.rs
```diff
@@ -25,6 +25,129 @@ pub struct CoreUtils {
...
+    pub gateway_addresses: Vec<ipnet::IpNet>,
+    pub ipv6_enabled: bool,
+    // result for podman
+    pub net_addresses: Vec<types::NetAddress>,
+    pub nameservers: Vec<IpAddr>,
+}
+
+const IPAM_HOSTLOCAL: &str = "host-local";
+const IPAM_DHCP: &str = "dhcp";
+const IPAM_NONE: &str = "none";
```
flouthoc, line 40: Not a blocker: Should we move these constants to https://github.com/containers/netavark/blob/main/src/network/constants.rs as they might get used by other parts of code in future.

src/network/core_utils.rs
```diff
@@ -25,6 +25,129 @@ pub struct CoreUtils {
     pub networkns: String,
 }
 
+pub struct IPAMAddresses {
```
flouthoc, line 28: This can be moved to https://github.com/containers/netavark/blob/main/src/network/internal_types.rs

flouthoc (approved):
LGTM, just two nits above.

src/network/core_utils.rs
Luap99: yeah good idea

src/network/core_utils.rs
Luap99: done

Luap99:
@flouthoc @mheon PTAL

flouthoc (approved):
LGTM

/lgtm
/approve

PR #272, opened 2022-03-30
[WIP]run podman e2e tests with latest netavark

baude:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,7 +17,7 @@ env:
     # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
     SCRIPT_BASE: "./contrib/cirrus"
     FEDORA_NAME: "fedora-35"
-    IMAGE_SUFFIX: "c6118659411148800"
+    IMAGE_SUFFIX: "c5476394632216576"
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
 
 
@@ -28,8 +28,8 @@ gcp_credentials: ENCRYPTED[d6efdb7d6d4c61e3831df2193ca6348bb02f26cd931695f69d419
 gce_instance:  &standard_vm
     image_project: "libpod-218412"
     zone: "us-central1-c"
-    cpu: 2
-    memory: "4Gb"
+    cpu: 8 
+    memory: "8Gb"
     disk: 200  # GB, do not set <200 per gcloud warning re: I/O performance
     image_name: "${FEDORA_NETAVARK_IMAGE}"
 
@@ -127,6 +127,19 @@ integration_task:
   alias: "integration"
   depends_on:
     - "unit"
+  cargo_cache: *ro_cargo_cache
+  targets_cache: *ro_targets_cache
+  bin_cache: *ro_bin_cache
+  setup_script: *setup
+  main_script: *main
+
+podman_e2e_task:
+  alias: "podman_e2e"
+    #depends_on:
+    #    - "unit"
+  env:
+    GOPATH: &gopath "/var/tmp/go"
+    GOCACHE: &gocache "/var/tmp/go/cache/go-build"
   cargo_cache: *ro_cargo_cache
   targets_cache: *ro_targets_cache
   bin_cache: *ro_bin_cache
--- a/contrib/cirrus/runner.sh
+++ b/contrib/cirrus/runner.sh
@@ -59,6 +59,10 @@ _run_integration() {
     make integration
 }
 
+_run_podman_e2e() {
+	NETWORK_BACKEND=netavark make -C /var/tmp/go/src/github.com/containers/podman localintegration
+}
+
 show_env_vars
 
 msg "************************************************************"
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -19,7 +19,6 @@ msg "************************************************************"
 msg "Setting up runtime environment"
 msg "************************************************************"
 
-dnf -y install make automake gcc gcc-c++ kernel-devel
 
 show_env_vars
 
@@ -27,4 +26,19 @@ if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
 	# Setup short-name for rustembedded/cross
 	# TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
 	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
+fi
+
+if [[ "$CIRRUS_TASK_NAME" == "podman_e2e" ]]; then
+	# build netavark
+	make
+	cp -v bin/netavark /usr/libexec/podman/
+	# clone podman upstream main
+	git clone https://github.com/containers/podman /var/tmp/go/src/github.com/containers/podman
+	rm /usr/bin/podman
+	make -C /var/tmp/go/src/github.com/containers/podman/ install.tools
+	# compile podman
+	make -C /var/tmp/go/src/github.com/containers/podman podman
+	# compile and install aardvark-dns head
+	dnf -y copr enable rhcontainerbot/podman-next
+	dnf -y install aardvark-dns
 fi


.cirrus.yml
```diff
@@ -28,8 +28,8 @@ gcp_credentials: ENCRYPTED[d6efdb7d6d4c61e3831df2193ca6348bb02f26cd931695f69d419
 gce_instance:  &standard_vm
     image_project: "libpod-218412"
     zone: "us-central1-c"
-    cpu: 2
-    memory: "4Gb"
+    cpu: 8 
+    memory: "8Gb"
```
Luap99, line 32: Does this affect all jobs? I think we want to only set this for the e2e test to save some resources/money

Luap99 (commented):
Do we really want to run all e2e test?
I would prefer if we could reduce this somehow. First, running this test will take a long time, second some tests are flaking often so this will get annoying.

.cirrus.yml
cevich: Yes

.cirrus.yml
```diff
@@ -187,7 +199,10 @@ centos9_build_task:
         image: quay.io/libpod/centos9rust
     script:
         - cargo build
-
+podman_e2e_test:
+    alias: podman_e2e
+    depends_on: build
+
```
cevich, line 205: This will do nothing (does not end in `_task` and has no scripts or anything), what is it?

baude: left over turd

contrib/cirrus/setup.sh
```diff
@@ -28,3 +27,19 @@ if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
...
 	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
 fi
+
+if [[ "$CIRRUS_TASK_NAME" == "podman_e2e" ]]; then
+	# build netavark
+	make
+	cp -v bin/netavark /usr/libexec/podman/
+	# clone podman upstream main
+	git clone https://github.com/containers/podman /var/tmp/go/src/github.com/containers/podman
+	rm /usr/bin/podman
```
cevich, line 37: This is ripe for causing headaches.  An easier solution that also avoids compile time is to just grab Lokesh's podman-next COPR and `dnf update -y podman`.

contrib/cirrus/setup.sh
```diff
@@ -28,3 +27,19 @@ if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
...
+	make -C /var/tmp/go/src/github.com/containers/podman/ install.tools
+	# compile podman
+	make -C /var/tmp/go/src/github.com/containers/podman podman
+	# compile and install aardvark-dns head
+	git clone http://github.com/containers/aardvark-dns /var/tmp/go/src/github.com/containers/aardvark-dns
+	make -C /var/tmp/go/src/github.com/containers/aardvark-dns/
+	cp -v /var/tmp/go/src/github.com/containers/aardvark-dns/bin/aardvark-dns /usr/libexec/podman/
```
cevich, line 44: aardvark-dns is also available in podman-next COPR, no need to compile.

baude:
> Do we really want to run all e2e test? I would prefer if we could reduce this somehow. First, running this test will take a long time, second some tests are flaking often so this will get annoying.

Of course not, I'm just tryhing to get it to run at this point before I decide to pare it back.  Btw, most of the team thinks we should :)

contrib/cirrus/setup.sh
```diff
@@ -28,3 +27,18 @@ if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
 	# TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
 	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
 fi
+
+if [[ "$CIRRUS_TASK_NAME" == "podman_e2e" ]]; then
+	# build netavark
+	make
+	cp -v bin/netavark /usr/libexec/podman/
+	# clone podman upstream main
+	git clone https://github.com/containers/podman /var/tmp/go/src/github.com/containers/podman
```
cevich, line 36: idea (don't know it will work):

```bash
cd $GOPATH/src/github.com/containers/podman/
source ./contrib/cirrus/lib.sh
remove_packaged_podman_files
export TEST_ENVIRON=netavark  # set this in .cirrus.yml
install_test_configs
...
```

contrib/cirrus/runner.sh
```diff
@@ -59,6 +59,10 @@ _run_integration() {
     make integration
 }
 
+_run_podman_e2e() {
+	make -C /var/tmp/go/src/github.com/containers/podman localintegration
```
cevich, line 63: I think this needs `export NETWORK_BACKEND=netavark` somewhere

.cirrus.yml
```diff
@@ -133,6 +133,19 @@ integration_task:
...
+podman_e2e_task:
+  alias: "podman_e2e"
+  depends_on:
+    - "unit"
+  env:
+    GOPATH: &gopath "/var/tmp/go"
+    GOPATH: &gocache "/var/tmp/go/cache/go-build"
+  cargo_cache: *ro_cargo_cache
+  targets_cache: *ro_targets_cache
+  bin_cache: *ro_bin_cache
```
cevich, line 145: This will get you a working `/var/tmp/netavark/bin/netavark` binary. 

The latest Aardvark is available by curl & unzip: https://api.cirrus-ci.com/v1/artifact/github/containers/aardvark-dns/success/binary.zip

.cirrus.yml
```diff
@@ -17,7 +17,7 @@ env:
     # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
     SCRIPT_BASE: "./contrib/cirrus"
     FEDORA_NAME: "fedora-35"
-    IMAGE_SUFFIX: "c6118659411148800"
+    IMAGE_SUFFIX: "c5476394632216576"
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
```
cevich, line 22:
```yaml
    FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
````

File: .cirrus.yml (hunk 1 above)
Comment (cevich) [reply .cirrus.yml:32]: k

File: contrib/cirrus/setup.sh [hunk 9]
```diff
@@ -28,3 +27,18 @@ if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
 	# TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
 	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
 fi
+
+if [[ "$CIRRUS_TASK_NAME" == "podman_e2e" ]]; then
+	# build netavark
+	make
```
cevich, line 33: This shouldn't be necessary, the binary should already be in `/var/tmp/netavark/bin/netavark` (because of `bin_cache: *ro_bin_cache`).  If it's not there, something horrible has broken in CI.

.cirrus.yml
```diff
@@ -133,6 +133,19 @@ integration_task:
   setup_script: *setup
   main_script: *main
 
+podman_e2e_task:
+  alias: "podman_e2e"
+    #depends_on:
+    #    - "unit"
```
cevich, line 139: In order for the cache to restore properly, this task should at least depend on `build` at a minimum

edsantiago:
@baude in case you didn't see my IRC comment last week: 
```irc
16:07:05     esm     baude, I don't know if this helps, but one common factor in all the failing tests, the final debug line is "loading seccomp profile". That's where the hang happens.
16:07:29     esm     baude, "loading seccomp profile" also appears in successful tests, but in those cases it is followed by other debug lines
```

contrib/cirrus/runner.sh
```diff
@@ -59,6 +59,10 @@ _run_integration() {
     make integration
 }
 
+_run_podman_e2e() {
+	NETWORK_BACKEND=netavark make -C /var/tmp/go/src/github.com/containers/podman localintegration
```
cevich, line 63: There's a clash over the meaning of `$DEBUG`, the automation libraries use 0 | 1 value but the e2e tests use empty/non-empty:  https://github.com/containers/podman/blob/main/test/e2e/common_test.go#L852  The automation libraries default it's value to '0', that's one reason why you see different behavior using `make localintegration` (by hand) vs `runner.sh` - the later is loading the automation libraries.  So...I think the above line should be:

```
export NETWORK_BACKEND=netavark
export -n DEBUG
make -C /var/tmp/go/src/github.com/containers/podman localintegration
```

edsantiago: This is my annual reminder that noone should ever use `DEBUG` without a namespace prefix.

cevich: Lol...hang on...let me clean off my hindsight spectacles :eyeglasses: Oh! Damnit! :angry:

PR #273, opened 2022-03-31
test/helpers.bash: assert handle corner cases correctly

Luap99:
Right now `assert t:[1] != t:[1]` passes. This is obviously incorrect
and some test might not be working correctly because of this. We have to
special case this like the "==" case.

I discovered this in buildah, since we use the same logic here we should
fix it to prevent unexpected bugs.

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -222,6 +222,11 @@ function assert() {
         if [ "$actual_string" = "$expect_string" ]; then
             return
         fi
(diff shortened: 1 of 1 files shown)


Luap99:
@edsantiago PTAL
same for aardvark https://github.com/containers/aardvark-dns/pull/114

flouthoc (approved):
LGTM
/approve
/lgtm

PR #275, opened 2022-04-05
isolate podman networks

### LINKED ISSUE #154: Look into blocking/permitting network to network communication
See https://github.com/containers/podman/issues/5805

cdoern:
add support for a new option "isolate", a boolean that determines whether or not a network
can send/recieve data outside of its bridge. This is done by creating a new chain "NETAVARK_ISOLATION" that drops external outgoing connections

resolves #154

Signed-off-by: cdoern <cbdoer23@g.holycross.edu>

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -117,10 +117,29 @@ impl Setup {
                         };
                         continue;
                     }
+
+                    // parse isolation option here so we can avoid unwrapping and storing it unecessarily
+                    let isolation_config =
+                        match network.options.as_ref().and_then(|map| map.get("isolate")) {
+                            Some(isolation) => match isolation.parse() {
+                                Ok(isolation) => isolation,
+                                Err(err) => {
+                                    return Err(std::io::Error::new(
+                                        std::io::ErrorKind::Other,
+                                        format!("could not parse isolation option: {}", err),
+                                    )
+                                    .into())
+                                }
+                            },
+                            // default is to not isolate podman networks
+                            None => false,
+                        };
+
                     let id_network_hash = CoreUtils::create_network_hash(net_name, MAX_HASH_SIZE);
                     let sn = SetupNetwork {
                         net: network.clone(),
                         network_hash_name: id_network_hash.clone(),
+                        isolation: isolation_config,
                     };
                     firewall_driver.setup_network(sn)?;
                     let port_bindings = network_options.port_mappings.clone();
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -16,7 +16,7 @@ use std::process::{Command, Stdio};
 const SYSTEMD_CHECK_PATH: &str = "/run/systemd/system";
 const SYSTEMD_RUN: &str = "systemd-run";
 
-#[derive(Debug, Clone)]
+#[derive(Clone, Debug)]
 pub struct AardvarkEntry {
     pub network_name: String,
     pub network_gateway_v4: String,
--- a/src/firewall/iptables.rs
+++ b/src/firewall/iptables.rs
@@ -43,6 +43,17 @@ pub fn new() -> NetavarkResult<Box<dyn firewall::FirewallDriver>> {
 
 impl firewall::FirewallDriver for IptablesDriver {
     fn setup_network(&self, network_setup: SetupNetwork) -> NetavarkResult<()> {
+        let interface = match network_setup.net.network_interface {
+            Some(iface) => iface,
+            None => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "failed to get interface".to_string(),
+                )
+                .into())
+            }
+        };
+
         if let Some(subnet) = network_setup.net.subnets {
             for network in subnet {
                 let is_ipv6 = network.subnet.network().is_ipv6();
@@ -50,15 +61,27 @@ impl firewall::FirewallDriver for IptablesDriver {
                 if is_ipv6 {
                     conn = &self.conn6;
                 }
-                let chains = varktables::types::get_network_chains(
+
+                let chain_result = varktables::types::get_network_chains(
                     conn,
                     network.subnet,
                     network_setup.network_hash_name.clone(),
                     is_ipv6,
+                    interface.to_string(),
+                    network_setup.isolation,
                 );
 
-                for chain in chains {
-                    chain.add_rules()?;
+                match chain_result {
+                    Ok(chains) => {
+                        for c in chains {
+                            c.add_rules()?
+                        }
+                    }
+                    Err(e) => {
+                        return Err(
+                            std::io::Error::new(std::io::ErrorKind::Other, e.to_string()).into(),
+                        )
+                    }
                 }
 
                 add_firewalld_if_possible(&network);
@@ -70,6 +93,17 @@ impl firewall::FirewallDriver for IptablesDriver {
     // teardown_network should only be called in the case of
     // a complete teardown.
     fn teardown_network(&self, tear: TearDownNetwork) -> NetavarkResult<()> {
+        let interface = match tear.config.net.network_interface {
+            Some(iface) => iface,
+            None => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "failed to get interface".to_string(),
+                )
+                .into())
+            }
+        };
+
         // Remove network specific general NAT rules
         if let Some(subnet) = tear.config.net.subnets {
             for network in subnet {
@@ -83,21 +117,32 @@ impl firewall::FirewallDriver for IptablesDriver {
                     network.subnet,
                     tear.config.network_hash_name.clone(),
                     is_ipv6,
+                    interface.to_string(),
+                    tear.config.isolation,
                 );
-                for chain in &chains {
-                    // Because we only call teardown_network on complete teardown, we
-                    // just send true here
-                    chain.remove_rules(true)?;
-                }
 
-                for chain in chains {
-                    match &chain.td_policy {
-                        None => {}
-                        Some(policy) => {
-                            if tear.complete_teardown && *policy == OnComplete {
-                                chain.remove()?;
+                match chains {
+                    Ok(chains) => {
+                        for c in &chains {
+                            // Because we only call teardown_network on complete teardown, we
+                            // just send true here
+                            c.remove_rules(true)?;
+                        }
+                        for c in chains {
+                            match &c.td_policy {
+                                None => {}
+                                Some(policy) => {
+                                    if tear.complete_teardown && *policy == OnComplete {
+                                        c.remove()?;
+                                    }
+                                }
                             }
                         }
+                    }
+                    Err(e) => {
+                        return Err(
+                            std::io::Error::new(std::io::ErrorKind::Other, e.to_string()).into(),
+                        )
                     }
                 }
 
--- a/src/firewall/varktables/types.rs
+++ b/src/firewall/varktables/types.rs
@@ -7,6 +7,8 @@ use crate::network::internal_types::PortForwardConfig;
 use crate::network::types::Subnet;
 use ipnet::IpNet;
 use iptables::IPTables;
+use log::debug;
+use std::error::Error;
 use std::net::IpAddr;
 
 //  Chain names
@@ -178,7 +180,9 @@ pub fn get_network_chains(
     network: IpNet,
     network_hash_name: String,
     is_ipv6: bool,
-) -> Vec<VarkChain<'_>> {
+    interface_name: String,
+    isolation: bool,
+) -> Result<Vec<VarkChain<'_>>, Box<dyn Error>> {
     let mut chains = Vec::new();
     let prefixed_network_hash_name = format!("{}-{}", "NETAVARK", network_hash_name);
 
@@ -215,44 +219,94 @@ pub fn get_network_chains(
     ));
     chains.push(postrouting_chain);
     if !is_ipv6 {
-        // NETAVARK_FORWARD
-        let mut netavark_forward_chain =
-            VarkChain::new(conn, FILTER.to_string(), NETAVARK_FORWARD.to_string(), None);
-        netavark_forward_chain.create = true;
-
-        // Create incoming traffic rule
-        // CNI did this by IP address, this is implemented per subnet
-        netavark_forward_chain.build_rule(VarkRule::new(
-            format!(
-                "-d {} -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT",
-                network
-            ),
-            Some(TeardownPolicy::OnComplete),
-        ));
-
-        // Create outgoing traffic rule
-        // CNI did this by IP address, this is implemented per subnet
-        netavark_forward_chain.build_rule(VarkRule::new(
-            format!("-s {} -j ACCEPT", network),
-            Some(TeardownPolicy::OnComplete),
-        ));
-        chains.push(netavark_forward_chain);
-
         // FORWARD chain
-        // Insert the rule into the first position
         let mut forward_chain = VarkChain::new(conn, FILTER.to_string(), FORWARD.to_string(), None);
+        // if we do not create the chains as we need them with isolation,
+        // the entire process fails since when you finally do get around to creating the chains
+        // within the []chains array, they reference and jump to locations
+        // that do not exist yet, causing errors
+        add_chain_unique(
+            forward_chain.driver,
+            &forward_chain.table,
+            &forward_chain.chain_name,
+        )?;
+
+        // used to prepend specific rules
+        let mut ind = 1;
+
+        debug!("Isolate this network?: {}", isolation);
+        if isolation {
+            // NETAVARK_ISOLATION_1
+            let mut netavark_isolation_chain = VarkChain::new(
+                conn,
+                FILTER.to_string(),
+                "NETAVARK_ISOLATION".to_string(),
+                None,
+            );
+            netavark_isolation_chain.create = true;
+
+            // -A FORWARD -j NETAVARK_ISOLATION_1
+            forward_chain.build_rule(VarkRule {
+                rule: format!("-j {}", "NETAVARK_ISOLATION"),
+                position: Some(ind),
+                td_policy: Some(TeardownPolicy::OnComplete),
+            });
+            ind += 1;
+
+            // NETAVARK_ISOLATION -i bridge_name ! -o bridge_name -j DROP
+            netavark_isolation_chain.build_rule(VarkRule::new(
+                format!(
+                    "-i {} ! -o {} -j {}",
+                    interface_name, interface_name, "DROP"
+                ),
+                Some(TeardownPolicy::OnComplete),
+            ));
+
+            // PUSH CHAIN
+            chains.push(netavark_isolation_chain);
+        }
+
+        // forward needs to be added to "chains" first so its rule are created before NETAVARK_FORWARD. To do this we
+        // need to create the chain so its rules can be referenced
+        add_chain_unique(conn, FILTER, NETAVARK_FORWARD)?;
+
         forward_chain.build_rule(VarkRule {
             rule: format!(
                 "-m comment --comment 'netavark firewall plugin rules' -j {}",
                 NETAVARK_FORWARD
             ),
-            position: Some(1),
+            position: Some(ind),
             td_policy: Some(TeardownPolicy::Never),
         });
         chains.push(forward_chain);
+
+        // NETAVARK_FORWARD
+        let mut netavark_forward_chain =
+            VarkChain::new(conn, FILTER.to_string(), NETAVARK_FORWARD.to_string(), None);
+        netavark_forward_chain.create = true;
+
+        if !isolation {
+            // Create incoming traffic rule
+            // CNI did this by IP address, this is implemented per subnet
+            netavark_forward_chain.build_rule(VarkRule::new(
+                format!(
+                    "-d {} -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT",
+                    network
+                ),
+                Some(TeardownPolicy::OnComplete),
+            ));
+
+            // Create outgoing traffic rule
+            // CNI did this by IP address, this is implemented per subnet
+            netavark_forward_chain.build_rule(VarkRule::new(
+                format!("-s {} -j ACCEPT", network),
+                Some(TeardownPolicy::OnComplete),
+            ));
+        }
+        chains.push(netavark_forward_chain);
     }
 
-    chains
+    Result::Ok(chains)
 }
 
 pub fn get_port_forwarding_chains<'a>(
--- a/src/network/internal_types.rs
+++ b/src/network/internal_types.rs
@@ -17,6 +17,8 @@ pub struct SetupNetwork {
     pub net: types::Network,
     // hash id for the network
     pub network_hash_name: String,
+    // isolation determines whether the network can communicate with others outside of its interface
+    pub isolation: bool,
 }
 
 #[derive(Clone, Debug)]
--- a/test/100-bridge-iptables.bats
+++ b/test/100-bridge-iptables.bats
@@ -89,6 +89,7 @@ fw_driver=iptables
     run_in_host_netns iptables -nvL FORWARD
     assert "${lines[2]}" == "    0     0 NETAVARK_FORWARD  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* netavark firewall plugin rules */" "FORWARD rule"
     run_in_host_netns iptables -nvL NETAVARK_FORWARD
+    # rule 1 should be DROP for any existing networks
     assert "${lines[2]}" == "" "NETAVARK_FORWARD rule 1 is empty"
     assert "${lines[3]}" == "" "NETAVARK_FORWARD rule 2 is empty"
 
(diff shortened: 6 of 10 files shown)


cdoern:
@mheon @Luap99 I could have done this *completely* wrong since its my first time messing with rust, netavark, and network stuff in general. I am basically reading the IPs from aardvark and trying to make iptables rules

flouthoc:
@cdoern This looks really cool ! It would be really nice if we can drop dependency on aardvark cause there can be instances where aardvark is not on the system so these config might not be present.

Two ideas which i could think of are:

* We already create iptables rules for other networks which go through netavark and afaik netavark annotate them with netavark specific comments. Could we read iptables rules back and filter with `netavark` specific comments to get information about other network ? Although you can check first if they have the info which you need or not.

* Modify libimage so it propagates minimal network info of other networks as well while performing setup.

Luap99:
We need to come up with some option to activate/deactivate this behaviour first.
Currently this is allowed so the default should stay to not break users.

@mheon How does this work with you firewalld backend?

src/dns/aardvark.rs
```diff
@@ -7,23 +7,33 @@ use std::fs;
...
+use std::net::Ipv6Addr;
 use std::net::Ipv4Addr;
 use std::path::Path;
 use std::process::{Command, Stdio};
 
 const SYSTEMD_CHECK_PATH: &str = "/run/systemd/system";
 const SYSTEMD_RUN: &str = "systemd-run";
 
-#[derive(Debug, Clone)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
Luap99, line 21: Please do not add Serialize, Deserialize. These types should never be written as json

test/100-bridge-iptables.bats
```diff
@@ -321,3 +322,16 @@ EOF
...
+    result1="$output"
+    assert_json "$result1" 'has("podman")' == "true" "object key exists"
+
+    run_netavark --file ${TESTSDIR}/testfiles/connectbridge.json setup $(get_container_netns_path)
+    result2="$output"
+    assert_json "$result2" 'has("podman1")' == "true" "object key exists"
+
+    # This should fail, right?
+    run_in_container_netns ping -c 1 10.88.0.2
+}
```
Luap99, line 337: This test does not do what it should.
You setup two networks on one container. That should always work. We need to test that we cannot communicate between two containers on different subnets but still make sure that it works on the same network. The test infra currently does not expose a second container but you should be able to add a second namespace manually.

Luap99 (commented):
We need to handle this inside the firewall driver!
Maybe I am missing something but why do you need more information than what is already available?

src/dns/aardvark.rs
```diff
@@ -338,4 +348,87 @@ impl Aardvark {
...
+                     }
+                }
+            }
+            combined_list.insert(gateway_list.clone(), net_list.clone());
+            
+        }
+
+
+
+               // format written to the file was: entry.container_id, entry.container_ip_v4, entry.container_ip_v6, container_names
```
flouthoc, line 396: `cargo fmt` should fix indentation here

cdoern:
@Luap99 I agree we need some enabler/disabler. As far as I can tell there is no persistent record of created podman networks besides the aardvark files. If that is incorrect I can switch it to use wherever those entries are stored.

@flouthoc are the iptables entries per system or per container? can I access them after they are created besides querying aardvark? All of the aardvark functions hat exist seem to be the best way of dealing with existing ips/networks

cdoern:
@Luap99 that is how I would pass the option to enable this, right? Can that also work for passing all existing podman networks?

Luap99:
What existing networks? You should only ever configure rules for the current network you have. Did you look at the cni implementation: https://github.com/containernetworking/plugins/commit/22dd6c553dbb9fd3e34502c0de16dd17c7461a83

mheon:
@Luap99 RE: Firewalld - Honestly, I have no idea if it's possible. To avoid using permanent rules we had to stick everything in the same zone, and zones are usually how you'd achieve isolation. We could probably add rich rules to the zone to accomplish the same thing?

mheon:
I admit I'm a little confused about reusing Aardvark infrastructure; we don't really care about containers, which is what Aardvark mainly deals with, but networks instead.

cdoern:
@mheon @Luap99 I thought the original podman issue was dealing with containers on separate networks, not the same, meaning we will have to block communication between separately created networks right?

Luap99:
Yes you want to block only between different networks but that doesn't mean that you have to know them all. Such a approach is not scalable at all and you cannot ever rely on aardvak info since it can be disabled.
CNI has about the same information as netavark about network configs so you should be able to get it working if you just reimplement this code here. https://github.com/containernetworking/plugins/commit/22dd6c553dbb9fd3e34502c0de16dd17c7461a83

cdoern:
@Luap99 ok that makes more sense I will take another look at that CNI implementation and work from there. Thanks for the feedback

cdoern:
@Luap99 tried to implement this better, still figuring out the testing. Let me know if I'm on the right track

test/helpers.bash
```diff
@@ -66,23 +80,41 @@ function get_container_netns_path() {
...
-    run_helper nsenter -n -t $HOST_NS_PID "$@"
+    if [ "$1" == "1" ]; then
+        shift 1
+        run_helper nsenter -n -t $HOST1_NS_PID "$@"
+    elif [ "$1" == "2" ]; then
+        shift 1
+        run_helper nsenter -n -t $HOST2_NS_PID "$@"
+    else 
+        die
+    fi
```
Luap99, line 117: The hostns can only be one netns! You cannot have more the one host.

test/helpers.bash
```diff
@@ -66,23 +80,41 @@ function get_container_netns_path() {
...
-    run_helper nsenter -n -t $CONTAINER_NS_PID "$@"
+    if [ "$1" == "1" ]; then
+        shift 1
+        run_helper nsenter -n -t $CONTAINER1_NS_PID "$@"
+    elif [ "$1" == "2" ]; then
+        shift 1
+        run_helper nsenter -n -t $CONTAINER2_NS_PID "$@"
+    else    
+        die
+    fi
```
Luap99, line 101: This is not portable at all. Look at the aardvark tests which already handle more than one containerns do this https://github.com/containers/aardvark-dns/blob/main/test/helpers.bash

src/network/types.rs
```diff
@@ -173,6 +173,10 @@ pub struct NetInterface {
     /// Subnets list of assigned subnets with their gateway.
     #[serde(rename = "subnets")]
     pub subnets: Option<Vec<NetAddress>>,
+
+    /// Isolation determines whether the network will be able to communicate to others
+    #[serde(rename = "isolation")]
+    pub isolation: bool,
```
Luap99, line 179: This should not be returned. Podman already knows this since the setting is in the network config. The return values are only for dynamic values such as ip/mac addresses.

src/network/core.rs
```diff
@@ -275,6 +290,20 @@ impl Core {
...
+                    return Err(std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        format!("unable to parse isolation type: {}", err),
+                    ))
+                }
+            },
+            // default is to not isolate podman networks
+            None => false,
+        };
+
```
Luap99, line 306: This is not relevant, isolation should only be used with the bridge driver.

src/firewall/varktables/types.rs
```diff
@@ -163,6 +163,8 @@ pub fn get_network_chains(
     network: IpNet,
     network_hash_name: String,
     is_ipv6: bool,
+    network_name: String,
```
Luap99, line 166: This should be the bridge (interface) name and not the network name.

src/commands/setup.rs
```diff
@@ -118,10 +119,15 @@ impl Setup {
                         continue;
                     }
                     let id_network_hash = CoreUtils::create_network_hash(net_name, MAX_HASH_SIZE);
+                    let hash_int = status_block.to_owned().interfaces.unwrap();
```
Luap99, line 122: never use unwrap(), return a reasonable error instead. Also you should not get the option from the status. just parse the option from the network config here!

src/firewall/iptables.rs
```diff
@@ -44,11 +44,14 @@ impl firewall::FirewallDriver for IptablesDriver {
                 if is_ipv6 {
                     conn = &self.conn6;
                 }
+
                 let chains = varktables::types::get_network_chains(
                     conn,
                     network.subnet,
                     network_setup.network_hash_name.clone(),
                     is_ipv6,
+                    network_setup.net.name.to_string(),
```
Luap99, line 53: this must be the interface name not network name.

src/firewall/iptables.rs
```diff
@@ -77,6 +80,8 @@ impl firewall::FirewallDriver for IptablesDriver {
                     network.subnet,
                     tear.config.network_hash_name.clone(),
                     is_ipv6,
+                    tear.config.net.name.to_string(),
```
Luap99, line 83: this must be the interface name not network name.

src/commands/setup.rs
```diff
@@ -118,10 +119,15 @@ impl Setup {
...
+                    let opt = hash_int.get(&per_network_opts.interface_name);
+                    let interf = opt.unwrap();
```
flouthoc, line 124: We could avoid `unwap()` here as well.

src/network/core.rs
cdoern: Ok I was not sure which to put it in since you linked the MTU example to macvlan above, I will remove this

test/helpers.bash
cdoern: Ok, that makes sense. making multiple container namespaces makes sense though, right?

cdoern:
@Luap99 what is the easiest way to avoid unwrap? some sort of match statement?

Luap99:
yes use match, it should look something like this: https://github.com/containers/netavark/blob/9437d41962197a1eb5a2edc3334d2edcf80b8286/src/network/core_utils.rs#L66-L77

cdoern:
@Luap99 how can i actually test this? I vendored in this netavark branch into podman and common and tried to create two networks with this option and then run two containers pinging eachother but the pings did not fail... Do i have to set some ENV Variables for the netavark binary?

cdoern:
ah I think I see the issue. I did not tell the forward chain to jump to netavark_isolation_1

cdoern:
FYI I tested this in podman (I just changed the `CONTAINERS_HELPER_BINARY_DIR` variable) and it works perfectly between two containers on isolated networks :)

test/helpers.bash
cdoern: I will modify it to be an array of NS's but the logic of passing an arg for which one still holds true in aardvark

cdoern:
@Luap99 I am unsure as to why the `run_nc_test` function is repeatedly failing. It seems to be something I am not understanding about bash arrays. Any help on fixing the tests would be appreciated!

cdoern:
Hm actually my local failures might be because of something else as they did not happen here... I'll take another look in the morning at what's actually failing.

cdoern:
@Luap99 @mheon  @flouthoc PTAL I made my own version of CNI's isolation rules. Not sure if they are too rigid

src/commands/setup.rs
```diff
@@ -117,11 +118,30 @@ impl Setup {
...
+                                }
+                            },
+                            // default is to not isolate podman networks
+                            None => false,
+                        };
+
                     let id_network_hash = CoreUtils::create_network_hash(net_name, MAX_HASH_SIZE);
+                    //let hash_int = status_block.to_owned().interfaces.unwrap();
+                    //let opt = hash_int.get(&per_network_opts.interface_name);
+                    // let interf = opt.unwrap();
```
Luap99, line 138: Looks like this can be removed

src/commands/setup.rs
```diff
@@ -117,11 +118,30 @@ impl Setup {
...
+                                Err(err) => {
+                                    bail!("{} could not parse isolation option", err);
+                                }
+                            },
+                            // default is to not isolate podman networks
+                            None => false,
+                        };
+
                     let id_network_hash = CoreUtils::create_network_hash(net_name, MAX_HASH_SIZE);
+                    //let hash_int = status_block.to_owned().interfaces.unwrap();
```
flouthoc, line 136: Could you please remove these commented lines.

src/firewall/iptables.rs
```diff
@@ -37,18 +37,34 @@ pub fn new() -> Result<Box<dyn firewall::FirewallDriver>, Box<dyn Error>> {
...
+            match network_setup.net.network_interface {
+                Some(iface) => interface = iface,
+                None => {
+                    return Err(Box::new(crate::error::NetavarkError {
+                        error: "failed to get interface".to_string(),
+                        errno: 1,
+                    }));
+                }
+            }
+        }
```
Luap99, line 51: This looks very complex, i think you just need the inner match block with `let interface = match ...`

src/firewall/iptables.rs
```diff
@@ -64,6 +80,18 @@ impl firewall::FirewallDriver for IptablesDriver {
...
+            match tear.config.net.network_interface {
+                Some(iface) => interface = iface,
+                None => {
+                    return Err(Box::new(crate::error::NetavarkError {
+                        error: "failed to get interface".to_string(),
+                        errno: 1,
+                    }));
+                }
+            }
+        }
```
Luap99, line 94: same here, also we should avoid to error in cleanup and try to remove as much as possible.

src/network/core.rs
```diff
@@ -55,6 +55,21 @@ impl Core {
...
+                    return Err(std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        format!("unable to parse isolation type: {}", err),
+                    ))
+                }
+            },
+            // default is to not isolate podman networks
+            None => false,
+        };*/
+
```
Luap99, line 72: remove this

src/network/internal_types.rs
```diff
@@ -11,12 +11,14 @@ pub struct TeardownPortForward {
 }
 
 //  SetupNetwork contains options for setting up a container
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
Luap99, line 14: I think this can be removed, we do not want to serialize and deserialize this.

test/helpers.bash
```diff
@@ -55,7 +60,8 @@ function create_netns() {
 }
 
 function get_container_netns_path() {
-    echo /proc/$CONTAINER_NS_PID/ns/net
+    local which=$1
+    echo /proc/"${CONTAINER_NS_PIDS[$which]}"/ns/net
```
Luap99, line 64: lets make `0` the default argument so we can remove all the `get_container_netns_path 0` diffs from this PR

test/helpers.bash
```diff
@@ -74,13 +80,15 @@ function run_netavark() {
 ################
 #
 function run_in_container_netns() {
-    run_helper nsenter -n -t $CONTAINER_NS_PID "$@"
+    local i=$1
+    shift 1
+    run_helper nsenter -n -t "${CONTAINER_NS_PIDS[$i]}" "$@"
```
Luap99, line 85: same here, 

lets make `0` the default argument so we can remove all the `run_in_container_netns 0` diffs from this PR

src/network/core.rs
```diff
@@ -55,6 +55,21 @@ impl Core {
             None => 0,
         };
 
+        /*
+        let isolation_config = match network.options.as_ref().and_then(|map| map.get("isolate")) {
```
flouthoc, line 59: We can remove this block.

Luap99 (commented):
only checked the style of the code, I will test the firewalls rules later

src/firewall/varktables/types.rs
```diff
@@ -200,41 +203,104 @@ pub fn get_network_chains(
...
-        // Insert the rule into the first position
         let mut forward_chain = VarkChain::new(conn, FILTER.to_string(), FORWARD.to_string(), None);
+        let res = add_chain_unique(
+            forward_chain.driver,
+            &forward_chain.table,
+            &forward_chain.chain_name,
+        );
+        match res {
+            Ok(_) => {}
+            Err(_) => {
```
flouthoc, line 215: can use `if` here and we should return error instead of `std::process:exit(`

src/firewall/varktables/types.rs
```diff
@@ -200,41 +203,104 @@ pub fn get_network_chains(
...
+            chains.push(netavark_isolation_chain);
+        }
+
+        // forward needs to be added to "chains" first so its rule are created before NETAVARK_FORWARD. To do this we
+        // need to create the chain so its rules can be referenced
+        let res = add_chain_unique(conn, FILTER, NETAVARK_FORWARD);
+        match res {
+            Ok(_) => {}
+            Err(_) => {
+                std::process::exit(1);
```
flouthoc, line 266: should use `if let` here and return error instead of `process:exit(1)`

src/firewall/iptables.rs
```diff
@@ -37,18 +37,34 @@ pub fn new() -> Result<Box<dyn firewall::FirewallDriver>, Box<dyn Error>> {
 
 impl firewall::FirewallDriver for IptablesDriver {
     fn setup_network(&self, network_setup: SetupNetwork) -> Result<(), Box<dyn Error>> {
+        let mut interface = "".to_string();
+        if let Some(_int) = network_setup.net.network_interface.clone() {
```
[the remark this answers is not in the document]
cdoern: removing the clone results in `Some(iface...`) to give me an error saying "use of moved value"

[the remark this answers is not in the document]
cdoern: actually removing the if fixes that issue. Why can I remove the if here? doesn't that check for the existence of the Option?

src/firewall/iptables.rs
cdoern: @Luap99 how does that work? I am unfamiliar with how rust errors work

test/helpers.bash
cdoern: this one might be weird. The other function is east, check if there is an argument and if so use it but here the number of arguments is unpredictable. Should I for now just say if arg1 == "1" and hardcode it?

cdoern:
@Luap99 @flouthoc @mheon PTAL

src/commands/setup.rs
```diff
@@ -117,10 +117,25 @@ impl Setup {
...
                         continue;
                     }
+
+                    // parse isolation option here so we can avoid unwrapping and storing it unecessarily
+                    let isolation_config =
+                        match network.options.as_ref().and_then(|map| map.get("isolate")) {
+                            Some(isolation) => match isolation.parse() {
+                                Ok(isolation) => isolation,
+                                Err(err) => {
+                                    bail!("{} could not parse isolation option", err);
```
flouthoc, line 127: At higher level setup calls we are using wrapping via ` return Err(std::io::Error::new(` throughout the code, could we wrap this in error instead of `bail` for consistency.

src/firewall/iptables.rs
```diff
@@ -37,22 +37,40 @@ pub fn new() -> Result<Box<dyn firewall::FirewallDriver>, Box<dyn Error>> {
...
 
-                for chain in chains {
-                    chain.add_rules()?;
+                match chain_result {
+                    Ok(chains) => {
+                        for c in chains {
+                            c.add_rules()?
+                        }
+                    }
+                    Err(_e) => {}
```
flouthoc, line 73: Lets not ignore `err` here and return error if there is any.

src/firewall/iptables.rs
```diff
@@ -77,22 +105,29 @@ impl firewall::FirewallDriver for IptablesDriver {
...
+                                None => {}
+                                Some(policy) => {
+                                    if tear.complete_teardown && *policy == OnComplete {
+                                        c.remove()?;
+                                    }
+                                }
                             }
                         }
                     }
+                    Err(_e) => {}
```
flouthoc, line 130: Lets not ignore `error` here.

src/firewall/varktables/types.rs
```diff
@@ -200,44 +203,96 @@ pub fn get_network_chains(
...
-        netavark_forward_chain.build_rule(VarkRule::new(
-            format!("-s {} -j ACCEPT", network),
-            Some(TeardownPolicy::OnComplete),
-        ));
-        chains.push(netavark_forward_chain);
-
         // FORWARD chain
-        // Insert the rule into the first position
         let mut forward_chain = VarkChain::new(conn, FILTER.to_string(), FORWARD.to_string(), None);
+        add_chain_unique(
```
flouthoc, line 208: Could you add a comment above about why is `add_chain_unique` here is needed ?

src/firewall/varktables/types.rs
```diff
@@ -200,44 +203,96 @@ pub fn get_network_chains(
...
+            forward_chain.driver,
+            &forward_chain.table,
+            &forward_chain.chain_name,
+        )?;
+
+        // used to prepend specific rules
+        let mut ind = 1;
+
+        debug!("ISOLATION: {}", isolation);
```
flouthoc, line 217: This debug log is leaving no context, it would be hard to figure out what does this means later.

src/firewall/varktables/types.rs
```diff
@@ -200,44 +203,96 @@ pub fn get_network_chains(
...
+
+            // -A FORWARD -j NETAVARK_ISOLATION_1
+            forward_chain.build_rule(VarkRule {
+                rule: format!("-j {}", "NETAVARK_ISOLATION"),
+                position: Some(ind),
+                td_policy: Some(TeardownPolicy::OnComplete),
+            });
+            ind += 1;
+
+            // NETAVARK_ISOLATION -i bridge_name ! -o bridge_name -j DROP
```
flouthoc, line 236: I have not reviewed this IP table rule but this would need ack from other maintainers.

flouthoc: @mheon @Luap99 PTAL

src/commands/setup.rs
cdoern: ok, I will do the same for the other ignored errors below

cdoern:
friendly ping @mheon @Luap99

cdoern:
> The iptables rules only work partially. It looks like the iptable rules are not removed on network teardown. Also I have no idea what the RTURN chain is doing here:
> 
> ```
> Chain NETAVARK_ISOLATION (1 references)
>  pkts bytes target     prot opt in     out     source               destination         
>    36  2357 DROP       all  --  podman4 !podman4  0.0.0.0/0            0.0.0.0/0           
>    27  1956 RETURN     all  --  *      *       0.0.0.0/0            0.0.0.0/0           
>     0     0 DROP       all  --  podman5 !podman5  0.0.0.0/0            0.0.0.0/0    
> ```
> 
> because of it podman5 is never reached.

I think I know why the teardown is not happening. As for RETURN, not sure either. I am going to work on finishing this up today

cdoern:
@Luap99 PTAL, I think I resolved both of your issues.

Luap99:
You also have to rebase

Luap99:
@mheon PTAL

src/firewall/varktables/types.rs
```diff
@@ -163,7 +164,9 @@ pub fn get_network_chains(
     network: IpNet,
     network_hash_name: String,
     is_ipv6: bool,
-) -> Vec<VarkChain<'_>> {
+    interface_name: String,
+    isolation: bool,
+) -> Result<Vec<VarkChain<'_>>, Box<dyn Error>> {
```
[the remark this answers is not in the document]
cdoern: add_chain_unique propagates the error back up with `?`. At least thats what I was trying to do @Luap99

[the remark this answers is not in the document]
Luap99: You are right, now I see it. The `?` is just super easy to miss while looking at the code on github.

test/100-bridge-iptables.bats
```diff
@@ -321,3 +322,39 @@ EOF
...
+    expected_rc=1 run_in_container_netns 1 ping -c 1 10.88.0.2
+
+    run_netavark --file ${TESTSDIR}/testfiles/connectbridge.json teardown $(get_container_netns_path 1)
+     
+    run_netavark --file ${TESTSDIR}/testfiles/connectbridge.json setup $(get_container_netns_path)
+    result2="$output"
+    assert_json "$result2" 'has("podman1")' == "true" "object key exists"
+
+    # this should not fail
+    run_in_container_netns ping -c 1 10.88.0.2
```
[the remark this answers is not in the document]
cdoern: do you mean ping a container with isolate to one without isolate?

cdoern:
@Luap99 @mheon validation seems to be picking up changes I did not make... should I try repushing or should I change what it wants me to?

test/testfiles/connectbridge.json
```diff
@@ -0,0 +1,32 @@
...
+            "driver": "bridge",
+            "id": "f3ac3c903b76f20e8a122b1eb2ba393ab2519ba516626be5b490b66dc96b54b7",
+            "internal": false,
+            "ipv6_enabled": false,
+            "name": "podman1",
+            "network_interface": "podman1",
+            "subnets": [
+                {
+                    "gateway": "10.88.1.2",
+                    "subnet": "10.88.1.0/22"
```
Luap99, line 24: I just noticed this now, the subnet is overlapping with the other one used in your test. This can cause false positives.

Luap99 (commented):
LGTM, I will merge and fixup the test myself.

PR #289, opened 2022-05-02
WIP: Add a dedicated NetavarkError type

mheon:
Subsequently, plumb this error through most of Netavark. The interface creation logic is still mostly on std::io::Error, but that natively converts into NetavarkError via From so I'm not touching that right now.

Still to do: error equality, and possibly dedicated types for iptables and firewalld so we can better track where in the code errors occur.

Also add a NetavarkResult and convert most of our code to use it.


src/error/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+impl NetavarkError {
+    // TODO: There has to be a better way of doing this
+    pub fn from_str(string: &str) -> NetavarkError {
+        NetavarkError::Message(string.to_string())
+    }
+
+    // TODO: There has to be a better way of doing this
+    pub fn make_chain(string: String, chained: NetavarkError) -> NetavarkError {
```
Luap99, line 39: can we name this `wrap` instead to better match the go term?

mheon: Sure

mheon:
I'll ask around to see if there's a better way, but if the answer is macros, I'll probably leave the todos - I still don't fully grok the way Rust macros are defined.

src/error/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &*self {
+            NetavarkError::Message(s) => write!(f, "{}", s),
+            NetavarkError::ExitCode(s, _) => write!(f, "{}", s),
+            NetavarkError::Chain(s, e) => write!(f, "{}: {}", s, e),
+            NetavarkError::Io(e) => write!(f, "IO error: {}", e),
+            NetavarkError::Dbus(e) => write!(f, "DBus error: {}", e),
+            NetavarkError::DbusVariant(e) => write!(f, "DBus Variant Error: {}", e),
+            NetavarkError::Sysctl(e) => write!(f, "Sysctl error: {}", e),
+            NetavarkError::Serde(e) => write!(f, "JSON Decoding error: {}", e),
```
flouthoc, line 82: Do we need `netlink` as well here or maybe we can leave it for now. We can extend this for netlink later.

mheon:
Dropping WIP

src/error/mod.rs
flouthoc: Yes they are already wrapped up so no issues there. Just thought this also adds `prefix` with type of error but I agree they are wrapped so not a blocker.

mheon:
I think this will go green now. Accidentally removed a ! in the iptables logic.

flouthoc:
> LGTM but I would like to change all the network errors too, there is no reason why they should be io errors.

Yes looks like they are showing up as IO error as of now. Maybe a new type for them in a folloup up PR ?

flouthoc (approved):
LGTM

One nit above but i hope its not part of this PR.

PR #295, opened 2022-05-06
Makefile: add target for release tarball

Luap99:
Add a target to automatically create the vendor tarball and the extra
files. This can only be used on a version which has not the -dev suffix
to prevent user errors.


Luap99:
@baude @mheon @flouthoc PTAL

This should automatically create the release files which then should be uploaded to github. I think it would be great if we could make this automatically in CI.

flouthoc (approved):
LGTM  nice idea

mheon:
Maybe add an `rm -rf` for the vendor directory in there to ensure it doesn't get committed to the repo?

Otherwise LGTM

PR #300, opened 2022-05-20
dns: wait for aardvark server just after spawning `aardvark-dns`.

flouthoc:
Since starting DNS server is the last task for netavark's `network
setup` so netavark must verify at least a few times if aardvark-dns is
in a state where it can serve request before passing context back to 
container manager.

Potential fix: https://github.com/containers/podman/issues/14173

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -12,6 +12,7 @@ use std::net::IpAddr;
 use std::net::Ipv4Addr;
 use std::path::Path;
 use std::process::{Command, Stdio};
+use std::{thread, time};
 
 const SYSTEMD_CHECK_PATH: &str = "/run/systemd/system";
 const SYSTEMD_RUN: &str = "systemd-run";
@@ -76,9 +77,8 @@ impl Aardvark {
         false
     }
 
-    pub fn start_aardvark_server(&self) -> Result<()> {
+    pub fn start_aardvark_server(&mut self) -> Result<()> {
         log::debug!("Spawning aardvark server");
-
         let mut aardvark_args = vec![];
         // only use systemd when it is booted, see sd_booted(3)
         if Path::new(SYSTEMD_CHECK_PATH).exists() && Aardvark::is_executable_in_path(SYSTEMD_RUN) {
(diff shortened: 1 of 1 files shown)


Luap99:
I don't think this is the right way. There is still no guarantee that this waits for aardvark to be ready.

flouthoc:
@Luap99 Yes as of now it only waits for while and then proceeds. I am just afraid that if we wait for infinite duration and aardvark fails this netavark would end up failing as well.

I can think of two options. Do you have any suggestion.
* Wait till aardvark pid shows up and not break on time up limit.
* Perform a dummy dns request from netavark to aardvark and proceed if true.

src/dns/aardvark.rs
```diff
@@ -115,6 +115,24 @@ impl Aardvark {
...
+        // Starting aardvark server is the last task for netavark so in some env
+        // setup netavark will end up exiting way sooner then aardvark process
+        // is actually ready to serve so after starting aardvark we have to wait
+        // till aardvark process is ready. If it does not shows up atleast retry
+        // 10 times with a delay of 500ms and then return.
+        let mut retry_count = 10;
+        while retry_count > 0 {
+            let aardvark_pid = self.get_aardvark_pid();
+            if aardvark_pid != -1
+                && signal::kill(Pid::from_raw(aardvark_pid), Signal::SIGHUP).is_ok()
```
mheon, line 127: Avoid SIGHUP, send signal 0 if possible - it takes no action on the target process, just returns if the process is alive.

Luap99:
I think the best thing to do is do a fork on aardvark side. Basically netavark spawns aardvark and then waits for exit. aardvark then sets up everything and once it is ready if forks itself and the parent exits so netavark know it is ready.

This will also fix issues with the error reporting since we can keep the stderr open until the fork.

This however does not fix the issue when adding new entries. In this case we still would need a bidirectional way to update so aardvark could signal us that it is ready.

flouthoc:
I think we could this i am trying out a spike here: https://github.com/containers/aardvark-dns/pull/148 and we might need few changes on netavark end as well.

Luap99:
we went with https://github.com/containers/netavark/pull/308 so I am closing this one

PR #323, opened 2022-07-01
dns: start aardvark-dns on a different port

martinetd:
I've taken a stab at a proof of concept first -- as said in the issue about port 53 shouldn't be used (link below) it seems simple enough to use DNAT for this.

This WIP version has at least two problems:
 - I've hardcoded port 1153 as alternative port, we should either use an ephemeral port (problem: we can't know if it's free unless we try to bind to it, so we'd need some retry...) or simpler just make it configurable at network level as a driver-specific option. I'd favor the later.
 - I'm not actually checking if dns settings are enabled to create the forward rule, that should obviously be checked.

Is there anything else I missed?

Fixes: https://github.com/containers/aardvark-dns/issues/13

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -111,6 +111,9 @@ impl Setup {
                         network,
                         &self.network_namespace_path,
                     )?;
+                    // get DNS server IPs
+                    let dns_server_ips: Vec<IpAddr> =
+                        status_block.dns_server_ips.clone().unwrap_or_default();
                     response.insert(net_name.to_owned(), status_block);
                     if network.internal {
                         match &network.network_interface {
@@ -200,7 +203,8 @@ impl Setup {
                                 subnet_v4: net_v4,
                                 container_ip_v6: addr_v6,
                                 subnet_v6: net_v6,
-                                dns_port: if network.dns_enabled { dns_port } else { 53 },
+                                dns_port,
+                                dns_server_ips,
                             };
                             // Need to enable sysctl localnet so that traffic can pass
                             // through localhost to containers
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -11,6 +11,7 @@ use crate::network::{core_utils, types};
 use clap::Parser;
 use log::{debug, info};
 use std::collections::HashMap;
+use std::env;
 use std::fs;
 use std::net::IpAddr;
 use std::path::Path;
@@ -69,6 +70,19 @@ impl Setup {
         core_utils::CoreUtils::apply_sysctl_value(IPV4_FORWARD, "1")?;
 
         let mut response: HashMap<String, types::StatusBlock> = HashMap::new();
+
+        let dns_port = match env::var("NETAVARK_DNS_PORT") {
+            Ok(port_string) => match port_string.parse() {
+                Ok(port) => port,
+                Err(e) => {
+                    return Err(NetavarkError::Message(format!(
+                        "Invalid NETAVARK_DNS_PORT {}: {}",
+                        port_string, e
+                    )))
+                }
+            },
+            Err(_) => 53,
+        };
 
         // Perform per-network setup
         for (net_name, network) in network_options.network_info.iter() {
@@ -186,6 +200,7 @@ impl Setup {
                                 subnet_v4: net_v4,
                                 container_ip_v6: addr_v6,
                                 subnet_v6: net_v6,
+                                dns_port: if network.dns_enabled { dns_port } else { 53 },
                             };
                             // Need to enable sysctl localnet so that traffic can pass
                             // through localhost to containers
@@ -255,7 +270,8 @@ impl Setup {
                 }
             };
 
-            let mut aardvark_interface = Aardvark::new(path_string, rootless, aardvark_bin);
+            let mut aardvark_interface =
+                Aardvark::new(path_string, rootless, aardvark_bin, dns_port);
 
             if let Err(er) = aardvark_interface.commit_netavark_entries(
                 network_options.container_name,
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -142,21 +142,14 @@ impl Setup {
                         isolation: isolation_config,
                     };
                     firewall_driver.setup_network(sn)?;
-                    let port_bindings = network_options.port_mappings.clone();
-                    match port_bindings {
+                    match per_network_opts.static_ips.as_ref() {
                         None => {}
-                        Some(i) => {
-                            let container_ips =
-                                per_network_opts.static_ips.as_ref().ok_or_else(|| {
-                                    std::io::Error::new(
-                                        std::io::ErrorKind::Other,
-                                        "no container ip provided",
-                                    )
-                                })?;
+                        Some(container_ips) => {
+                            let port_bindings = network_options.port_mappings.clone();
                             let networks = network.subnets.as_ref().ok_or_else(|| {
                                 std::io::Error::new(
                                     std::io::ErrorKind::Other,
-                                    "no network address provided",
+                                    "IP assigned but no network address provided",
                                 )
                             })?;
                             let mut has_ipv4 = false;
@@ -186,7 +179,7 @@ impl Setup {
                             let spf = PortForwardConfig {
                                 net: network.clone(),
                                 container_id: network_options.container_id.clone(),
-                                port_mappings: i.clone(),
+                                port_mappings: port_bindings.unwrap_or_default(),
                                 network_name: (*net_name).clone(),
                                 network_hash_name: id_network_hash.clone(),
                                 container_ip_v4: addr_v4,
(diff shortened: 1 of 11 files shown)


martinetd:
Sorry for the push spam to anyone subscribed -- the failing tests weren't so much related to the dns redirection rules than just the existing code wasn't working so well with an empty port forward config (as opposed to not set)
 - we weren't creating a chain but using it, making setup fail
 - we were trying to remove chains we didn't create, making teardown fail
.. and I have no idea why the clippy failures didn't bother anyone else before me.

I've also taken the liberty of updating the iptables backend test to use `iptables -S` instead of `-L` -- I just can't read -L output so it was more natural for me to fix the test in the following commit later on, but if someone is adamant that -L is better I can drop this patch. These four patches are unrelated to the dns feature so would be happy to create a separate PR for them if you'd like.

the rest is as before, waiting for feedback in the related issue before I start worrying about adding a config value or something else.

martinetd:
updated to fix ipv6, needed to put the ip into brackets for to-destination like we do with container ip below. For some reason my `/run/containers/networks/aardvark-dns/<network>` first line only contained ipv4 though so I had to manually fix the file and restart aardvark to bind to both IPs

(speaking of restarting aardvark, with this and a > 1024 port I could run aardvark-dns as a non-root user after giving read permission to `/run/containers/` and pre-creating/giving the user `/run/containers/networks/aardvark-dns/aardvark.pid` -- that might be worth doing eventually. Changing permissions is a bit too big for me to chew on all at once.)


@Luap99 I'd like to get started on some config option, what would you think is best? I was thinking network driver option e.g. `podman network create --opt x y` but your suggestion of containers.conf might be better for users? I didn't see how to access it from netavark though, while network.options is easy enough to access.
(... and will need to update go's containers common libnetwork to allow the new options as well...)

flouthoc:
@martinetd I think 1153 or any other configurable port is fine but could we bind and unbind at netavark end to register ephemeral port ? and then pass to aardvark ? ( just thinking out loud my idea could be wrong )

martinetd:
> @martinetd I think 1153 or any other configurable port is fine but could we bind and unbind at netavark end to register ephemeral port ? and then pass to aardvark ? ( just thinking out loud my idea could be wrong )

hmm, we definitely could. create a socket and bind until we got a free port in netavark, then pass the fd to be used as is in aardvark should work -- it'll be harder to debug/run aardvark manually though (need to figure out what port was used, remember what argument it was so aardvark tries to bind itself, try again...)
We could also just close the socket and start aardvark with that port but that leaves room for a race we probably want to avoid.

overall I still think config is better if only for debugability, but happy to go either way if there's a preference.

mheon:
Don't really see any reason that we should do the allocation on the NV side - why would it need to know the port used?

We should definitely have a configuration knob to force the port to a fixed value, but otherwise finding a random open port seems reasonable. In Podman, we have logic like this for handling Expose, and the kernel offers assistance - there's a way to have the kernel give you a binding to a random, open, high-numbered (over 1024) port (I believe it's by binding to `:0` but don't quote me on that). Duplicating that logic here seems like a safe way of operating.

flouthoc:
> overall I still think config is better if only for debugability, but happy to go either way if there's a preference.

@martinetd No preference :) i think going with fixed port will be easier to maintain above was just an idea.

> Don't really see any reason that we should do the allocation on the NV side

@mheon Only to check port availablity and pass the port number to aardvark not the fd. But I think going with fixed port is a lot easier to implement here.

martinetd:
> Don't really see any reason that we should do the allocation on the NV side - why would it need to know the port used?

netavark is the one setting up iptables -- and there is no standard syntax for /etc/resolv.conf to use an alternate port so we need some iptables like I've added to have things "just work" in the container.
Unless we want aardvark to setup the NAT rules itself, but I'd rather work towards lowering aardvark privileges now we can have it use a port > 1024 than the other way around...

With the current version of this PR you end up with a nat rule like this:
```
-A NETAVARK-HOSTPORT-DNAT -d 10.89.1.1/32 -p udp -m udp --dport 53 -m comment --comment "dnat name: test id: 5ccfb3473b64702816330ce3f3c0b48deedc223a1e9ec440fab96472f3bfcb26" -j DNAT --to-destination 10.89.1.1:1153
```
(the comment is necessary so each container has their own rule that can be removed when container stops, we'd only need it once per network but I didn't see any option that would remove the rule on network removal)

> We should definitely have a configuration knob to force the port to a fixed value, but otherwise finding a random open port seems reasonable. In Podman, we have logic like this for handling Expose, and the kernel offers assistance - there's a way to have the kernel give you a binding to a random, open, high-numbered (over 1024) port (I believe it's by binding to :0 but don't quote me on that). Duplicating that logic here seems like a safe way of operating.

Yes, just finding a free port in itself is easy. It's agreeing on a port between both netavark and aardvark that is a bit annoying -- config item, random port decided by aardvark and reported back through a pipe to netavark, port allocated by netavark and passed on to aardvark... There's plenty of ways of doing it, but it's all rather intrusive.

mheon:
Ahh. I see your point.

Random assignment within Netavark and passing to Aardvark from there could be a good compromise, then.

mheon:
Manual configuration ought to be fine, so long as a sane default is chosen.

martinetd:
Spent a bit of time on this today.
I've split the pure fixes out to #329, this branch is based on it so it'll get a bit simpler once that other branch is merged.

For what matters here, I've added a NETAVARK_DNS_PORT env variable and tested by manually exporting it, but... the thing gets unset by the time the teardown is called, so we can't rely on it to remove the iptables rule, so that's not really practical for tests.
Once that's added to containers common though teardown will be called through it and I believe that should work, but there's only one way to find out -- I'll open a PR there as well tomorrow.
(the default is left at 53 so at least that won't bother anyone)

I believe this PR to be "ready enough" if that works as expected -- please have a look when you can.

TODO before I mark it ready:
 - open PR for containers/common/libnetwork common that sets env from containers.conf
 - have a look at firewalld backend, I set it up earlier but didn't get time to finish with the other bugs
 - (I'd say the same about nftables but that's not present yet, even firewalld doesn't seem ever used unless configured manually with NETAVARK_FW?)
 - add some tests to make sure we set and cleanup the rules appropriately if port isn't 53

Thanks!

martinetd:
I think I'm almost done with this. Definitely not the code I'm most proud of (especially the firewalld handling... with hindsight it would have been better to split the dbus array filling in subfunctions...) but it works™, and I've added tests to prove it.

I say almost because the tests worked locally and fail on CI: looking again it's because netavark setup running as root successfully runs aardvark-dns, but I was running in a `podman unshare --rootless-netns` where starting aardvark fails; and I'm not sure what to do with that:
 - should we assume tests run as root as they apparently do in the CI, at which point I can remove the ugly nc test and just run some command (err, I don't suppose bind utils (dig or host) will be installed will they?...)
 - if so, should we actually check aardvark started?
 - if not, should the test just kill aardvark if it was running?
 - or should it opportunistically try to use it if it's there, or fallback to nc? or skip the test?

that's really more a policy decision than actual coding so I'm leaving that to you all to decide.

That aside it should all work, I've also tested manually with https://github.com/containers/common/pull/1084 with no problem.

martinetd:
To echo back on the previous comment and https://github.com/containers/netavark/pull/329#issuecomment-1177466852 on the other PR, I've expanded the check a bit to actually test aardvark:
 - try to guess if we're running in a podman unshare --rootless-netns or as root for netavark rootless argument (that affects aardvark being started with systemd-run or systemd-run --user)
 - passing a config dir so the config is generated properly
 - testing the config contents, including that both v4 and v6 gateways are listed on first line
 - testing that aardvark runs
 - testing an actual dns query through the port redirection, dig is apparently available on test machine (didn't see any difference with aardvark-dns test setup which uses it)

If you think I've overdone it we can probably add a bunch of e.g. skip if e.g. /usr/libexec/podman/aardvark-dns or dig aren't present, but actual dns resolution aside this is all netavark's job and not aardvark-dns's so I think this belongs here.
(fun fact, netavark ignores if we request just A or AAAA through dig and gives both IPs anyway for containername.dns.podman -- it works on normal queries though)


That tests fixes from the other PR here but the test is all mixed up and I'm not sure it'd be worth the effort of untangling it to test just the config generation -- at this point it might make more sense to just close #329 and finish this here since I got it ready faster than I thought I would; leaving that decision to you for whatever is easier to review.

At the very least don't think I'll be touching this anymore unless explicitly requested...

Luap99 (commented):
Nice work, @mheon PTAL at the firewalld code.

martinetd:
Thanks!
(force push just now just fixes a few stray tabs in tests I just noticed, no code change)

mheon:
Still has the work in progress label - safe to remove & merge?

martinetd:
> Still has the work in progress label - safe to remove & merge?

Yes, I removed WIP from commit messages / draft from github when it became safe to merge but I didn't notice that label. Honestly no idea how to remove it either, so leaving that to you :)

Thanks!

Luap99:
I bet the bot forgot to remove it when you drop the draft status. The openshift bot seems to be very slow recently.

PR #322, opened 2022-06-30
Cirrus: Gather arm and aarch64 netavark binaries

cevich:
In order to test `aardvark-dns` (and maybe elsewhere) a recent binary
build of netavark from the CI system is needed.  Previously this was
always an `x86_64` binary along with a `.debug` version.  Add support for
archiving both `arm` (32 bit) and `aarch64` binaries.  Name them with
the architecture-target as a suffix so they may be distinguished from
the (necked) `netavark` binary.

e.g. `bin/netavark.aarch64-unknown-linux-gnu`

*note:* A VM image update was also required

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #324, opened 2022-07-01
core,lint: omit the unused `let` binding

flouthoc:
After the recent cargo update validate fails because of using `let`
binding on unit value further we are not interested in using this value
so drop it.

See: https://rust-lang.github.io/rust-clippy/master/index.html#let_unit_value

Fixes issue in: https://github.com/containers/netavark/pull/322
Fixes broken CI on current main branch: https://cirrus-ci.com/build/4539848676605952

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -127,7 +127,7 @@ impl Core {
             gw_ipaddr_clone.push(*gw_ip)
         }
         //call configure bridge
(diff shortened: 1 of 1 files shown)


flouthoc:
@Luap99 Merge me please this should unblock CI for https://github.com/containers/netavark/pull/322 and other PR

PR #325, opened 2022-07-01
GHA: Add Cirrus-cron failure notifications

cevich:
Requires: https://github.com/containers/buildah/pull/4088

This re-uses the buildah-workflow to send e-mail to podman-monitor if
any Cirrus-cron job failed, or the workflow itself fails.


cevich:
@flouthoc assuming the required buildah PR merges, this this should send the podman-monitor list e-mail notifications when any Cirrus-cron job fails.  This is why we didn't notice the recent validation-barf on `main`.

cevich:
**note:** I've pre-added all the required secret values needed by this action.

cevich:
Waiting on https://github.com/containers/buildah/pull/4102

cevich:
@Luap99 I just rebased.  Yes, assuming tests pass (they should) please  merge and then I need to manually check/verify it actually works (since it can't be tested in a PR :disappointed:)

cevich:
Fix is in here: https://github.com/containers/buildah/pull/4108

cevich:
Fixed: https://github.com/containers/netavark/runs/7327439481?check_suite_focus=true

PR #328, opened 2022-07-04
fixup isolate networks test

Luap99:
Fix some problems with the isolation test. Fix the overlapping subnets
to make sure we correctly route. Add `-w 1` to ping to prevent it from
just hanging. Turn off dns on the network since we do not need it and
fix the test comments to actually reflect what we do.


flouthoc (approved):
/lgtm
/approve
LGTM

PR #329, opened 2022-07-06
misc fixes

martinetd:
While working on #323 I found a few issues, I had intended to just push them together but I found two more today so might as well fix separately.

Individual commit messages have better descriptions, but to recap:
 - the chain.create logic in iptables was never correct, we cannot skip creation of the dn chain as it's referred to in unchecked path. That only really matters with next PR as we start doing it all the time but it's also probably possible to pass an empty vec as opposed to None and trigger this today.
 - we were trying to delete too many chains, causing chain does not exist errors in teardown when they aren't created in setup
 - the logic to cleanup `/run/containers/networks/aardvark-dns/<network>` files was a bit broken, in particular when multiple lines belonged to a container and that was the last to shut down (which happens on dualstack network) we wouldn't correctly identify the new file as empty and didn't delete it. That caused aardvark to not shutdown when all configs are gone. This isn't really a problem normally but when playing with changing ports we need aardvark to be stopped.
 - the way we're handling subnets means we would only ever listen to one of the IPs configured for a container in /etc/resolv.conf.
This part is still not 100% coherent, resolv.conf content comes from ipam config whatever that is and we use gateway addresses for listening IPs here, but it's a start: at least with this aardvark listens to both ipv4 and ipv6 when both are set.

We don't actually need the distinction of gateway v4 or v6 as we just treat it as a string so I simplified that a bit as well and removed unneeded `mut` flags when I saw them. Hopefully that won't make a difference.

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -296,27 +296,22 @@ impl Aardvark {
     pub fn delete_entry(&mut self, container_id: String, network_name: String) -> Result<()> {
         let path = Path::new(&self.config).join(network_name);
         let file_content = fs::read_to_string(&path)?;
-        let lines: Vec<&str> = file_content.split('\n').collect();
+        let lines: Vec<&str> = file_content.split_terminator('\n').collect();
 
-        let mut idx = 1;
+        let mut idx = 0;
         let mut file = File::create(&path)?;
 
-        for line in &lines {
+        for line in lines {
             if line.contains(&container_id) {
-                if lines.len() <= 3 {
-                    // delete the file and return
-                    // since there is nothing in the network
-                    fs::remove_file(&path)?;
-                    return Ok(());
-                }
-                idx += 1;
                 continue;
             }
             file.write_all(line.as_bytes())?;
-            if idx < lines.len() {
-                file.write_all(b"\n")?;
-            }
+            file.write_all(b"\n")?;
             idx += 1;
+        }
+        // nothing left in file (only header), remove it
+        if idx <= 1 {
+            fs::remove_file(&path)?
         }
         Ok(())
     }
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -19,8 +19,7 @@ const SYSTEMD_RUN: &str = "systemd-run";
 #[derive(Clone, Debug)]
 pub struct AardvarkEntry {
     pub network_name: String,
-    pub network_gateway_v4: String,
-    pub network_gateway_v6: String,
+    pub network_gateway: String,
     pub container_id: String,
     pub container_ip_v4: String,
     pub container_ip_v6: String,
@@ -144,9 +143,24 @@ impl Aardvark {
 
         Ok(())
     }
-    pub fn commit_entries(&mut self, entries: Vec<AardvarkEntry>) -> Result<()> {
-        for entry in entries {
-            match self.commit_entry(entry.clone()) {
+    pub fn commit_entries(&self, entries: Vec<AardvarkEntry>) -> Result<()> {
+        for entry in &entries {
+            let path = Path::new(&self.config).join(&entry.network_name);
+            if !path.exists() {
+                // create file with all gateways IPs first, as that info will
+                // not be available to commit_entry()
+                let mut file = OpenOptions::new().write(true).create(true).open(&path)?;
+                let gateways = entries
+                    .clone()
+                    .into_iter()
+                    .filter(|e| e.network_name == entry.network_name)
+                    .map(|e| e.network_gateway)
+                    .collect::<Vec<String>>()
+                    .join(",");
+                file.write_all(gateways.as_bytes())?;
+                file.write_all("\n".as_bytes())?;
+            }
+            match self.commit_entry(entry) {
                 Err(er) => {
                     return Err(std::io::Error::new(
                         std::io::ErrorKind::Other,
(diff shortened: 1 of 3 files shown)


src/dns/aardvark.rs
```diff
@@ -160,25 +159,44 @@ impl Aardvark {
         Ok(())
     }
 
-    pub fn commit_entry(&mut self, entry: AardvarkEntry) -> Result<()> {
+    fn commit_entry(&self, entry: AardvarkEntry) -> Result<()> {
         let mut data: String;
         let path = Path::new(&self.config).join(entry.network_name);
         let file_exists = path.exists();
-        let mut file = OpenOptions::new().append(true).create(true).open(&path)?;
+        let mut file = OpenOptions::new()
```
Luap99, line 166: I think the main idea here was to increase performance by only appending. I see the you fix this dual stack bug but IMO it must be fixed already before that. IPv4 and v6 should be submitted at once.

If I look at https://github.com/containers/aardvark-dns/blob/3baa1d79c19c6def206e960a294093ef10f3cc1e/src/test/config/podman_v6_entries/podman_v6_entries this config file is clearly wrong.
It should be something like this if I understand the format correctly
```
10.89.0.1, fdfd:733b:dc3:220b::3
7b46c7ad93fcbcb945c35286a5ba19d6976093e2ce39d2cb38ba1eba636404ab 10.89.0.2 fdfd:733b:dc3:220b::2 test1,7b46c7ad93fc
88dde8a2489780d3c8c90db54a9a97faf5dbe4f555b23e27880ca189dae0e2b0 10.89.0.3 fdfd:733b:dc3:220b::3 test2,88dde8a24897
```

@flouthoc @mheon PTAL

flouthoc: > I think the main idea here was to increase performance by only appending. I see the you fix this dual stack bug but IMO it must be fixed already before that. IPv4 and v6 should be submitted at once.

Yes we agreed upon appending only logic for performance purpose.

> If I look at https://github.com/containers/aardvark-dns/blob/3baa1d79c19c6def206e960a294093ef10f3cc1e/src/test/config/podman_v6_entries/podman_v6_entries this config file is clearly wrong.
It should be something like this if I understand the format correctly

@Luap99 Testfile here is only added for a unit test testing of lookup logic, first line ensures spinning server on both ipv6 and ipv4 but its not necessary to provide both of them however we don't spin any server during unit testing it only tests resolution logic so https://github.com/containers/aardvark-dns/blob/3baa1d79c19c6def206e960a294093ef10f3cc1e/src/test/config/podman_v6_entries/podman_v6_entries should not be considered as standard so I'll rephrase the filename or amend it to match standards but i think its still a valid file cause aardvark server would still spin up server on provided `v4` ip and could contain mixed entries of `v4 and v6` which will be served on resolution logic ( again it should not be treated as standard )

Luap99: @flouthoc This is not just an example file.
It is exactly what netavark creates for a dual stack network `podman network create --ipv6`
```
cat /run/containers/networks/aardvark-dns/podman1
10.89.0.1
338947753ff897448fa754e8926b4ea68daa03840d7df26c8e839704f3f8f745 10.89.0.2  romantic_jang,338947753ff8
338947753ff897448fa754e8926b4ea68daa03840d7df26c8e839704f3f8f745  fd3b:2fee:9907:c91e::2 romantic_jang,338947753ff8
```
This is exactly the issue @martinetd is trying to fix here. The full config generation is wrong. The ID/names should never be duplicated instaed both IPv4/v6 should be on the same line.

Luap99:
@martinetd Can you add a test for the iptables changes.

martinetd:
> hmm I guess the checks don't currently check if we remove chains or not, I'll add that.

I had missed it, it's currently checked by this line:
```
    expected_rc=1 run_in_host_netns iptables -nvL NETAVARK-1D8721804F16F -t nat
```
So this makes sure we still remove chains that we created and I didn't break that.

The failure case this fixes is hard to test currently (I have no idea how to create a config that would consider an empty vec instead of None for port forwarding), but it'll be tested with the next PR anyway as that will no longer make a difference -- I don't think that warrants any new test.

src/dns/aardvark.rs
martinetd: Ok so I've updated this patch to check if file exists before appending the entry, so at that point we can just gather all gateways and write the first line first.
That adds an extra open() but I believe that's a cheap price to pay.
I agree grouping everything on a single line would be better, but that's not reasonable with the current syntax allowing multiple subnets. A better fix would change aardvark syntax to have a list of all IPs instead of ipv4/ipv6 as a csv, but that's more than I'd like to chew and it doesn't make that fix any less valid. Better solutions can be done later, perfect is the enemy of good and all that.

and if you'd let me be an old fart for a moment here, let me just say we're already rewriting the file on container shutdowns (can't remove a line without that), so doing it on startup isn't that much worse... And anyway rewriting the whole file or just appending literally have the same cost as long as the file is <4k (which most will be), so that's premature optimization to me.

Luap99:
> > @martinetd Can you add a test for the iptables changes.
> 
> hmm I guess the checks don't currently check if we remove chains or not, I'll add that. If I'm going to modify the test though I'd like to bring in my patch converting it from `iptables -L` to `iptables -S` (present in #323) so I won't need to do this twice -- I didn't get any comment on that, is it acceptable?

I am fine with using `-S` if you prefer that.

If there is no way to trigger that now I fine with merging. I just have to manually check the aardvark config changes since there is no way to test it in this CI, all dns tests are the aardvark repo (something we should fix so we can test these changes in netavark as well)

PR #332, opened 2022-07-08
rework isolate option to match CNI

### LINKED ISSUE #331: isolate nor working correctly
While testing cni isolate option against the netavark one I found some issues.

- IPv6 rules are not created, therefore ipv6 traffic is not isolated at all.
- Isolation blocks all external communication which is wrong, this is what internal is used for. If we want to follow what docker/cni does we should only block communication between two isolated networks.

This should be fixed before we make a new release.

Luap99:
The isolate option doe snot work as expected. It currently prevents all
traffic does is not on the same network. Instead it should only block
traffic between two isolated networks, never external connections.

Also it did not work for ipv6 at all.

Fixes #331

### DIFF
--- a/src/firewall/iptables.rs
+++ b/src/firewall/iptables.rs
@@ -1,10 +1,9 @@
 use crate::error::{NetavarkError, NetavarkResult};
 use crate::firewall;
 use crate::firewall::firewalld;
-use crate::firewall::varktables;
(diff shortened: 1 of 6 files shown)


Luap99:
@mheon PTAL

mheon:
LGTM

baude:
@Luap99 really  nicely done code, thanks for this contribution!

/lgtm

PR #338, opened 2022-07-13
Cirrus: enable Fedora 36 aarch64 tasks on EC2

lsm5:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -21,28 +21,24 @@ env:
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
     AARDVARK_DNS_BRANCH: "main"
     AARDVARK_DNS_URL: "https://api.cirrus-ci.com/v1/artifact/github/containers/aardvark-dns/success/binary.zip?branch=${AARDVARK_DNS_BRANCH}"
+    FEDORA_NETAVARK_AARCH64_AMI_ID: "ami-00d7ccde5eaacc50e"
 
 
 gcp_credentials: ENCRYPTED[d6efdb7d6d4c61e3831df2193ca6348bb02f26cd931695f69d41930b1965f7dab72a838ca0902f6ed8cde66c7deddae2]
 
-
-# Default VM to use unless set or modified by task
-gce_instance:  &standard_vm
-    image_project: "libpod-218412"
-    zone: "us-central1-c"
-    cpu: 2
-    memory: "4Gb"
-    disk: 200  # GB, do not set <200 per gcloud warning re: I/O performance
-    image_name: "${FEDORA_NETAVARK_IMAGE}"
+aws_credentials: ENCRYPTED[949b114326ca5686a3aa2990536a479cf8e5a931611b443e50579ea787faa21eb172f6fc0ebc79d7e817e4bebf417940]
 
 
 build_task:
   alias: "build"
   # Compiling is very CPU intensive, make it chooch quicker for this task only
-  gce_instance:
-    <<: *standard_vm
-    cpu: 8
-    memory: "8Gb"
+  gce_instance: &standard_build_gce_x86_64
+     image_project: "libpod-218412"
+     zone: "us-central1-c"
+     disk: 200  # GB, do not set <200 per gcloud warning re: I/O performance
+     cpu: 8
+     memory: "8Gb"
+     image_name: "${FEDORA_NETAVARK_IMAGE}"
   cargo_cache: &cargo_cache
     # Populating this cache depends on execution of setup.sh, and runner.sh
     # to builds of all release, debug, plus unit-tests.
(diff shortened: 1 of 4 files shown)


.cirrus.yml
```diff
@@ -113,43 +132,104 @@ validate_task:
...
+  main_script: *main
+
+
 # This task is critical.  It updates the "last-used by" timestamp stored
 # in metadata for all VM images.  This mechanism functions in tandem with
 # an out-of-band pruning operation to remove disused VM images.
 meta_task:
     alias: meta
     name: "VM img. keepalive"
+    gce_instance: *standard_gce_x86_64
```
cevich, line 232: Oh no, this is wrong.  The meta_task runs inside a specific container image.

lsm5: ah ok

lsm5: done.

Makefile
```diff
@@ -78,7 +78,7 @@ build_unit: $(CARGO_TARGET_DIR)
 # Test build cross-architecture
 # Ref: https://github.com/cross-rs/cross
 .PHONY: build_cross
-build_cross: build_cross.aarch64-unknown-linux-gnu build_cross.arm-unknown-linux-gnueabi
+build_cross: build_cross.arm-unknown-linux-gnueabi
```
cevich, line 81: Are you sure about this?  My understanding (could be wrong) is: `aarch64` means ARMv8 64-bit, and `arm` means ARMv7 32-bit.  My knowledge is based on playing around with Fedora on Raspbery-pi's a few years ago :wink:

cevich (approved):
This looks good.  You kept it close to the aardvark work which helps.  Just a minor nit and a question.  Otherwise LGTM assuming the tests are also green.

Makefile
cevich: Raspbery-Pi comment Clarification: Meaning my knowledge is old, possibly irreverent, and should be discarded quickly upon the slightest whiff of being incorrect :smile:

lsm5: yes, so this change is expected, right? We don't need aarch64 in the cross task anymore. We only care about the armv7 one unless we wanna send that off to ec2's armv7 instances?

.cirrus.yml
```diff
@@ -74,30 +69,54 @@ build_task:
   upload_caches: [ "cargo", "targets", "bin" ]
 
 
+build_aarch64_task:
+  alias: "build_aarch64"
+  ec2_instance: &standard_build_ec2_aarch64
+    image: ami-0f89dfa699fe590b0
```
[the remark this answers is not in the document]
lsm5: done. PTAL

Makefile
cevich: @flouthoc opinions?

lsm5:
@cevich lemme know if you got any ideas about the success task and CIRRUS_BUILD_ID. Else, maybe it's worthwhile to just change across the board to use rpms from rhcontainerbot/podman-next and get rid of the binary.zip part right away. I was hoping to delay that a bit.

lsm5:
@cevich Made some more changes to get rid of binary.zip in success task. If you're cool with the recent changes I can squash commits and re-push.

.cirrus.yml
```diff
@@ -194,27 +261,20 @@ success_task:
...
-    # Cross-compiled binary artifact is required by other CI systems
-    - curl --fail --location -O --url https://api.cirrus-ci.com/v1/artifact/build/${CIRRUS_BUILD_ID}/build_cross/binary.zip
-    - unzip binary.zip
-    - rm -f binary.zip
-    - mv bin/* ./
-    - rm -rf bin
-  # Upload tested binary for consumption downstream
-  # https://cirrus-ci.org/guide/writing-tasks/#artifacts-instruction
-  binary_artifacts:
-    path: ./*netavark*
```
cevich, line 220: Please leave this, it ensures we store the build artifacts when Cirrus runs on the branch, but only after all tasks are successful.

I'm not opposed to using your copr repos during setup of various repos (instead of downloading this zip) but I still want to produce the zip in case it's needed for some reason later.

contrib/cirrus/setup.sh
```diff
@@ -19,7 +19,11 @@ msg "************************************************************"
 msg "Setting up runtime environment"
 msg "************************************************************"
 
-dnf -y install make automake gcc gcc-c++ kernel-devel
+dnf -y copr enable rhcontainerbot/podman-next
+dnf -y install make automake gcc gcc-c++ kernel-devel aardvark-dns
```
cevich, line 23: This is fine, but please check c/automation_images if `make`, `automake`, `gcc`, `gcc-c++`, and `kernel-devel` are already installed in the fedora-netavark images.  If so, we don't need to install them here, if not please open a PR to add them over there.

(It's a relatively minor thing, so no need to actually incorporate new images)

contrib/cirrus/setup.sh
```diff
@@ -19,7 +19,11 @@ msg "************************************************************"
...
+
+# Remove netavark executable
+rm /usr/libexec/podman/netavark
```
cevich, line 26: I think this will fail if the file doesn't exist: Use the `-f` argument please.  Also see https://github.com/containers/automation_images/pull/149

contrib/cirrus/setup.sh
lsm5: will do

contrib/cirrus/setup.sh
lsm5: yes

lsm5: I left a few comments on the automation_images PR. PTAL there

.cirrus.yml
lsm5: done

contrib/cirrus/setup.sh
lsm5: done

contrib/cirrus/runner.sh
```diff
@@ -52,19 +58,33 @@ _run_verify_vendor() {
...
+    _run_integration
+}
+
 show_env_vars
 
 msg "************************************************************"
 msg "Toolchain details"
 msg "************************************************************"
+rustup install stable
+rustup default stable
```
cevich, line 87: Minor nit: If these lines are going to stay for now, it might be nice to add a comment pointing maintainers at https://github.com/containers/automation_images/issues/150

lsm5: done. PTAL

contrib/cirrus/setup.sh
```diff
@@ -19,7 +19,12 @@ msg "************************************************************"
 msg "Setting up runtime environment"
 msg "************************************************************"
 
-dnf -y install make automake gcc gcc-c++ kernel-devel
+# FIXME (@lsm5): These dependencies are better off installed in the
+# c/automation_images repo
+# https://github.com/containers/automation_images/issues/151
+if [[ $(uname -m) ]]; then
```
cevich, line 25: I don't understand this...won't it always be true?

contrib/cirrus/runner.sh
```diff
@@ -52,19 +58,37 @@ _run_verify_vendor() {
...
 show_env_vars
 
 msg "************************************************************"
 msg "Toolchain details"
 msg "************************************************************"
+
+# FIXME (@lsm5): rustup commands should be moved to c/automation_images repo
+# https://github.com/containers/automation_images/issues/150
+rustup install stable
+rustup default stable
```
cevich, line 90: I noticed in the Makefile for the old cross target, we were doing a:

`rustup target add aarch64-unknown-linux-gnu`, maybe that's what we're missing for arm?

contrib/cirrus/setup.sh
lsm5: ugh lol i forgot to add the equal aarch64. can't wait for the weekend

.cirrus.yml
```diff
@@ -21,28 +21,24 @@ env:
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
     AARDVARK_DNS_BRANCH: "main"
     AARDVARK_DNS_URL: "https://api.cirrus-ci.com/v1/artifact/github/containers/aardvark-dns/success/binary.zip?branch=${AARDVARK_DNS_BRANCH}"
+    FEDORA_NETAVARK_AARCH64_AMI_ID: "ami-02da46d8fa51e5316"
```
cevich, line 24: OH! This is the wrong image!  [ami-02da46d8fa51e5316 is the one for podman.  ](https://cirrus-ci.com/task/5118432189349888).  I found this by looking up the image in EC2 and noticed the `Name` tag "fedora-podman-aws-arm64-c5353795591864320".  This is why you're having packaging and compiler issues.  The image you want is `ami-00d7ccde5eaacc50e` which is for `fedora-netavark-aws-arm64` :grin:

cevich: Ugh, this is horrible.  I really need to do something to make that less confusing.  Otherwise every single `ami-*` looks just like every other `ami-*` :cry:   Why won't AWS just let us use the damn name!

.cirrus.yml
```diff
@@ -76,30 +72,39 @@ build_task:
...
+    # This simply prevents rebuilding bin/netavark for every subsequent task.
+    folder: "$CIRRUS_WORKING_DIR/bin"
+    fingerprint_key: "bin_v2_${CIRRUS_TAG}${DEST_BRANCH}${CIRRUS_PR}_aarch64" # Cache only within same tag, branch, or PR (branch will be 'pull/#')
+    reupload_on_changes: true
   setup_script: *setup
   main_script: *main
-  # Upload cross-compiled binary for collection by success task
-  # https://cirrus-ci.org/guide/writing-tasks/#artifacts-instruction
-  binary_artifacts:
-    path: ./bin/netavark*
```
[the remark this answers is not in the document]
cevich: Maybe update the comment though to delete "cross-".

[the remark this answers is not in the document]
lsm5: done

PR #342, opened 2022-07-13
fix integration read only /proc test

Luap99:
When there are no port mapping we do not have to set the sysctl.
This made it through CI because me merged two PR without rebasing.


flouthoc (approved):
LGTM
/lgtm
/hold

PR #308, opened 2022-06-15
netavark, dns: don't `double-fork` aardvark instead wait for the aardvark process to return.

flouthoc:
Netavark now does not double-forks aardvark-dns's server instead it
waits for aardvark-process to return back and success return means
aardvark-dns is ready to serve requests and now forking happens at
aardvark end.

See: https://doc.rust-lang.org/std/process/struct.Command.html#method.spawn

This needs: https://github.com/containers/aardvark-dns/pull/148

Should help in:
* https://github.com/containers/podman/issues/14173
* https://github.com/containers/podman/issues/14171

Alternative to: https://github.com/containers/netavark/pull/300

PR #344, opened 2022-07-14
Use latest aardvark-dns binary

baude:
Instead of using a packaged version of aardvark-dns, we grabbed the
latest main branch zip copy and use it instead.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,8 +17,10 @@ env:
     # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
     SCRIPT_BASE: "./contrib/cirrus"
     FEDORA_NAME: "fedora-36"
-    IMAGE_SUFFIX: "c6457378097856512"
+    IMAGE_SUFFIX: "c5353795591864320"
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
+    AARDVARK_DNS_BRANCH: "main"
+    AARDVARK_DNS_URL: "https://api.cirrus-ci.com/v1/artifact/github/containers/aardvark-dns/success/binary.zip?branch=${AARDVARK_DNS_BRANCH}"
 
 
 gcp_credentials: ENCRYPTED[d6efdb7d6d4c61e3831df2193ca6348bb02f26cd931695f69d41930b1965f7dab72a838ca0902f6ed8cde66c7deddae2]
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -24,18 +24,20 @@ dnf -y install make automake gcc gcc-c++ kernel-devel
 show_env_vars
 
 if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
-	# Setup short-name for rustembedded/cross
-	# TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
-	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
+    # TODO: Remove this when we're building natively, but remember to still
+    # collect the artifact in success_task for downstream CI use.
+    dnf reinstall -y netavark aardvark-dns
+    # Setup short-name for rustembedded/cross
+    # TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
+    echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
+else
+    req_env_vars AARDVARK_DNS_URL
+
+    set -x  # show what's happening
+    curl --fail --location -o /tmp/aardvark-dns.zip "$AARDVARK_DNS_URL"
+    mkdir -p /usr/libexec/podman
+    cd /usr/libexec/podman
+    rm -f aardvark-dns*
+    unzip -o /tmp/aardvark-dns.zip
+    chmod a+x /usr/libexec/podman/aardvark-dns
 fi
-
-req_env_vars AARDVARK_DNS_URL
-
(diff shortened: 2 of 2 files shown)


baude:
this should fix #308

.cirrus.yml
```diff
@@ -19,6 +19,7 @@ env:
     FEDORA_NAME: "fedora-36"
     IMAGE_SUFFIX: "c6457378097856512"
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
+    AARDVARK_DNS_URL: "https://api.cirrus-ci.com/v1/artifact/github/containers/aardvark-dns/success/binary.zip?branch=${NETAVARK_BRANCH}"
```
cevich, line 22: Would you mind renaming & defining `NETAVARK_BRANCH` -> `AARDVARK_BRANCH`?

contrib/cirrus/setup.sh
```diff
@@ -28,3 +28,14 @@ if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
 	# TODO: We can move this to quay.io if we reach rate-limits, hopefully that's not gonna happen for netavark
 	echo '  "rustembedded/cross" = "docker.io/rustembedded/cross"'  >> /etc/containers/registries.conf.d/000-shortnames.conf
 fi
+
+req_env_vars AARDVARK_DNS_URL
+
```
cevich, line 33: Below looks good, and let's avoid any ambiguity by adding something like:

```
# Make sure these are NOT present on the system, since we want to test recent versions built from source
rpm -e --nodeps aardvark-dns netavark || true
```

contrib/cirrus/setup.sh
```diff
@@ -28,3 +28,14 @@ if [[ "$CIRRUS_TASK_NAME" == "build_cross" ]]; then
...
+set -x  # show what's happening
+curl --fail --location -o /tmp/aardvark-dns.zip "$AARDVARK_DNS_URL"
+mkdir -p /usr/libexec/podman
+cd /usr/libexec/podman
+rm aardvark-dns
```
cevich, line 38: This will fail if the file isn't there.  Maybe do `rm -f aardvark-dns*` to be sure.

contrib/cirrus/setup.sh
baude: i think we do want them on the system because they drag in the proper dependencies; otherwise this gets unpredicatable when new deps are added.  We can and do just overwrite things.

cevich: Hmmm, it's a fair point.  My concern is other than test-failures, developers will have no assurances tests are **REALLY** running with the intended binaries if ambiguity is possible.  Hmmm.

Maybe the middle-ground is to do similar to podman CI: Install the RPMs (resolving deps) then at runtime going through the package's file-listing and clobber them.  (I can work on this.)

Otherwise we're entirely relying on tests (and humans using get_ci_vm) to 100% flawlessly never ever ever accidentally  use the packaged binary :frowning:

cevich: Okay this is now done in the latest https://github.com/containers/automation_images/pull/149 .  You're welcome to wait for the GCE & AWS images to build or just go ahead with this PR.  Note: We ***do*** definitely need the download code you added below.

flouthoc:
LGTM just waiting for unresolved comments and discussion above.

cevich (changes requested):
If we leave `AARDVARK_DNS_BRANCH` undefined Cirrus will default to `main` which will silently do the wrong thing when a new release-branch is created in the future.

Since you'll need to re-run the tests again anyway, I've got newly built images with the netavark/aardvark-dns packages installed + binaries stripped out.  It's up to you if you want to incorporate them here, or I'm fine if that happens in another PR:

* GCE image ID is c5353795591864320
* fedora-aws is ami-09e5978f141900855
* fedora-netavark-aws-arm64 is ami-00d7ccde5eaacc50e
* fedora-podman-aws-arm64 is ami-02da46d8fa51e5316

cevich:
Sorry @baude these kinds of changes can be so finicky to get right.  I do appreciate you're taking the initiative to open this PR and start the work.  With the `AARDVARK_DNS_BRANCH` definition this is fully LGTM and ready to go in.

Updating the images is optional, and arguably risky.  Totally fine if that happens in another PR.

cevich (approved):
/lgtm

flouthoc:
Is it missing `containernetworking-plugins` in new image ?

Luap99:
No it is missing netavark which makes podman fall back to cni which then errors because cni is missing. I don't understand why we exclude the netavark rpms.
Either way the cross build does not need special networking, can we add `--network host`?

flouthoc:
> > Either way the cross build does not need special networking, can we add --network host?
> 
> It's using the cross-rs tool to call podman, not sure if you can specify options like that. I say we just ignore the failure and manually merge, since @lsm5 PR can probably get merged really soon.

SGTM if others are cool with it.

flouthoc (approved):
LGTM

Looks like some one needs to do admin merge since one check is red.

cevich:
> Looks like some one needs to do admin merge since one check is red.

@lsm #338 is rebased on this PR, so I think if I merge that one it will take care of this one.

flouthoc:
Cool lets wait for #338 then

cevich:
Ugh, so 338 is going to need some more time.  I've got an idea...

flouthoc:
@cevich Looks like total success depends on build_cross, we gotta remove that dependency as well.

cevich:
Oof, isn't that what I just did?  Hrmmm. What did I miss.

cevich:
right.  aardvark CI needs that :rofl:

cevich:
Okay, so hopefully that will get it.  Once again for the third-time :confounded:

cevich:
Okay, that's going to work.  Anyone: please feel free to merge this assuming it passes and looks good.  I've got to get some lunch.

PR #350, opened 2022-07-17
Fix Makefile for docs and main

hoshsadiq:
The Makefile in the docs directory was unable to find the `mandown` binary despite it being in the PATH. Exporting it fixes it.

Also `[[ ... ]]` does not work with `sh`, which `make` uses by default. Using `[ ... ]` works fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ build: bin $(CARGO_TARGET_DIR)
 .PHONY: clean
 clean:
 	rm -rf bin
-	if [[ "$(CARGO_TARGET_DIR)" == "targets" ]]; then rm -rf targets; fi
+	if [ "$(CARGO_TARGET_DIR)" = "targets" ]; then rm -rf targets; fi
 	$(MAKE) -C docs clean
 
 .PHONY: docs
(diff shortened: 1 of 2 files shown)


flouthoc (approved):
LGTM
Could you please sign your commits.

hoshsadiq:
Done

Makefile
```diff
@@ -103,7 +103,7 @@ vendor-rm-windows:
 .PHONY: vendor-tarball
 vendor-tarball: build vendor
 	VERSION=$(shell bin/netavark --version | cut -f2 -d" ") && \
-	[[ $$VERSION == *-dev ]] && echo "version ends with -dev" && exit 1; \
+	[ $$VERSION = *-dev ] && echo "version ends with -dev" && exit 1; \
```
Luap99, line 106: This does not work:
`[ 1.4.1-dev = *-dev ] && echo yes`

Luap99 (changes requested):

Makefile
hoshsadiq: TIL! That's now fixed by doing a switch statement.

Luap99:
Please squash the commits.

I think [aardvark-dns](https://github.com/containers/aardvark-dns) has the same problem, will you fix it there as well?

hoshsadiq:
Done! I'll try to do aardvark-dns soon.

PR #353, opened 2022-07-18
Cirrus: Verify expected binary artifacts present

cevich:
Other CI systems depend/expect a full set of compiled binaries are
included in the success task's binary artifact (zip file).  However,
it's possible some mistake in CI could silently lead to one or more
binaries not being produced.  For example, a `bin_cache` failure.  Fix
this by verifying all expected/required files are present.  If not,
print the name of the missing binary to assist in
troubleshooting/debugging.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -99,10 +99,12 @@ build_aarch64_task:
   # However, we don't want to confuse architectures.
   art_prep_script:
     - cd bin
-    - ln netavark netavark.$(uname -m)-unknown-linux-gnu
-    - ln netavark.debug netavark.debug.$(uname -m)-unknown-linux-gnu
+    - ls -la
+    - mv netavark netavark.$(uname -m)-unknown-linux-gnu
+    - mv netavark.debug netavark.debug.$(uname -m)-unknown-linux-gnu
+    - mv netavark.info netavark.info.$(uname -m)-unknown-linux-gnu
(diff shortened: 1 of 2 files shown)


cevich:
@lsm5 PTAL when you have a moment

flouthoc (approved):
LGTM

.cirrus.yml
```diff
@@ -285,17 +287,28 @@ success_task:
...
-    # Cross-compiled binary artifact is required by other CI systems
-    - curl --fail --location -O --url https://api.cirrus-ci.com/v1/artifact/build/${CIRRUS_BUILD_ID}/build_aarch64/armbinary.zip
+    - set -x
+    - curl --fail --location -O --url ${API_URL_BASE}/build_aarch64/armbinary.zip
     - unzip armbinary.zip
     - rm -f armbinary.zip
     - mv bin/* ./
     - rm -rf bin
+  artifacts_test_script:  # Other CI systems depend on all files being present
+    - ls -la
```
flouthoc, line 309: Are we using output of this `ls -la` anywhere or is it only for debugging.

cevich: Yes it shows the files that are actually present, since they are renamed in the build task. It makes it easier to diagnose any test failure, as opposed to requiring a developer to manually download the artifact zip to check.

PR #359, opened 2022-07-26
Bump to v1.1.0

mheon:
No branch, since we only seem to do RHEL branches on this repo.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,11 @@
 # Release Notes
 
+## v1.1.0
+* Netavark is now capable of starting Aardvark on a port other than 53 (controlled by `dns_bind_port` in `containers.conf`). Firewall rules are added to ensure DNS still functions properly despite the port change.
+* Added the ability to isolate networks. Networks with the isolate option set cannot communicate with other networks with the isolate option set.
+* Improved the way Aardvark is launched to avoid potential race conditions where DNS would not be ready when containers were started.
+* Fixed a bug where Aardvark could not be run in environments with a read-only `/proc` (e.g. inside a container).
+
 ## v1.0.3
 * Updated dependenciess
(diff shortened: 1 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -1,5 +1,11 @@
 # Release Notes
 
+## v1.1.0
+* Netavark no longer starts Aardvark bound to port 53, but instead uses a high port and forwards traffic from 53 to this port using firewall rules. This allows the host to use Port 53 without conflict.
```
Luap99, line 4: I thought the default is still 53 and you have to change it in cotnainers.conf?

Luap99: Yes https://github.com/containers/netavark/blob/eccad5c2235d38d239ff24a7c9044ed245c56fc2/src/commands/setup.rs#L74-L85

There is no way to choose a random one without storing it somewhere.

Luap99: I mean this should not be part of netavark release notes, netavark uses NETAVARK_DNS_PORT or defaults to 53.
If podman overwrites that we should document it there.

Please change it

mheon: Repushed. I want to mention this here since the core functionality is in Netavark and we should have a pointer that v1.1.0 is the earliest supporting version, but I changed the release note to make it clear this is optional.

RELEASE_NOTES.md
```diff
@@ -1,4 +1,10 @@
 # Release Notes
+65;6800;1c
```
Luap99, line 2: typo?

mheon: Damn it. Fixed & force-pushed.

Luap99:
@mheon don't forget to push that tag :)

mheon:
Oops. Can do.

PR #360, opened 2022-07-28
create network driver interface

Luap99:
I am sorry, I know that this is almost impossible to review properly but I couldn't
figure out how to split this out in smaller pieces.

We should abstract the internal driver logic behind a interface. This
makes to code easier to maintain since each driver is isolated.

Also this will allow us much better error handling, in case of errors it
is important to still clean up as much as possible, i.e. when we setup
two networks and the second one fails we should teardown the first as
well.

setup routing sysctl only once
It is enough to only setup the sysctl once.

add test for static mac address

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -2,22 +2,18 @@
 use crate::dns::aardvark::Aardvark;
 use crate::error::{NetavarkError, NetavarkResult};
 use crate::firewall;
-use crate::firewall::iptables::MAX_HASH_SIZE;
 use crate::network;
-use crate::network::core_utils::CoreUtils;
-use crate::network::internal_types::{PortForwardConfig, SetupNetwork};
-use crate::network::types::Subnet;
+use crate::network::driver::{get_network_driver, DriverInfo};
 use crate::network::{core_utils, types};
 use clap::Parser;
-use log::{debug, info};
+use log::{debug, error, info};
 use std::collections::HashMap;
 use std::env;
-use std::fs;
-use std::net::IpAddr;
+use std::fs::{self, File};
+use std::os::unix::prelude::AsRawFd;
 use std::path::Path;
 
 const IPV4_FORWARD: &str = "net.ipv4.ip_forward";
-const IPV6_FORWARD: &str = "/proc/sys/net/ipv6/conf/all/forwarding";
 
 #[derive(Parser, Debug)]
 pub struct Setup {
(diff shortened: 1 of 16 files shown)


Luap99:
@mheon @flouthoc PTAL

mheon:
I like the approach, and the core interface looks good. Have not had a chance to look at the larger scope of the changes, but it's mostly code moving around, right? Not freshly-written?

Luap99:
> I like the approach, and the core interface looks good. Have not had a chance to look at the larger scope of the changes, but it's mostly code moving around, right? Not freshly-written?

Yes the actual netlink stuff is still the same, aardvark setup also but I plan to integrate this more into the interface in follow up PRs.

I changed that we only open the netns fd once and pass it around and changed a bunch of references, e.g. `String` vs `&str`, to make the compiler happy and prevent unnecessary clones.

Things to fix in follow ups:
- use single netlink connection (or two for host and container ns)
- get rid of the async stuff
- fix weird/broken aardvark config generation
- fix error handling to cleanup as much as possible, parts are already done in this PR but there is still more to do.

flouthoc:
On initial look it looks great to me. Will give another detailed review tommorow.

Luap99:
Any update?
I want to get this merged this week since I am on PTO the next four weeks.

src/commands/setup.rs
```diff
@@ -282,8 +161,8 @@ impl Setup {
 
             if let Err(er) = aardvark_interface.commit_netavark_entries(
                 network_options.container_name,
-                network_options.container_id,
-                network_options.networks,
+                network_options.container_id.clone(),
```
flouthoc, line 164: Why do we need to clone these now.

src/network/internal_types.rs
```diff
@@ -28,9 +28,7 @@ pub struct TearDownNetwork {
 }
 
 #[derive(Clone, Debug)]
-pub struct PortForwardConfig {
-    //  network object
-    pub net: types::Network,
+pub struct PortForwardConfig<'a> {
```
flouthoc, line 31: Why is `lifetime` added, do we need this or is this added as a precaution ?

Luap99: yes we need this, the linter will complain if this was added when it is not required.

src/commands/setup.rs
Luap99: we have to clone it otherwise we cannot use it again
I will rework the aardvark setup next

flouthoc:
@Luap99 overall LGTM did you get a chance to run `podman` integration on these changes ?

Luap99:
> @Luap99 overall LGTM did you get a chance to run `podman` integration on these changes?

I can check if you want but I given that the test passes here I see no reason why it should fail in podman

flouthoc:
> > @Luap99 overall LGTM did you get a chance to run `podman` integration on these changes?
> 
> I can check if you want but I given that the test passes here I see no reason why it should fail in podman

@Luap99 if its easy then sure, since PR is big so could miss a few things while reviewing but `e2e` integration with `podman` would be nice otherwise I can try running it tommorow with `podman` and will merge it.

Luap99:
I run the podman tests and found one issue, should be fixed with the last commit.

PR #369, opened 2022-08-04
firewall types: remove unnecessary clones

Luap99:
Restructure the types so that we do not need to clone the data.

firewall: PortForwardConfig use IpNet instead of Subnet
We are only interested in a single IpNet so no need to keep the full
Subnet info.


Luap99:
@baude @flouthoc PTAL

baude:
LGTM

PR #370, opened 2022-08-05
rework aardvark setup 

Luap99:
Based on #369 

The driver already has all information so the aardvark code should not
have to pass it again. This simplifies the code and creates smaller
aardvark config files since it does not duplicate the contianer id and
names just to set more then one ip.

This does not touch the aardvark teardown logic.

Requires https://github.com/containers/aardvark-dns/pull/183

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -91,6 +91,7 @@ impl Setup {
             let mut driver = get_network_driver(DriverInfo {
                 firewall: firewall_driver.as_ref(),
                 container_id: &network_options.container_id,
+                container_name: &network_options.container_name,
                 netns_container: ns_fd,
                 network,
                 per_network_opts,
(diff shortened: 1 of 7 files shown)


mheon:
LGTM on my end - looks like an overall simplification of the code, and seems more logical.

flouthoc (approved):
code LGTM
/lgtm
/hold need to verify if CI is pulling latest aardvark and run against podman integration once.

baude:
lets get this in ... @flouthoc were you able to verify?

flouthoc:
ah i missed manually verifying this with podman integration my bad but lets get this in cause i think @Luap99 already did so, @Luap99 could you rebase since some new aardvark changes went in and it would be nice to rerun CI with latest aardvark and lets merge this.

flouthoc (approved):
LGTM
/lgtm

PR #375, opened 2022-08-09
aardvark,commit: acquire fs lock when performing commit to avoid `race` across parallel invocations.

flouthoc:
* We should avoid overriding configs when another instance of aardvark
  is trying to commit configs on the same path.

* On certain system a race exists where more than one aardvark instance
  are started in the frame where one instance has not yet completed
  updating its `aardvark.pid` causing more than one instance to get
  started and eventually causing conflits on the requested ports.

Some conditions are reported on `low power hardware which tries to start
a significant amount of containers` and it looks like the case matches with what is described in the second point.

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -1,4 +1,5 @@
 use crate::network::types;
+use fs2::FileExt;
 use nix::sys::signal::{self, Signal};
 use nix::unistd::Pid;
 use std::collections::HashMap;
@@ -14,6 +15,7 @@ use std::process::{Command, Stdio};
 
 const SYSTEMD_CHECK_PATH: &str = "/run/systemd/system";
 const SYSTEMD_RUN: &str = "systemd-run";
+const AARDVARK_COMMIT_LOCK: &str = "aardvark.lock";
 
 #[derive(Clone, Debug)]
 pub struct AardvarkEntry {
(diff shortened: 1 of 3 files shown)


src/dns/aardvark.rs
```diff
@@ -144,6 +146,36 @@ impl Aardvark {
         Ok(())
     }
     pub fn commit_entries(&self, entries: Vec<AardvarkEntry>) -> Result<()> {
+        // Acquire fs lock to ensure other instance of aardvark cannot commit
+        // or start aardvark instance till already running instance has not
+        // completed its `commit` phase.
+        let lockfile_path = Path::new(&self.config)
```
flouthoc, line 152: Path for lock file is intentionally `<run-root-base>/networks/aardvark.lock` instead of `<run-root-base>/networks/aardvark-dns/aardvark.lock` to make this patch backward compatible with older `aardvark` versions.

flouthoc:
@baude @mheon PTAL, could we ask the folks to try this patch on top of `netavark` `v1.1.0` and `aardvark` `v1.1.0`. 

I have played while trying to run multiple containers in parallel but it is exactly like what is mentioned in the issue that `race` is very rare and seen on `low power hardware` ( I think on the `edge` platforms ) for which the issue was reported against.

So while I was not able to reproduce the actual issue but I think following patch should close it.

/hold till we get ackd.

baude:
Confirmed this fixed the issue reported ... ready for merging.

LGTM

baude:
@mheon @Luap99 please review and merge

TomSweeneyRedHat:
Addresses: https://bugzilla.redhat.com/show_bug.cgi?id=2116481  Once this BZ merges, please assign the BZ to Jindrich.

src/dns/aardvark.rs
flouthoc: cc @Luap99 @mheon just fyi we will have a lock file now for aardvark commits, but we can change path later if there is any disagreement on this.

Luap99:
Podman already locks all operations in the network backend so there should never be two different netavark processes at any given time.

I think the problem here is that we do not wait for aardvark setup. Currently we return before the child writes its pid. So the actual fix is to make aardvark parent wait until the child wrote the pid file.
I don't think this can fix the double start issue at all, it just made it less likely.

baude:
@flouthoc please consider @Luap99 feedback ... and if in agreement, submit a pr

flouthoc:
@Luap99 Lets discuss this here: https://github.com/containers/aardvark-dns/pull/220

PR #399, opened 2022-08-31
Packit: initial enablement

lsm5:
This commit will run COPR builds on every PR against all active
releases of CentOS Stream and Fedora, thus allowing buildability checks before the
PR merges.

Builds are done on a custom COPR project:
`rhcontainerbot/packit-builds`.
Ref: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/

The build targets are set in the copr itself, so we don't need to
explicitly mention them in `.packit.yaml`, making upstream configuration
a lot simpler.

The `spec.rpkg` file meant for rpm builds post-pr-merge at
`rhcontainerbot/podman-next` copr gets reused for packit builds, so the
packit jobs are independent of Fedora / CentOS dist-git.

### DIFF
--- a/netavark.spec.rpkg
+++ b/netavark.spec.rpkg
@@ -31,7 +31,9 @@ Source: {{{ git_dir_pack }}}
 BuildRequires: make
 BuildRequires: cargo
 BuildRequires: git-core
-BuildRequires: golang-github-cpuguy83-md2man
+BuildRequires: go-md2man
+BuildRequires: protobuf-c
+BuildRequires: protobuf-compiler
 Recommends: aardvark-dns
 Provides: container-network-stack
 %if 0%{?fedora}
@@ -45,8 +47,7 @@ OCI network stack.}
 %description %{_description}
 
 %prep
-#{{{ git_dir_setup_macro }}}
-%autosetup -Sgit -T -b 0 -n %{name}
+{{{ git_dir_setup_macro }}}
 
 %build
 %{__make} CARGO="%{__cargo}" build
--- a/.packit.sh
+++ b/.packit.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+# Packit's default fix-spec-file often doesn't fetch version string correctly.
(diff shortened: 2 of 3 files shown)


edsantiago:
LGTM but it's not actually passing. Logs are too hard for me to read (I will learn, just not today)

lsm5:
> LGTM but it's not actually passing. Logs are too hard for me to read (I will learn, just not today)

I was gonna tag you after I fixed all the build issues lol :D.

So, main thing is netavark and aardvark use an additional Source1 tarball which contains the vendored deps along with some magic in the %prep section that I need to deal with.

I'm thinking of adding a new Makefile target that will allow vendored tarball creation for any commit, not just the release ones, get that Makefile PR merged first and then work on this one. (Same for aardvark)

lsm5:
so there were some successful runs in the past https://copr.fedorainfracloud.org/coprs/packit/containers-netavark-399/build/4803972/ . Don't know why it's messing up now.

Luap99:
@lsm5 Same question here, do we still need this?

lsm5:
> @lsm5 Same question here, do we still need this?

yes as well

lsm5:
@cevich @edsantiago @Luap99 PTAL

netavark.spec.rpkg
```diff
@@ -45,8 +47,7 @@ OCI network stack.}
 %description %{_description}
 
 %prep
-#{{{ git_dir_setup_macro }}}
-%autosetup -Sgit -T -b 0 -n %{name}
+{{{ git_dir_setup_macro }}}
```
lsm5, line 50: FYI, `rpkg spec` in .packit.yaml expands this to `%setup foo bar baz` so that's why the last in .packit.sh checks for `%setup` and replaces with autosetup.

cevich (approved):
It's mostly all "magic" to me, but I gave a look-through and don't see anything surprising.

LGTM

edsantiago:
LGTM, thanks for the cleanup.

PR #412, opened 2022-09-19
zbus update and a fix

zeenix:
Hi there,

zbus maintainer here. :wave: I wanted to help out a bit with zbus part of this project so providing this PR.


flouthoc:
Hi @zeenix , Thank you so much, PR looks good but CI is red.

zeenix:
> Hi @zeenix , Thank you so much, PR looks good but CI is red.

Thanks for the quick response. `make test` was fine on my Fedora machine. I'll check..

mheon:
Not really? The API process for firewalld is very synchronous - we can't do much while waiting for responses from the server.

Thanks for this, by the way - I've really been neglecting the firewalld/dbus bits of the project.

Luap99:
The main problem with zbus is the MSRV requirement, which makes it impossible to build on stock debian/ubuntu, see https://github.com/containers/netavark/pull/355 and the ubuntu20_build job

zeenix:
> Not really? The API process for firewalld is very synchronous - we can't do much while waiting for responses from the server.

Ah ok. There is [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.spawn_blocking) API for turning blocking calls to async but I'm not familiar enough with the code to say if that's worth it.

> Thanks for this, by the way - I've really been neglecting the firewalld/dbus bits of the project.

No worries.

> The main problem with zbus is the MSRV requirement, which makes it impossible to build on stock debian/ubuntu, see #355 and the ubuntu20_build job

Most of the MSRV bumps were not planned ones but we [tackled the issue of unintentional bumps](https://gitlab.freedesktop.org/dbus/zbus/-/merge_requests/535) a month back. What's left now is to decide how recent an MSRV we can require w/o breaking things for debian. If 6 months old releases are OK then that's good enough for me as well.

zeenix:
> The main problem with zbus is the MSRV requirement, which makes it impossible to build on stock debian/ubuntu, see #355 and the ubuntu20_build job

I'm having a bit of difficulty figuring out the latest MSRV that would be good for debian. Could you please let me know which version is that and I'll see if I can port to that version?

Luap99:
For ubuntu 2004 at least it seems to be 1.59 at the moment: https://packages.ubuntu.com/focal/rustc.
Debian stable seems to shippe a much older version with 1.48: https://tracker.debian.org/pkg/rustc.
I don't think we have to support debian given that the version is very old but v1.59 seems reasonable to me.

zeenix:
> I don't think we have to support debian given that the version is very old but v1.59 seems reasonable to me.

Agreed and I've a much better chance of succeeding. :) I'll see what I can do.

zeenix:
I briefly looked into what 1.60 features zbus depends on and one that seems impossible to port back from (w/o it being a breaking change) is [use of `dep:` features](https://gitlab.freedesktop.org/dbus/zbus/-/blob/main/zbus/Cargo.toml#L24) introduced in cargo 1.60. Any ideas?

Luap99:
> I briefly looked into what 1.60 features zbus depends on and one that seems impossible to port back from (w/o it being a breaking change) is [use of `dep:` features](https://gitlab.freedesktop.org/dbus/zbus/-/blob/main/zbus/Cargo.toml#L24) introduced in cargo 1.60. Any ideas?

Can we just wait for ubuntu to update the version? Looks like 22.10 ships with rust v1.61 so I guess we could wait and see if they will backport it to the other ubuntu versions as well.

zeenix:
> Can we just wait for ubuntu to update the version? Looks like 22.10 ships with rust v1.61 so I guess we could wait and see if they will backport it to the other ubuntu versions as well.

Seems [they've done that already](https://packages.ubuntu.com/focal/rust-all). I rebased again my work (was trivial actually) so the Ubuntu20 CI jobs still fails, maybe it needs an update?

zeenix:
oh, I had forgotten that the issue wasn't just that they don't ship Rust 1.60 but rather that they don't ship zbus 3. Since they do have the MSRV for zbus 3, they should be able to upgrade now.

zeenix:
@Luap99 I feel like there is a bit of a[ chicken&egg situation](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1013729#37) involved if you want to only depend on rust crates already packaged in the distro.

Wouldn't it be better to leave that problem to distros themselves and just ensure that this project along w/ all its deps can be built with Rust tools available in the distro in question?

Luap99:
> @Luap99 I feel like there is a bit of a[ chicken&egg situation](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1013729#37) involved if you want to only depend on rust crates already packaged in the distro.
> 
> Wouldn't it be better to leave that problem to distros themselves and just ensure that this project along w/ all its deps can be built with Rust tools available in the distro in question?

Wait, the debian/ubuntu cargo install does not use the official crates.io index? Do they only pull from debian sources? If this is the case I agree we should not wait for them. I only care about a to fast moving MSRV.

Luap99:
Hmm, maybe this error is caused by https://github.com/rust-lang/cargo/issues/10623?

Luap99:
The latest version should be 1.61 in ubuntu 2004, however our image is not updating on quay.io? https://quay.io/repository/libpod/ubuntu20rust
@baude PTAL

zeenix:
> Wait, the debian/ubuntu cargo install does not use the official crates.io index?

I am not sure about the details but the CI clearly shows that current failure is about too old zbus. Otherwise, we'd have this resolved 22 days ago. :)

zeenix:
> Hmm, maybe this error is caused by [rust-lang/cargo#10623](https://github.com/rust-lang/cargo/issues/10623)?

Ah, makes sense.

zeenix:
> I only care about a to fast moving MSRV.

As we have it checked in the CI now in zbus, accidental bumps hopefully would never happen anymore and I don't intend to bump to too new MSRV, no matter how appealing a new feature might be. :) I promise you that. :) I think a 6 months grace period is good enough in the Rust world but I'll still first check availability in Ubuntu LTS before bumping.

Luap99:
Yes we should be good in the future, I look into why the quay.io auto builder did not rebuild the image so far.

In any case could you rebase this PR, we no longer use tokio for netlink and just make blocking calls.

zeenix:
> In any case could you rebase this PR

Sure.



> we no longer use tokio for netlink and just make blocking calls

Interesting.

Luap99 (approved):
LGTM

zeenix:
Yay!

PR #417, opened 2022-09-27
refactor, clippy: edit syntax as per `clippy` standards

flouthoc:
Makes CI validation green again by applying following changes

* clippy: edit expressions that could be replaced by the question mark operator.
* clippy: use if let instead of single match

Ref: https://rust-lang.github.io/rust-clippy/master/index.html#question_mark
Ref: https://rust-lang.github.io/rust-clippy/master/index.html#single_match

### DIFF
--- a/src/network/core.rs
+++ b/src/network/core.rs
@@ -88,14 +88,13 @@ impl Core {
                 ));
             }
 
-            if let Err(err) = core_utils::CoreUtils::configure_netns_interface_async(
+            core_utils::CoreUtils::configure_netns_interface_async(
                 &container_veth_name_clone,
                 netns_ipaddr_clone,
                 &gw_ipaddr_clone,
                 static_mac_clone,
-            ) {
-                return Err(err);
-            }
+            )?;
+
             debug!(
                 "Configured static up address for {}",
                 container_veth_name_clone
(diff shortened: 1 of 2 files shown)


flouthoc:
@cevich PTAL this should make CI green.

flouthoc:
So for me `rustup update` updates all components including clippy, I think whenever we add a stable toolchain, i think clippy as a component is also updated here: https://github.com/containers/netavark/blob/main/contrib/cirrus/runner.sh#L84

But I need to confirm and figure out a way to lock individual component.

Pros and cons of locking `clippy` and not updating it with toolchain:

Cons
* Maintainers can easily forget to update clippy and upon manual update it can create a bigger hammer to edit lot of code at once.

Pros:
* CI never gets blocked.

baude:
LGTM

cevich:
@flouthoc another idea is (and resembles the policy for podman/buildah/skopeo CI):  Install the latest/greatest available at the time of VM image build.  That's an activity which happens at a minimum every 6-months (i.e. Fedora Release cycle), but often more frequently.  This removes some of the human-memory-factors, but (based on history) makes integrating new VM images harder.  Though...this is an ongoing challenge most of us have a grasp of.

PR #429, opened 2022-09-30
netlink rework to use low level sync API

Luap99:
Instead of doing a lot of async calls that we cannot use anyway we
call the netlink socket directly with blocking io.

This should allow us for more fine grade control over the netlink
messages and gives much better performance and the binary size
decreased by over 1 MB.

TBD: fix very ugly error messages and provide better error context in
general.


Luap99:
@baude @mheon @flouthoc PTAL This is still WIP because I have to fix error messages but this should give you a general idea of how it looks. I know this is very hard to review and I very sorry for that but I didn't see a better way since I had to remove the async stuff.

Luap99:
> @Luap99 I gave a quick look, overall looks good, did you see any performance improvement moving from async to sync ?

Yes, see the first commit with the test script. 10 msec faster for the simple bridge setup which is a 15-20% improvement. If you profile it you will see that by far the lowest part is the iptables setup due the many fork/exec. I think we can get rid of that with @mheon firewalld and nftables work.

> Change looks big enough, could you check with buildah and podman integration test once ?

Yes defnitly

mheon:
I did a general perusal of the code and I like what I see - and the diffstat is very encouraging!

Don't have time to do a more in-depth review but I'm fully in favor of what I've seen.

Luap99:
OK I think the test here should pass now but I still have to test against podman.

This is the current performance improvements `perf stat`:
```
Performance counter stats for './bin/netavark -f ./test/testfiles/simplebridge.json setup /proc/61255/ns/net':

             41.70 msec task-clock                #    0.951 CPUs utilized          
               142      context-switches          #    3.406 K/sec                  
                 5      cpu-migrations            #  119.914 /sec                   
             5,165      page-faults               #  123.872 K/sec                  
       137,994,997      cycles                    #    3.310 GHz                    
       153,819,338      instructions              #    1.11  insn per cycle         
        31,546,726      branches                  #  756.582 M/sec                  
         1,022,081      branch-misses             #    3.24% of all branches        

       0.043864514 seconds time elapsed

       0.019482000 seconds user
       0.023931000 seconds sys
```
vs current main:
```
 Performance counter stats for 'bin/netavark-old -f ./test/testfiles/simplebridge.json setup /proc/65446/ns/net':

             52.23 msec task-clock                #    0.895 CPUs utilized          
               525      context-switches          #   10.051 K/sec                  
                97      cpu-migrations            #    1.857 K/sec                  
             5,278      page-faults               #  101.045 K/sec                  
       179,101,185      cycles                    #    3.429 GHz                    
       170,088,478      instructions              #    0.95  insn per cycle         
        35,094,598      branches                  #  671.874 M/sec                  
         1,122,272      branch-misses             #    3.20% of all branches        

       0.058371980 seconds time elapsed

       0.027518000 seconds user
       0.026843000 seconds sys
```
Take a look with `strace -c` you can see the most time is spend waiting on iptables, I think switching to nftables will give a big performance boost.
```
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ------------------
 77.94    0.022365         657        34           wait4
  4.99    0.001432         119        12           sendto
  3.40    0.000977          27        36           clone3
  2.06    0.000592           3       182           close
  1.58    0.000452          10        44           openat
  1.57    0.000450           2       206           prlimit64
  1.31    0.000377           2       137        16 read
  1.23    0.000353           5        68           pipe2
  1.15    0.000330           6        51           poll
...
------ ----------- ----------- --------- --------- ------------------
100.00    0.028695          25      1136        22 total
```

src/network/netlink.rs
```diff
@@ -0,0 +1,407 @@
...
+use log::trace;
+use netlink_packet_route::{
+    nlas::link::{Info, InfoData, InfoKind, Nla},
+    AddressMessage, LinkMessage, NetlinkHeader, NetlinkMessage, NetlinkPayload, RouteMessage,
+    RtnlMessage, AF_INET, AF_INET6, IFF_UP, NLM_F_ACK, NLM_F_CREATE, NLM_F_DUMP, NLM_F_EXCL,
+    NLM_F_REQUEST, RTN_UNICAST, RTPROT_STATIC, RT_SCOPE_UNIVERSE, RT_TABLE_MAIN,
+};
+use netlink_sys::{protocols::NETLINK_ROUTE, SocketAddr};
+
+pub struct Socket {
```
baude, line 19: add  #[derive(Clone )] please

Luap99: Why do you need to clone it? The problem is that you end up having the objects that use the same underlying socket and it will result in problems when you start using it in parallel.

I can definitely add it but it must be carefully used so it would not cause bugs.

Luap99: AFAIK copy is just clone without having to explicitly call .clone() which means that you start cloning without realizing it which is IMO a bad thing

baude: yeah, lets skip this.

baude:
are you contemplating a change from wip to ready soon?

Luap99:
> are you contemplating a change from wip to ready soon?

It should be ready by the end of the day

Luap99:
Tests are green, this is ready from my side. We still need to add the remove addr calls fro the dhcp proxy and better error handling but this can happen in a later PR to make review easier.

PR #434, opened 2022-10-07
Do not use ipv6 autoconf

Luap99:
When we run container netavark (or in the future the dhcp proxy) should set ip addresses. The ipv6 auto configuration is not wanted in almost all cases so we just disable it for now.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2126243

I was able to reproduce this and can confirm that this fixes the issue for me. I don't think we can test it in CI.

### DIFF
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -1248,6 +1248,30 @@ impl CoreUtils {
             }
         };
         tokio::spawn(_connection);
+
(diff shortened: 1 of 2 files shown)


baude:
LGTM, nice work.

Luap99:
pushed a new version to support a read only /proc (just ignore that case) to fix the test failure

PR #436, opened 2022-10-10
aardvark-dns: verify if `init` is actually `systemd` before using `systemd-run`

flouthoc:
On certain system like WSL2 the VM image contains `systemd` but `init` process is not actually `systemd`. 
For such use-cases `netavark` wrongly assumes that `systemd` is present and attempts to start `aardvark-dns` 
with `systemd` therefore actually verify if `init` is `systemd` otherwise spawn `aardvark-dns` without `systemd`.

Closes: https://github.com/containers/podman/discussions/16082

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -6,13 +6,14 @@ use std::fs;
 use std::fs::File;
 use std::fs::OpenOptions;
 use std::io::Result;
-use std::io::{prelude::*, ErrorKind};
+use std::io::{prelude::*, BufReader, ErrorKind};
 use std::net::Ipv4Addr;
 use std::net::{IpAddr, Ipv6Addr};
 use std::path::Path;
 use std::process::{Command, Stdio};
 
 const SYSTEMD_CHECK_PATH: &str = "/run/systemd/system";
+const SYSTEMD_INIT_COMM: &str = "systemd";
 const SYSTEMD_RUN: &str = "systemd-run";
 const AARDVARK_COMMIT_LOCK: &str = "aardvark.lock";
 
@@ -46,6 +47,16 @@ impl Aardvark {
             aardvark_bin,
             port: port.to_string(),
         }
+    }
+
+    // On success returns the init command or returns empty string.
+    fn get_init_comm() -> String {
+        if let Ok(file) = File::open("/proc/1/comm") {
+            if let Some(line) = BufReader::new(file).lines().flatten().next() {
+                return line.trim().to_string();
+            }
+        }
+        "".to_string()
     }
 
     // On success retuns aardvark server's pid or returns -1;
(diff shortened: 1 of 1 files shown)


Luap99 (changes requested):
IMO this check is wrong, see https://github.com/containers/podman/issues/16022

As per sd_booted(3) it should be enough to check for the systemd directory to see if systemd is booted.

What we actually have to do is to use the systemd API and see if it works since the systemd booted check can never work for user sessions.

flouthoc:
> IMO this check is wrong, see [containers/podman#16022](https://github.com/containers/podman/issues/16022)
> 
> As per sd_booted(3) it should be enough to check for the systemd directory to see if systemd is booted.
> 
> What we actually have to do is to use the systemd API and see if it works since the systemd booted check can never work for user sessions.

@Luap99 Is this also true for distro where people have installed package systemd but are not actually it ? I am thinking of a use-case where the systemd as a component exists on the host but it is not actually used.

Luap99:
/run/systemd/system should be on a tmpfs, systemd will be create it on boot. I don't think it will be created by installing packages.

flouthoc:
Patch is already verified here: https://github.com/containers/podman/discussions/16082#discussioncomment-3841188 and there is already check for `/run/systemd/system` is this enough for now ? and add calling systemd API when we see any issues in future ? Although I don't see any issues in invoking `/usr/bin/true` with systemd-run as sanity check.

giuseppe:
I agree with @Luap99, checking that directory should be enough, it is equivalent to checking `/proc/1/comm`.  We should change it in Podman as well.

> Although I don't see any issues in invoking `/usr/bin/true` with systemd-run as sanity check.

IMO it is better to not introduce these expensive checks for some weird use cases

drawks:
> I agree with @Luap99, checking that directory should be enough, it is equivalent to checking `/proc/1/comm`. We should change it in Podman as well.
> 
> > Although I don't see any issues in invoking `/usr/bin/true` with systemd-run as sanity check.
> 
> IMO it is better to not introduce these expensive checks for some weird use cases

Is there no method to specifically tell netavark that you don't want it to use systemd?

Regardless of how `weird` the use case is, we currently have a situation where if you are running modern podman on modern linux distros and you explicitly specify that you don't want to use systemd (for instance by passing `--cgroup-manager=cgroupfs` to podman) that netavark makes no effort to follow the operator's intent and instead uses a relatively arbitrary heuristic to determine if systemd specific code paths should be used.

I think it is probably (and provably) fine to lean on a heuristic driven automatic decision for the common case, but not offering some lever for operators to pull to specify a behavior feels sub-optimal.

I'm commenting on this existing issue/MR to gauge the interest in revisting this topic and make sure I have the correct understanding before I invest further effort in floating my own MR.

TIA

drawks:
As an additional note to my previous comment, I did test shadowing `/run/systemd` with an empty directory to prevent the systemd specific behavior from being called and observed that netavark and aardvark-dns worked as I would have expected without requiring any interaction with a systemd dbus or active systemd user slice/session.

Issue #440, opened 2022-10-10
README.md on the main page lists v1.0.0 as the latest release


TomSweeneyRedHat:
I think I'd just remove it from the README.md, but if you want to keep it, it probably ought to be updated.

flouthoc:
I think github tags are already listed, following section just adds a maintenance overhead and other projects in github.com/containers also don't list their latest version in README.md so I think we should just remove it. https://github.com/containers/netavark/pull/442

PR #442, opened 2022-10-11
README.md: remove latest release section

flouthoc:
Netavark follows github releases which is already listed on homepage. 
* Remove section which needs additional maintenance.
* Follow parity with other projects in github.com/containers such as podman.

Closes: https://github.com/containers/netavark/issues/440

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,6 @@ $ make test
 ```
 Also see [./test](./test/README.md) for more information.
 
-## Latest release
-[v1.0.0](https://github.com/containers/netavark/releases/latest)
-
 ## Communications
 
 For general questions and discussion, please use Podman's


Luap99 (commented):
LGTM

baude:
lgtm

mheon:
This is not a good idea. The Github release page is *strictly chronological*. If we ever have to do a, say, 1.0.2 release, that will appear as the latest release on Github - despite us already having 1.1.0 out. We retain and manually curate the latest release link on Podman because of this. Don't see a reason why NV and AV don't merit one?

Luap99:
Well, the point is nobody seems to update it. So removing it sounds better to me than using the wrong version.

PR #444, opened 2022-10-11
netlink: add basic unit tests

Luap99:
Add some minimal unit tests for the netlink abstraction. While most code is tested via integration tests with bats.

This will be useful when I add the remove_addr/route calls that will only be used in the dhcp proxy so we cannot test it here otherwise.


flouthoc (commented):
This looks really nice,  maybe a new file like https://github.com/containers/aardvark-dns/blob/main/src/test/test.rs ?

src/network/netlink.rs
```diff
@@ -405,3 +405,73 @@ pub fn parse_create_link_options(msg: &mut LinkMessage, options: CreateLinkOptio
...
+            .expect("get_link failed");
+
+        let net = "10.0.0.2/24";
+        sock.add_addr(link.header.index, &net.parse().unwrap())
+            .expect("add_addr failed");
+
+        let out = String::from_utf8(run_command!("ip", "addr", "show", "test1").stdout)
+            .expect("convert to string");
+
+        assert!(out.contains(net), "addr does does not exists");
```
b-m-f, line 475: `does does`

src/network/netlink.rs
```diff
@@ -405,3 +405,73 @@ pub fn parse_create_link_options(msg: &mut LinkMessage, options: CreateLinkOptio
...
+    fn test_add_link() {
+        test_setup!();
+        let mut sock = Socket::new().expect("Socket::new()");
+
+        let name = String::from("test1");
+        sock.create_link(CreateLinkOptions::new(name.clone(), InfoKind::Dummy))
+            .expect("create link");
+
+        let out = String::from_utf8(run_command!("ip", "link", "show", &name).stdout)
+            .expect("convert to string");
```
b-m-f, line 450: How about `converting to string failed`?

src/network/netlink.rs
```diff
@@ -405,3 +405,73 @@ pub fn parse_create_link_options(msg: &mut LinkMessage, options: CreateLinkOptio
...
+    }
+
+    #[test]
+    fn test_add_link() {
+        test_setup!();
+        let mut sock = Socket::new().expect("Socket::new()");
+
+        let name = String::from("test1");
+        sock.create_link(CreateLinkOptions::new(name.clone(), InfoKind::Dummy))
+            .expect("create link");
```
b-m-f, line 447: Same here with adding a failed. Just for consistency with the `test_add_addr` test.

b-m-f:
Hope you dont mind my comments :)

Luap99:
Addressed the comments and moved it to a new file.

Luap99:
@mheon @baude @flouthoc PTAL

baude:
LGTM

PR #445, opened 2022-10-11
bridge: do not set gateway for internal network

### LINKED ISSUE #415: Internal networks should not have a gateway
When creating an internal network, netavark gives it a gateway.
Worse : when connecting a container to this network, the gateway is set as "default gateway".

The default behavior when creating an internal network should be not to assign it a gateway.
At least, we should have the possibility to explicitly not set a gateway.

```console
# podman network create --driver bridge --subnet 10.127.0.0/16 --internal internal
# podman network inspect internal
[
     {
          "name": "internal",
          "id": "c259e2b6cec93c896d8abebd5d98d85790e3e9e306970786268098eeb9eacbce",
          "driver": "bridge",
          "network_interface": "podman2",
          "created": "2022-09-22T15:48:55.780025615+02:00",
          "subnets": [
               {
                    "subnet": "10.127.0.0/16",
                    "gateway": "10.127.0.1"
               }
          ],
          "ipv6_enabled": false,
          "internal": true,
          "dns_enabled": true,
          "ipam_options": {
               "driver": "host-local"
          }
     }
]
# podman run --net internal alpine ip route
default via 10.127.0.1 dev eth0
10.127.0.0/16 dev eth0 scope link  src 10.127.0.2
```

I have podman version 4.1.1 with netavark 2:1.0.1-35

Luap99:
This is a regression. It made its way in made in because the test was not working as expected. The test and code is now fixed. An internal network should never have an default gateway since it cannot reach outside anyway.

Fixes #415

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -120,6 +120,7 @@ impl driver::NetworkDriver for Bridge<'_> {
             host_sock,
             netns_sock,
             data,
+            self.info.network.internal,
             self.info.netns_host,
             self.info.netns_container,
         )?;
@@ -429,6 +430,7 @@ fn create_interfaces(
     host: &mut netlink::Socket,
     netns: &mut netlink::Socket,
     data: &InternalData,
+    internal: bool,
     hostns_fd: RawFd,
     netns_fd: RawFd,
 ) -> NetavarkResult<String> {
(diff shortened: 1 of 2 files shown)


mheon:
LGTM

flouthoc (approved):
LGTM
/lgtm

PR #447, opened 2022-10-12
error: NetavarkError use Into<String> Trait

Luap99:
Instead of using two function for types String and str we can just use a generic which implements Into<String>. This allows us to only use one function for both.


Luap99:
@mheon @flouthoc PTAL

flouthoc (approved):
LGTM

PR #448, opened 2022-10-13
netlink: add del_{addr,route} calls

Luap99:
It is not required for netavark since we just delete links on teardown but the dhcp proxy might need to remove addresses and/or routes when it gets a new lease.

Unfortunately the del_route unit test doe snot seem to work since it requires "real" working routes which we cannot create in a custom netns.


Luap99:
@baude PTAL

baude:
LGTM, triggered the failed test; looks like CI gacked when running ...

PR #451, opened 2022-10-14
[CI:DOCS] GHA: Add Lintje commit validation

cevich:


cevich:
After giving lintje a try, I don't think it will work for us.  Can re-open if otherwise.

PR #452, opened 2022-10-17
netavark ,aardvark: accept and populate custom `dns_servers` for containers

flouthoc:
Netavark now accepts `dns_servers` as part of `NetworkOptions` which contains list of comma seperated custom DNS servers for containers. Aardvark-dns will use these as resolver for a specific container instead of host's default resolver.

Implements feature here and needs: https://github.com/containers/aardvark-dns/pull/240

### DIFF
--- a/src/commands/teardown.rs
+++ b/src/commands/teardown.rs
@@ -91,6 +91,7 @@ impl Teardown {
                 firewall: firewall_driver.as_ref(),
                 container_id: &network_options.container_id,
                 container_name: &network_options.container_name,
+                container_dns_servers: &network_options.dns_servers,
                 netns_host: hostns.fd,
                 netns_container: netns.fd,
                 network,
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -24,6 +24,7 @@ pub struct AardvarkEntry<'a> {
     pub container_ips_v4: Vec<Ipv4Addr>,
     pub container_ips_v6: Vec<Ipv6Addr>,
     pub container_names: Vec<String>,
+    pub container_dns_servers: &'a Option<Vec<IpAddr>>,
 }
 
 #[derive(Debug, Clone)]
@@ -249,9 +250,24 @@ impl Aardvark {
             .collect::<Vec<String>>()
             .join(",");
 
+        let dns_server = if let Some(dns_servers) = &entry.container_dns_servers {
+            if !dns_servers.is_empty() {
+                let dns_server_collected = dns_servers
+                    .iter()
+                    .map(|g| g.to_string())
+                    .collect::<Vec<String>>()
+                    .join(",");
+                format!(" {}", dns_server_collected)
+            } else {
+                "".to_string()
+            }
+        } else {
+            "".to_string()
+        };
+
         let data = format!(
-            "{} {} {} {}\n",
-            entry.container_id, ipv4s, ipv6s, container_names
+            "{} {} {} {}{}\n",
+            entry.container_id, ipv4s, ipv6s, container_names, dns_server
         );
 
         file.write_all(data.as_bytes())?; // return error if write fails
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -181,8 +181,16 @@ impl driver::NetworkDriver for Bridge<'_> {
                 container_ips_v4: ipv4,
                 container_ips_v6: ipv6,
                 container_names: names,
+                container_dns_servers: self.info.container_dns_servers,
             })
         } else {
+            // If --dns-enable=false and --dns was set then return following DNS servers
+            // in status_block so podman can use these and populate resolv.conf
+            if let Some(container_dns_servers) = self.info.container_dns_servers {
+                let _ = response
+                    .dns_server_ips
+                    .insert(container_dns_servers.clone());
+            }
             None
         };
 
--- a/test/100-bridge-iptables.bats
+++ b/test/100-bridge-iptables.bats
@@ -148,6 +148,39 @@ fw_driver=iptables
     run_netavark --file ${TESTSDIR}/testfiles/ipv6-bridge.json teardown $(get_container_netns_path)
 }
 
+@test "$fw_driver - bridge driver must generate config for aardvark with custom dns server" {
+    # TODO !!! Unkip after https://github.com/containers/aardvark-dns/pull/240
+    skip "unskip after https://github.com/containers/aardvark-dns/pull/240"
+    # get a random port directly to avoid low ports e.g. 53 would not create iptables
+    dns_port=$((RANDOM+10000))
+
+    # hack to make aardvark-dns run when really root or when running as user with
+    # podman unshare --rootless-netns; since netavark runs aardvark with systemd-run
+    # it needs to know if it should use systemd user instance or not.
+    # iptables are still setup identically.
+    rootless=false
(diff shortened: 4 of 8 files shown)


src/dns/aardvark.rs
```diff
@@ -249,11 +250,25 @@ impl Aardvark {
...
+                    .map(|g| g.to_string())
+                    .collect::<Vec<String>>()
+                    .join(",");
+                data = format!(
+                    "{} {} {} {} {}\n",
+                    entry.container_id, ipv4s, ipv6s, container_names, dns_server_collected
+                );
+            }
+        }
+
```
Luap99, line 271: why use almost the same format() string twice? just create the dns server string before and add a space at the beginning if it is not empty

src/commands/teardown.rs
```diff
@@ -91,6 +91,7 @@ impl Teardown {
                 firewall: firewall_driver.as_ref(),
                 container_id: &network_options.container_id,
                 container_name: &network_options.container_name,
+                container_dns_servers: &None,
```
Luap99, line 94: Why is this none? I understand that it is not required for teardown but it is inconsistent, just pass the nameservers along

test/100-bridge-iptables.bats
```diff
@@ -148,6 +148,37 @@ fw_driver=iptables
     run_netavark --file ${TESTSDIR}/testfiles/ipv6-bridge.json teardown $(get_container_netns_path)
 }
 
+@test "$fw_driver - bridge driver must generate config for aardvark with custom dns server" {
+    # get a random port directly to avoid low ports e.g. 53 would not create iptables
```
Luap99, line 152: just skip the test for now, then we can merge. get aardvark-dns tests working and once aardvark-dns is merged we can enable this test.

flouthoc: It looks to me that `aardvark`'s integration test is using netavark

Luap99 (commented):
Also this should return the passed in the servers in the response when dns is disabled so podman does not need to check for anything and can just write the result to resolv.conf.

flouthoc:
> Also this should return the passed in the servers in the response when dns is disabled so podman does not need to check for anything and can just write the result to resolv.conf.

I think what you want is already handled here, podman is already aware before spawning netavark if network has `dns_enabled` or not ( https://github.com/containers/podman/pull/16175 )

src/dns/aardvark.rs
```diff
@@ -249,9 +250,21 @@ impl Aardvark {
             .collect::<Vec<String>>()
             .join(",");
 
+        let mut dns_server: String = "".to_string();
```
Luap99, line 253: This is not required to be mutable, do it like this:
```
let y = if 12 * 15 > 150 {
    "Bigger"
} else {
    "Smaller"
};
```
From https://doc.rust-lang.org/reference/expressions/if-expr.html

src/network/bridge.rs
```diff
@@ -181,6 +181,7 @@ impl driver::NetworkDriver for Bridge<'_> {
                 container_ips_v4: ipv4,
                 container_ips_v6: ipv6,
                 container_names: names,
+                container_dns_servers: self.info.container_dns_servers.clone(),
```
Luap99, line 184: Can we keep the reference here and not clone?

Luap99:
> > Also this should return the passed in the servers in the response when dns is disabled so podman does not need to check for anything and can just write the result to resolv.conf.
> 
> I think what you want is already handled here, podman is already aware before spawning netavark if network has `dns_enabled` or not ( [containers/podman#16175](https://github.com/containers/podman/pull/16175) )

Exactly podman should never check for dns enabled, netavark should provide the exact servers that end up in resolv.conf. The podman check is much more costly since it requires a network inspect each time with locking,etc...

test/100-bridge-iptables.bats
flouthoc: ah i see what you mean, let me skip this for now.

src/dns/aardvark.rs
flouthoc: I'm not sure if I got you here, but below use-case has two nested expression so either we make `dns_server_collected` mutable or `dns_server` or otherway could to be return a result but it increases number of lines and makes it even more un-readable.

But I must be missing what you exactly have in mind so I'm cool if there any simple expression here.

test/100-bridge-iptables.bats
flouthoc: **Note!!!**: following test must be unskipped in another PR after https://github.com/containers/aardvark-dns/pull/240

src/dns/aardvark.rs
Luap99: just make it another if else block, or even better just remove the isempty check since it is not needed, when it is empty it will be None anyway.

flouthoc:
> > > Also this should return the passed in the servers in the response when dns is disabled so podman does not need to check for anything and can just write the result to resolv.conf.
> > 
> > 
> > I think what you want is already handled here, podman is already aware before spawning netavark if network has `dns_enabled` or not ( [containers/podman#16175](https://github.com/containers/podman/pull/16175) )
> 
> Exactly podman should never check for dns enabled, netavark should provide the exact servers that end up in resolv.conf. The podman check is much more costly since it requires a network inspect each time with locking,etc...

This is added to status_block in latest commit.

src/dns/aardvark.rs
flouthoc: Done.

flouthoc:
@Luap99 @baude @mheon Could you PTAL.

src/dns/aardvark.rs
```diff
@@ -249,9 +250,24 @@ impl Aardvark {
...
+                "".to_string()
+            }
+        } else {
+            "".to_string()
+        };
+
         let data = format!(
-            "{} {} {} {}\n",
-            entry.container_id, ipv4s, ipv6s, container_names
+            "{} {} {} {}{}\n",
```
mheon, line 269: Where are we prepending a space so that this is separated from the previous field?

flouthoc: @mheon Here https://github.com/containers/netavark/pull/452/files#diff-6307ba066c1817023f2a53b2c7c6a64d35bde6cf7f443cf5b07f34b9f163513bR260 `format!(" {}", dns_server_collected)`

test/100-bridge-iptables.bats
```diff
@@ -148,6 +148,39 @@ fw_driver=iptables
...
+    mkdir -p "$NETAVARK_TMPDIR/config"
+
+    NETAVARK_DNS_PORT="$dns_port" run_netavark --file ${TESTSDIR}/testfiles/dualstack-bridge-custom-dns-server.json \
+        --rootless "$rootless" --config "$NETAVARK_TMPDIR/config" \
+        setup $(get_container_netns_path)
+
+    # check aardvark config and running
+    run_helper cat "$NETAVARK_TMPDIR/config/aardvark-dns/podman1"
+    assert "${lines[0]}" =~ "10.89.3.1,fd10:88:a::1" "aardvark set to listen to all IPs"
+    assert "${lines[1]}" =~ "^[0-9a-f]{64} 10.89.3.2 fd10:88:a::2 somename 8.8.8.8$" "aardvark config's container"
```
Luap99, line 175: We should also test at least two custom servers but since we cannot test it right now I fine with keeping like this and add a second test for this once aardvark is ready

Luap99 (commented):
LGTM

PR #455, opened 2022-10-19
NetworkOptions::load(): do not open /dev/stdin

Luap99:
There is no need to open /dev/stdin, we should just read directly from fd 0 without having to call open().

Also convert the error inside the function to prevent duplication.

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -30,7 +30,7 @@ impl Setup {
 
     pub fn exec(
         &self,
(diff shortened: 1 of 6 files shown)


flouthoc:
Tests are failing with this PR

mheon:
LGTM

Luap99:
@baude @flouthoc PTAL

flouthoc (approved):
LGTM
/lgtm
/approve

PR #456, opened 2022-10-19
use cargo-vendor-filterer to create vendor tarball

### LINKED ISSUE #347: Use `cargo vendor-filterer`
We're using this in the coreos/ org, see https://github.com/coreos/fedora-coreos-tracker/issues/1240

Luap99:
This tool creates a much small vendor dir since we can only include the linux dependencies. Our current one on main is 46 MB big and with cargo vendor-filterer it is only 6 MB.

see https://github.com/coreos/cargo-vendor-filterer

Fixes #347


flouthoc (approved):
LGTM

PR #463, opened 2022-10-24
netlink: better errors on invalid querys

Luap99:
* netlink: improve error message on invalid netlink result
While unlikely when netlink returns something unexpected we should
provide more details so we know where it failed. This will most likely
only fail when we called something incorrectly or due a kernel bug.
If it run with trace log level we already display the raw netlink
messages so that should be enough info to start debugging.

* netlink: check sequence_number
Make sure we only parse messages that are for us. If the sequence_number
does not match it likely means we screwed up somewhere. Better to catch
this early than having the callers get unexpected responses which causes
hard to debug issues.

### DIFF
--- a/src/network/netlink.rs
+++ b/src/network/netlink.rs
@@ -360,6 +360,13 @@ impl Socket {
                     })?;
                 trace!("read netlink packet: {:?}", rx_packet);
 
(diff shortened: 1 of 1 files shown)


baude:
code lgtm

flouthoc (approved):
LGTM
/lgtm
/approve

PR #467, opened 2022-10-24
Ignore dns-enabled for macvlan

baude:
it makes no sense to have dns-enabled for macvlan environments regardless of what the JSON configuration file says.

Fixes bz#2137320


src/dns/aardvark.rs
```diff
@@ -313,7 +315,7 @@ impl Aardvark {
         let mut modified = false;
         let container_id = network_options.container_id;
         for (key, network) in network_options.network_info {
-            if network.dns_enabled {
+            if network.dns_enabled && network.driver != DRIVER_MACVLAN {
```
Luap99, line 318: I thought about this and lets make it == bridge instead, if we add more drivers this will bite us again, better to explicitly opt in.

test/testfiles/macvlan.json
```diff
@@ -23,7 +23,7 @@
           ],
           "ipv6_enabled": false,
           "internal": false,
-          "dns_enabled": false,
+          "dns_enabled": true,
```
Luap99, line 26: the test still pass, you have to actually call teardown in the test:
```diff
diff --git a/test/300-macvlan.bats b/test/300-macvlan.bats
index f3ed201..2b68afb 100644
--- a/test/300-macvlan.bats
+++ b/test/300-macvlan.bats
@@ -36,6 +36,9 @@ function setup() {
 
     run_in_container_netns cat /proc/sys/net/ipv6/conf/eth0/autoconf
     assert "0" "autoconf is disabled"
+
+    run_netavark --file ${TESTSDIR}/testfiles/macvlan.json teardown $(get_container_netns_path)
+    asset "" "no errors"
 }
 
 @test "macvlan setup internal" {
```

test/300-macvlan.bats
```diff
@@ -36,6 +36,9 @@ function setup() {
 
     run_in_container_netns cat /proc/sys/net/ipv6/conf/eth0/autoconf
     assert "0" "autoconf is disabled"
+
+    run_netavark --file ${TESTSDIR}/testfiles/macvlan.json teardown $(get_container_netns_path)
+    asset "" "no errors"
```
Luap99, line 41: mb, that must be `assert`

src/dns/aardvark.rs
```diff
@@ -1,4 +1,6 @@
+use crate::network::constants::{DRIVER_BRIDGE, DRIVER_MACVLAN};
```
Luap99, line 1: I assume lint will fail since macvlan is unused

Luap99 (commented):
/lgtm

PR #470, opened 2022-10-26
macvlan: fix name collision on hostns

Luap99:
When the container interface and host interface use the same link netavark fails with EEXISTS. This is a weird kernel behavior because we directly create the interface in the netns but the kernel still checks in the host ns and thus it will fail if the same name is already used there.

To fix this we need to create the interface with a tmp name then rename it in the netns. For performance reason we still try the quick way first and only fallback to tmp name if required.

### DIFF
--- a/src/network/macvlan.rs
+++ b/src/network/macvlan.rs
@@ -2,6 +2,7 @@ use std::{collections::HashMap, net::IpAddr, os::unix::prelude::RawFd};
 
 use log::debug;
 use netlink_packet_route::nlas::link::{InfoData, InfoKind, InfoMacVlan, Nla};
+use rand::distributions::{Alphanumeric, DistString};
 
 use crate::{
     dns::aardvark::AardvarkEntry,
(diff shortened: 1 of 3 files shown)


src/network/macvlan.rs
```diff
@@ -165,7 +166,48 @@ fn setup(
     opts.info_data = Some(InfoData::MacVlan(vec![InfoMacVlan::Mode(
         data.macvlan_mode,
     )]));
-    host.create_link(opts).wrap("create macvlan interface")?;
+    let mut result = host.create_link(opts.clone());
+    // Sigh, the kernel creates the interface first in the hostns before moving it into the netns.
+    // Therefore it can fail with EEXIST if the name is already used on the host. Create the link
+    // with tmp name, then rename it in the netns.
+    // If you change the iterations here make sure to match the number in early return case below as well.
+    for i in 0..3 {
```
flouthoc, line 174: Why is this iteration needed ?

Luap99: there is no guarantee that the random name is free, I think trying three times is enough.

flouthoc: but then why below has a check to return error when iterator is on second counter, let me mark that line

src/network/macvlan.rs
```diff
@@ -165,7 +166,48 @@ fn setup(
...
+            Err(err) => match err {
+                NetavarkError::Netlink(ref e) if -e.code == libc::EEXIST => {
+                    let random = Alphanumeric.sample_string(&mut rand::thread_rng(), 10);
+                    let tmp_name = "mv-".to_string() + &random;
+                    let mut opts = opts.clone();
+                    opts.name = tmp_name.clone();
+                    result = host.create_link(opts);
+                    if let Err(ref e) = result {
+                        // if last element return directly
+                        if i == 2 {
```
flouthoc, line 188: @Luap99 Just marking  for above thread.

src/network/macvlan.rs
flouthoc: This one https://github.com/containers/netavark/pull/470#discussion_r1006009190

flouthoc: ah nvm I think you want to throw error on the last try.

Luap99: no, for i in 0..3 means 0,1,2. like any classic for loop, for(int = 0; i<3;i++){}
therefore we need to return the error when we are in the last iteration.

flouthoc (approved):
LGTM

Luap99:
@baude @mheon PTAL
This is fixing a regression from https://github.com/containers/netavark/pull/429 (not yet released).

PR #474, opened 2022-11-01
Add support for route metrics

mheon:
Metrics allow us to specify which default route will be preferred when a container connects to multiple networks. By default, all routes will get a metric of 100. If two routes go to the same location (e.g. multiple default routes, one for each network, as Netavark does by default), the kernel by default round-robins traffic between all of them, but if the metrics for the routes differ, the route(s) with the lowest metric are chosen.

For confusing reasons, metric is actually called preference in Netlink - and, doubly confusing, there is also a "metrics" option for routes in Netlink as well. Which, naturally, is completely unrelated to route metric. Sigh.

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -16,7 +16,7 @@ use crate::{
 };
 
 use super::{
-    constants::{NO_CONTAINER_INTERFACE_ERROR, OPTION_ISOLATE, OPTION_MTU},
+    constants::{NO_CONTAINER_INTERFACE_ERROR, OPTION_ISOLATE, OPTION_METRIC, OPTION_MTU},
     core_utils::{self, get_ipam_addresses, join_netns, parse_option, CoreUtils},
     driver::{self, DriverInfo},
     internal_types::{
@@ -41,6 +41,8 @@ struct InternalData {
     mtu: u32,
     /// if this network should be isolated from others
     isolate: bool,
+    /// Route metric for any default routes added for the network
+    metric: Option<u32>,
     // TODO: add vlan
 }
 
@@ -69,6 +71,7 @@ impl driver::NetworkDriver for Bridge<'_> {
 
         let mtu: u32 = parse_option(&self.info.network.options, OPTION_MTU, 0)?;
         let isolate: bool = parse_option(&self.info.network.options, OPTION_ISOLATE, false)?;
+        let metric: u32 = parse_option(&self.info.network.options, OPTION_METRIC, 100)?;
 
         let static_mac = match &self.info.per_network_opts.static_mac {
             Some(mac) => Some(CoreUtils::decode_address_from_hex(mac)?),
@@ -82,6 +85,7 @@ impl driver::NetworkDriver for Bridge<'_> {
             ipam,
             mtu,
             isolate,
+            metric: Some(metric),
         });
         Ok(())
     }
@@ -620,7 +624,7 @@ fn create_veth_pair(
         .wrap("set container veth up")?;
 
     if !internal {
-        core_utils::add_default_routes(netns, &data.ipam.gateway_addresses)?;
+        core_utils::add_default_routes(netns, &data.ipam.gateway_addresses, data.metric)?;
     }
 
     Ok(mac)
--- a/src/network/netlink.rs
+++ b/src/network/netlink.rs
@@ -5,9 +5,10 @@ use std::{
 
 use crate::{
     error::{ErrorWrap, NetavarkError, NetavarkResult},
+    network::constants,
     wrap,
 };
-use log::trace;
+use log::{info, trace};
 use netlink_packet_route::{
     nlas::link::{Info, InfoData, InfoKind, Nla},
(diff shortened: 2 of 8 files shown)


mheon:
This is ready

src/network/netlink.rs
```diff
@@ -212,16 +212,22 @@ impl Socket {
         Ok(())
     }
 
-    fn create_route_msg(route: &Route) -> RouteMessage {
+    fn create_route_msg(route: &Route, metric: Option<u32>) -> RouteMessage {
```
Luap99, line 215: should we make this part of the route type instead of adding new parameters?
this will break @baude's work in the dhcp proxy, since he also reuses the route and addr calls there

src/network/netlink.rs
```diff
@@ -212,16 +212,22 @@ impl Socket {
...
         let mut msg = RouteMessage::default();
 
+        let final_metric = metric.unwrap_or(100);
```
Luap99, line 218: Can please make a comment why 100 is the default vs not setting it in the netlink call and let the kernel choose?
I assume this is because the kernel uses 0 as default which makes it highest prio?

src/network/netlink.rs
```diff
@@ -212,16 +212,22 @@ impl Socket {
...
         msg.header.protocol = RTPROT_STATIC;
         msg.header.scope = RT_SCOPE_UNIVERSE;
         msg.header.kind = RTN_UNICAST;
 
         let (dest_vec, dest_prefix, gateway_vec) = match route {
             Route::Ipv4 { dest, gw } => {
+                info!(
+                    "Adding route to {} via {} with metric {}",
+                    dest, gw, final_metric
+                );
```
Luap99, line 230: This is inconsistent with the other calls which do not log. In principle I am ok with this but we should be consistent across the code base. Currently logs are all over the place and it is not really clear what should be trace, debug and info level? 

Also you could just `"{}", route` since the Display trait is implemented which should allow you do not duplicate the log line twice.

test/100-bridge-iptables.bats
```diff
@@ -666,3 +666,21 @@ EOF
...
     assert '{"error":"add ip addr to bridge: failed to add ipv6 address, is ipv6 enabled in the kernel?: Netlink error: Permission denied (os error 13)"}' "error message"
 }
+
+@test "$fw_driver - route metric from config" {
+    run_netavark --file ${TESTSDIR}/testfiles/metric.json setup $(get_container_netns_path)
+
+    run_in_container_netns ip -j route list match 0.0.0.0
+    default_route="$output"
+    assert_json "$default_route" '.[0].dst' == "default" "Default route was selected"
+    assert_json "$default_route" '.[0].metric' == "200" "Route metric set from config"
```
Luap99, line 676: missing ipv6 tests

test/100-bridge-iptables.bats
```diff
@@ -666,3 +666,21 @@ EOF
...
+}
+
+@test "$fw_drive - default route metric" {
+    run_netavark --file ${TESTSDIR}/testfiles/dualstack-bridge.json setup $(get_container_netns_path)
+
+    run_in_container_netns ip -j route list match 0.0.0.0
+    default_route="$output"
+    assert_json "$default_route" '.[0].dst' == "default" "Default route was selected"
+    assert_json "$default_route" '.[0].metric' == "100" "Default metric was chosen"
```
Luap99, line 685: Should we add this to an existing test? I don't think we need an extra test for it but I am also fine with this.

src/network/netlink.rs
mheon: Sure, I can change this

mheon:
All comments addressed, tests should be passing

test/100-bridge-iptables.bats
mheon: I think having a separate test is useful, and we don't exactly have a problem with test runtimes getting excessive here

Luap99:
I still think macvlan should be supported as well, https://github.com/containers/netavark/pull/474#discussion_r1011681058

Luap99:
LGTM

mheon:
@baude @flouthoc PTAL

src/network/bridge.rs
```diff
@@ -82,6 +85,7 @@ impl driver::NetworkDriver for Bridge<'_> {
             ipam,
             mtu,
             isolate,
+            metric: Some(metric),
```
flouthoc, line 88: I think this is always going to be some value or `100`, since `parse_option` has a default of `100`. I think if not parse-able we should pass `None` here instead of default `100`.

mheon: Why? Non-parsable should be an error.

flouthoc: No my point was that metric defaults to 100 in parse_option, we should not default to any value instead default should be `None`.

mheon: That would require significant changes to the option parsing code that I would prefer to avoid for the moment. I'm willing to do it if you insist, though

flouthoc: No not necessary and blocker for this PR, we can do it in a new PR. I'll open a PR for that after this gets merged.

flouthoc (approved):
LGTM

PR #476, opened 2022-11-02
macvlan: remove tmp interface when name already used in netns

Luap99:
If the interface name is already used in the netns the rename call will fail and netavark error out. This is correct but before it returns it should remove the tmp interface otherwise we end up leaking it.


Luap99:
@baude PTAL

mheon:
Code LGTM

Luap99:
I rerun aarch integration test, somehow netavark was not available in the VM???
https://cirrus-ci.com/task/4596649891725312

flouthoc (approved):
LGTM

PR #477, opened 2022-11-07
macvlan modes: use upstream consts

Luap99:
Fixes a small TODO in the code. Also improve the error message including "io error" make no sense. I also added the actual mode name in the error.


mheon:
LGTM

PR #480, opened 2022-11-09
run cargo update

Luap99:
Most notably update cpufeatures to fix a panic on x86_64 cpus which do not have (full?) avx support, see https://github.com/containers/podman/issues/16194.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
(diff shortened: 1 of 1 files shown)


mheon:
`error: package `zbus_names v2.2.0` cannot be built because it requires rustc 1.60 or newer, while the currently active rustc version is 1.59.0`

Old Ubuntu version with old Rust?

Luap99:
> Old Ubuntu version with old Rust?

Nope that is their current rust version even in newer ubuntu releases. I revert the stuff we that it will break.

Luap99:
Tests are green

mheon:
LGTM

PR #493, opened 2022-11-17
fix rust doc comments

Luap99:
Rust doc comments need three slashes[1]. Using just `//` makes rust think they are a normal comment which should not be included in the API docs. This is important if you use function highlighting in my IDE for example.

[1] https://doc.rust-lang.org/reference/comments.html

### DIFF
--- a/src/error/mod.rs
+++ b/src/error/mod.rs
@@ -57,7 +57,7 @@ impl<T> ErrorWrap<T> for NetavarkResult<T> {
     }
 }
 
-// The main Netavark error type
+/// The main Netavark error type
 #[derive(Debug)]
 pub enum NetavarkError {
     // A string message
@@ -81,7 +81,7 @@ pub enum NetavarkError {
     List(NetavarkErrorList),
 }
 
-// Internal struct for JSON output
+/// Internal struct for JSON output
 #[derive(Debug, Serialize, Deserialize)]
 struct JsonError {
     pub error: String,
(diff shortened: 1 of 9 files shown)


Luap99:
@mheon @flouthoc @baude PTAL. I know I make this mistake as well but try to remember that you have to use `///` to comment functions or types in rust  unless you really do not want them to appear in the docs.

flouthoc (approved):
LGTM

src/error/mod.rs
```diff
@@ -81,7 +81,7 @@ pub enum NetavarkError {
     List(NetavarkErrorList),
 }
 
-// Internal struct for JSON output
+/// Internal struct for JSON output
```
mheon, line 84: Do we need this? It's not `pub`?

mheon:
General question: should we bother with rustdoc on non-public functions? It seems like we should focus on rustdoc for things that could actually be used outside of the package.

Luap99:
> General question: should we bother with rustdoc on non-public functions? It seems like we should focus on rustdoc for things that could actually be used outside of the package.

Well it should still show up as hint when I hover the function name in my IDE. I assume that if you render the docs as web page it will omit non public types/functions,

flouthoc:
> General question: should we bother with rustdoc on non-public functions? It seems like we should focus on rustdoc for things that could actually be used outside of the package.

@mheon Lot of editior extensions use those to show helpers so I guess we can keep it as it is.

PR #496, opened 2022-11-18
remove number of .clone() calls

Luap99:
In many places we really do not need to clone the values and can just use references with lifetimes.
I also removed the mut declaration in the aardvark interface which is not needed and thus just causes confusion.

There is a lot more I could do but these ones were very simple to fix and I don't want to spend more time on this right now.


baude:
yah! LGTM

PR #508, opened 2022-11-29
chrono: use from_timestamp_opt

baude:
chrono has deprecated from_timestamp in favor of from_timestamp_opt. this PR allows for new versions of chrono to be used.

Fixes #484

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -9,7 +9,10 @@ fn main() {
     // get timestamp
     let now = match env::var("SOURCE_DATE_EPOCH") {
         Ok(val) => {
-            let naive = NaiveDateTime::from_timestamp(val.parse::<i64>().unwrap(), 0);
+            let naive = match NaiveDateTime::from_timestamp_opt(val.parse::<i64>().unwrap(), 0) {
+                Some(n) => n,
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -9,7 +9,10 @@ fn main() {
     // get timestamp
     let now = match env::var("SOURCE_DATE_EPOCH") {
         Ok(val) => {
-            let naive = NaiveDateTime::from_timestamp(val.parse::<i64>().unwrap(), 0);
+            let naive = match NaiveDateTime::from_timestamp_opt(val.parse::<i64>().unwrap(), 0) {
```
mheon, line 12: Is there a reason we're doing unwrap()? `unwrap_or()` would be safer?

baude: just the way it was ... happy to do otherwise if folks like

Luap99: this is just build.rs it should fail here when it tries to compile but got in invalid SOURCE_DATE_EPOCH

Luap99 (commented):
LGTM

build.rs
mheon: Fair enough

PR #510, opened 2022-12-01
remove verify_vendor CI task

Luap99:
We removed vendoring a long time ago, therefore we do not need this check. I just saw it fail on one of my PRs because of a CI problem.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -149,31 +149,6 @@ validate_aarch64_task:
   setup_script: *setup
   main_script: *main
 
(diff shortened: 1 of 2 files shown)


Luap99:
@cevich @baude @mheon PTAL

cevich:
@mheon and @Luap99 I think we also run this on the release-branches (I think there's only one or two) where perhaps it's also going to get in the way?  Please give a quick think and consider if it should be clobbered there as well.

PR #514, opened 2022-12-06
release,make: add rule `make crate-publish` for publishing crates.io

flouthoc:
* Add new rule `make crate-publish` to automate publish at https://crates.io/crates/netavark
* Document steps for publishing to `crates.io`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,9 @@ LIBEXECDIR ?= ${PREFIX}/libexec
 LIBEXECPODMAN ?= ${LIBEXECDIR}/podman
 
 SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)
+# Get crate version by parsing the line that starts with version.
+CRATE_VERSION ?= $(shell grep ^version Cargo.toml | awk '{print $$3}')
+GIT_TAG ?= $(shell git describe --tags)
 
 # Set this to any non-empty string to enable unoptimized
 # build w/ debugging features.
@@ -46,6 +49,16 @@ $(CARGO_TARGET_DIR):
 build: bin $(CARGO_TARGET_DIR)
 	cargo build $(release)
 	cp $(CARGO_TARGET_DIR)/$(profile)/netavark bin/netavark$(if $(debug),.debug,)
+
+.PHONY: crate-publish
+crate-publish:
+	@if [ "$(CRATE_VERSION)" != "$(GIT_TAG)" ]; then\
+		echo "Git tag is not equivalent to the version set in Cargo.toml. Please checkout the correct tag";\
+		exit 1;\
+	fi
+	@echo "It is expected that you have already done 'cargo login' before running this command. If not command may fail later"
+	cargo publish --dry-run
+	cargo publish
 
 .PHONY: clean
 clean:
--- a/docs/publish-crate.md
+++ b/docs/publish-crate.md
@@ -1,0 +1,7 @@
+# Publishing netavark crate to crates.io
+### Steps
+* Make sure you have already done `cargo login` on your current session with a valid token.
+* `cd netavark`
+* Git checkout the version which you want to publish.
+* `make crate-publish`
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -47,6 +49,24 @@ build: bin $(CARGO_TARGET_DIR)
 	cargo build $(release)
 	cp $(CARGO_TARGET_DIR)/$(profile)/netavark bin/netavark$(if $(debug),.debug,)
 
+.PHONY: crate-publish
+crate-publish:
+	@if [ "$(publish-version)" = "" ]; then\
+		echo "publish-version not set. Usage example 'make crate-publish publish-version=v0.1.2'";\
+		echo "Make sure repo already contains tag for the version";\
+		exit 1;\
+	fi
```
Luap99, line 58: I would just call the variable `version`

Makefile
```diff
@@ -47,6 +49,24 @@ build: bin $(CARGO_TARGET_DIR)
...
+	@if [ "$(REMOTE_ORIGIN_URL)" != "https://github.com/containers/netavark" ]; then\
+		echo "You are working with a fork please use actual repo while publishing. Please clone https://github.com/containers/netavark";\
+		echo "Make sure repo already contains tag for $(publish-version)";\
+		exit 1;\
+	fi
```
Luap99, line 63: This looks wrong. Why am I forced to clone upstream as origin? My fork already has all the correct tags in it.

Makefile
```diff
@@ -47,6 +49,24 @@ build: bin $(CARGO_TARGET_DIR)
...
+	git checkout tags/$(publish-version)
+	git branch
+	@echo "It is expected that you have already done 'cargo login' before running this command. If not command may fail later"
+	cargo publish --dry-run
+	cargo publish
+
```
Luap99, line 69: you never switch back to the previous git state

Luap99 (commented):
I don't really see why we should automatically switch git branches. I would expect the author to know that they have to be on the correct commit.
You already need to do a `make vendor` and `git tag` on this commit anyway so I don't really see a reason why this should include the git logic at all.

Makefile
flouthoc: `cargo` does treat publishing from another repo as a clean publish, you will have to force `--allow-dirty` which will still work but its not the correct way.

Makefile
flouthoc: done.

flouthoc:
@Luap99 Cargo treats publishing from any other repo as a `dirty` release it expects that every release has to be done from the original repo and from the `tag` anything else is considered as a `dirty` release, `dirty` release is still allowed `--allow-dirty` but it is not the right practice which we should follow.

The whole git logic is to prevent dirty release overall.

Makefile
flouthoc: @Luap99 let me try again if `cargo` warns when publishing from a fork or not.

flouthoc: removed the check i think fork is also allowed

Luap99:
> @Luap99 Cargo treats publishing from any other repo as a `dirty` release it expects that every release has to be done from the original repo and from the `tag` anything else is considered as a `dirty` release, `dirty` release is still allowed using forced `--allow-dirty` but it is not the right practice which we should follow.
> 
> The whole git logic is to prevent dirty release overall.

But I have to original repo locally. Git is distributed all the changes are in my local repo unless I decided to only pull certain changes, all you have is different remotes configured but I don't see why cargo would care about that.

The cargo docs speak about a dirty repo, it git terms that mean uncommitted files. `git diff --quiet`. That should simply not be the case when you make a release you really do not need to pull the whole repo again for that.

I would also not include the git witch logic in the makefile, the author should normally already be on the correct commit.

If you want to make sure we only publish from correct git tags you could check `git describe --tags` matches the version Cargo.toml.

flouthoc:
@Luap99 @baude @mheon PTAL

Luap99 (commented):
LGTM

flouthoc:
@baude PTAL

PR #518, opened 2022-12-07
Kill the chicken and the egg

baude:
Reliable CI cannot depend on another task always working when there is a chicken and egg dependency between them.,  Just build the aardvark binary.

### DIFF
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -21,17 +21,9 @@ msg "************************************************************"
 
 show_env_vars
 
-req_env_vars AARDVARK_DNS_URL
-
-showrun curl --fail --location -o /tmp/aardvark-dns.zip "$AARDVARK_DNS_URL"
 mkdir -p /usr/libexec/podman
-cd /usr/libexec/podman
-rm -f aardvark-dns*
-showrun unzip -o /tmp/aardvark-dns.zip
-if [[ $(uname -m) != "x86_64" ]]; then
-    showrun mv aardvark-dns.$(uname -m)-unknown-linux-gnu aardvark-dns
-fi
-showrun chmod a+x /usr/libexec/podman/aardvark-dns
+cargo install --root /usr/libexec/podman --git https://github.com/containers/aardvark-dns
+mv /usr/libexec/podman/bin/aardvark-dns /usr/libexec/podman
 
 # Warning, this isn't the end.  An exit-handler is installed to finalize
 # setup of env. vars.  This is required for runner.sh to operate properly.


Luap99:
I would love to get @cevich involved first. The problem with that approach is that we actually have to compile the full binary and as we know rust compiles very slow which makes CI  longer than necessary

baude:
this fixes the problem right now ... we merge this and "other" options can be thought on later ... when we are not trying to get things merged and releases cut.  if this passes, merge it.

cevich:
Please don't merge this.

cevich (dismissed):
The root cause is: using [skip-ci] on an aardvark or netavark PR.  I've already contacted Cirrus-support about this two hours ago re: skip-ci merges should not be reflected in artifact download permalinks.

The issue can easily be worked around simply by merging some docs-change.  All that's needed is a build on main that produces artifacts.  Merging this PR will have unintended consequences for all netavark and aardvark CI.

Luap99:
> The issue can easily be worked around simply by merging some docs-change. All that's needed is a build on main that produces artifacts. Merging this PR will have unintended consequences for all netavark and aardvark CI.

No it cannot, we already did this, all builds on main fails as well thus it will never produce a new working tarball!!!
https://cirrus-ci.com/build/5656397707214848

contrib/cirrus/setup.sh
```diff
@@ -21,17 +21,9 @@ msg "************************************************************"
...
 mkdir -p /usr/libexec/podman
-cd /usr/libexec/podman
-rm -f aardvark-dns*
-showrun unzip -o /tmp/aardvark-dns.zip
-if [[ $(uname -m) != "x86_64" ]]; then
-    showrun mv aardvark-dns.$(uname -m)-unknown-linux-gnu aardvark-dns
-fi
-showrun chmod a+x /usr/libexec/podman/aardvark-dns
+cargo install --root /usr/libexec/podman --git https://github.com/containers/aardvark-dns
+mv /usr/libexec/podman/bin/aardvark-dns /usr/libexec/podman
```
cevich, line 26:
```suggestion
```suggestion
req_env_vars AARDVARK_DNS_URL

if showrun curl --fail --location -o /tmp/aardvark-dns.zip "$AARDVARK_DNS_URL"; then
    cd /usr/libexec/podman
    rm -f aardvark-dns*
    showrun unzip -o /tmp/aardvark-dns.zip
    if [[ $(uname -m) != "x86_64" ]]; then
        showrun mv aardvark-dns.$(uname -m)-unknown-linux-gnu aardvark-dns
    fi
    showrun chmod a+x /usr/libexec/podman/aardvark-dns
else
    cargo install --root /usr/libexec/podman --git https://github.com/containers/aardvark-dns
    mv /usr/libexec/podman/bin/aardvark-dns /usr/libexec/podman
fi
```
```

cevich (commented):
How about this as a more permanent escape-hatch.  Try to download the artifact, if that fails do it the cargo way (I'm assuming that works for both x86_64 and arm64).

contrib/cirrus/setup.sh
cevich: (confirmed the cargo-install does work for arm64)

cevich:
My intention behind the `if...else...fi` suggestion is that we avoid doing a rust compile at test-time, wasting valuable developer brain-cells.  However, another potentially simpler solution just occurred to me:  Couldn't we just install netavark/aardvark RPMs from @lsm5 podman-next COPR?

cevich:
nevermind...won't work on (new) release-branches.

cevich:
-sigh- let me spare poor @baude brain-cells and just work on my own PRs for this, using the cargo-fallback idea...

PR #520, opened 2022-12-07
Cirrus: Add backup to aardvark-binary artifact DL

cevich:
Fixes: https://github.com/containers/netavark/pull/518

See also: https://github.com/containers/aardvark-dns/pull/269

Around the time of this commit, a bug was discovered (the hard way) in Cirrus-CI:  If a `[skip-ci]` build merges, it will be marked successful but not publish any artifacts via the "latest" permalink. Since netavark and aardvark depend on each other's latest binaries, this can lead to a chicken-and-egg problem.

While the underlying Cirrus-CI bug has now been fixed, this is still a fragile situation for CI.  At the same time, it's beneficial to developers for CI runs to happen as quickly as possible.  Meaning, CI should try to avoid compiling when possible.

Resolve both issues by attempting to download from the permalink, but if that fails, fallback to compiling and building from the dependent repo using `cargo` (original implementation by @baude).  This is a more complex solution, but should save minutes of developer time for 99.99% of PRs.

### DIFF
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -18,20 +18,28 @@ trap "complete_setup" EXIT
 msg "************************************************************"
 msg "Setting up runtime environment"
 msg "************************************************************"
-
 show_env_vars
 
-req_env_vars AARDVARK_DNS_URL
-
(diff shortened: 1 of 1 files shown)


cevich:
[Negative test](https://cirrus-ci.com/task/6080374128771072)

Luap99 (commented):
LGTM

PR #527, opened 2022-12-10
add ipvlan support

M1cha:
Status:
- it compiles and there are integration tests which pass
- I tested this with podman and was able to ping the container from another physical device on the network.
- This is a ton of copy&paste because macvlan and ipvlan are so similar, we might want to reduce that. Let's discuss this.
  - `get_default_route_interface` should be easy to move to e.g. core_utils
  - `setup` could be moved bei either using generics or by passing `CreateLinkOptions` as an argument
  - the integration tests could completely move to generated testfiles to we can use a variable for the driver name

### DIFF
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -22,6 +22,10 @@ use sysctl::{Sysctl, SysctlError};
 
 use super::netlink;
 
+pub const IPVLAN_MODE_L2: u16 = 0;
+pub const IPVLAN_MODE_L3: u16 = 1;
+pub const IPVLAN_MODE_L3S: u16 = 2;
+
 pub struct CoreUtils {
     pub networkns: String,
 }
@@ -228,6 +232,19 @@ impl CoreUtils {
             // default to bridge
             name => Err(NetavarkError::msg(format!(
                 "invalid macvlan mode \"{}\"",
+                name
+            ))),
+        }
+    }
+
+    pub fn get_ipvlan_mode_from_string(mode: &str) -> NetavarkResult<u16> {
+        match mode {
+            // default to l2 when unset
+            "" | "l2" => Ok(IPVLAN_MODE_L2),
+            "l3" => Ok(IPVLAN_MODE_L3),
+            "l3s" => Ok(IPVLAN_MODE_L3S),
+            name => Err(NetavarkError::msg(format!(
+                "invalid ipvlan mode \"{}\"",
                 name
             ))),
         }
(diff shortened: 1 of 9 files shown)


Luap99:
Thanks for doing this.

> This is a ton of copy&paste because macvlan and ipvlan are so similar, we might want to reduce that. Let's discuss this. 

Yes we should de-duplicate as much as possible before merging at least for the code. Using generics and moving logic into shared functions is totally fine.
As for the test it would be nice to not duplicate all of it but it is not a requirement for me.

baude:
@M1cha do you want to do the deduplication?  if not, I would be open to merging and one of us coming back around to do it.  Though, we could use the help!  And in that light, thank you so very much for your contribution.

M1cha:
@baude yes, I'll do that tomorrow.(Sunday for me)

M1cha:
I managed to reduce copy&paste a lot by matching on the driver name and storing vlan-kind specific data inside an enum. I hope you like it.
It might have actually been faster to do that from the beginning... another lesson learned :grin:

src/network/vlan.rs
```diff
@@ -191,7 +225,7 @@ fn setup(
                         // if last element return directly
                         if i == 2 {
                             return Err(NetavarkError::msg(format!(
-                                "create macvlan interface: {}",
+                                "create vlan interface: {}",
```
Luap99, line 228: I don't like the error message changes, it should be clear what driver failed for a user and for me when I read the bug reports. `vlan` is already used with a different meaning in networking so it might confuse users.

I think you can fix this by replacing the vlan part with `{}` and then just pass the kind enum as argument. That would require you to implement the Display trait which should just print the name, see https://doc.rust-lang.org/std/fmt/trait.Display.html

---
This comment applies to all changes were your changed the error message obviously.

src/network/core_utils.rs
```diff
@@ -22,6 +22,10 @@ use sysctl::{Sysctl, SysctlError};
 
 use super::netlink;
 
+pub const IPVLAN_MODE_L2: u16 = 0;
+pub const IPVLAN_MODE_L3: u16 = 1;
+pub const IPVLAN_MODE_L3S: u16 = 2;
+
```
Luap99, line 28: are there no const in the upstream library somewhere?

M1cha: unfortunately these are missing.

src/network/vlan.rs
M1cha: good idea, done

Luap99 (commented):
LGTM

baude:
@M1cha thank you very much for your generous contribution!

/lgtm

PR #533, opened 2022-12-13
Add ability for MacVLAN dhcp

baude:
We can now support DHCP macvlan use cases for setup.  Added plumbing for additional arguments requeired as well, such as container mac address, etc.

Test forthcoming when teardown is complete.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ netlink-packet-route = "0.13"
 netlink-packet-core = "0.4.2"
 fs2 = "0.4.3"
 netlink-sys = "0.8.3"
+netavark_proxy = { git = "https://github.com/containers/netavark-dhcp-proxy"} 
+tokio = { version = "1.0", features = ["rt"] }
 
 [build-dependencies]
 chrono = { version = "0.4.22", default-features = false, features = ["clock"] }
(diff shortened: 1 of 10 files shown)


baude:
I need someone to merge https://github.com/containers/netavark-dhcp-proxy/pull/44 so this can be retried.

Cargo.toml
```diff
@@ -38,6 +38,8 @@ netlink-packet-route = "0.13"
 netlink-packet-core = "0.4.2"
 fs2 = "0.4.3"
 netlink-sys = "0.8.3"
+netavark_proxy = { git = "https://github.com/containers/netavark-dhcp-proxy"} 
+tokio = { version = "1.0", features = ["rt"] }
```
Luap99, line 42: Any chance you can remove the tokio dep? You do not really use the async socket since you just block on it anyway. The amount of added dependencies here looks scary, building this PR shows me about a 4MB binary size increase.

baude: i think it will be required because the code is async

test/testfiles/macvlan-dhcp.json
```diff
@@ -0,0 +1,27 @@
+{
```
M1cha, line 1: Am I missing something or is this testfile currently unused?

PR #553, opened 2023-01-12
Cirrus: Improve caching effectiveness

cevich:
Prior to this commit, the cache maintained for the main branch has grown
to over 16gigabytes.  Attempts to upload it are causing frequent branch-
level CI failures due to being larger than the maximum allowable.
Attempt to fix this by introducing a cache-grooming script and altering
the cache-keys to track with dependency specification and build settings.
Content for the grooming script comes mainly from three places:

1. [Cargo documentation](https://doc.rust-lang.org/nightly/cargo/guide/build-cache.html)
2. The rust-cache github action [(typescript) source code](https://github.com/Swatinem/rust-cache/tree/master/src)
3. This [cargo home in CI section](https://doc.rust-lang.org/nightly/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci)

For branch-level cache, the grooming script will completely clobber it's
cache at loosely-random intervals determined by the Cirrus Build ID.
This is intended to prevent the cache from growing indefinitely.  It's an
non-ideal solution in a space full of very limited possibilities.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -50,16 +50,15 @@ build_task:
     # all PRs & branches will share caches with other PRs and branches
     # for a given $DEST_BRANCH and vX value.  Adjust vX if cache schema
     # changes.
-    fingerprint_key: "cargo_v2_${DEST_BRANCH}_amd64"
+    fingerprint_script: echo -e "cargo_v3_${DEST_BRANCH}_amd64\n---\n$(<Cargo.lock)\n---\n$(<Cargo.toml)"yy
     # Required to be set explicitly since fingerprint_key is also set
     reupload_on_changes: true
   targets_cache: &targets_cache
-    # Similar to cargo_cache, but holds the actual compiled artifacts. This must
-    # be scoped similar to bin_cache to avoid binary pollution across cache
-    # contexts.  For example, two PRs that happen to coincidentally change
-    # and use cache.  Adjust vX if cache schema changes.
+    # Similar to cargo_cache, but holds the actual compiled dependent artifacts.
+    # This should be scoped to a hash of the dependency-metadata lock file.
+    # Cirrus-CI will automatically use separate caches for PRs and branches.
     folder: "$CARGO_TARGET_DIR"
-    fingerprint_key: "targets_v2_${CIRRUS_TAG}${DEST_BRANCH}${CIRRUS_PR}_amd64" # Cache only within same tag, branch, or PR (branch will be 'pull/#')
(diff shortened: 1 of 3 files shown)


cevich:
For comparison:
* [Cold cache](https://cirrus-ci.com/task/6681539996549120)
* [Warm cache](https://cirrus-ci.com/task/6340669548003328)

About one minute of time was saved by re-using the targets cache.  However, downloading the cache took 43 seconds :cry:  I'm doing something wrong.

cevich:
okay, my aging memory recalls now.  The super beneficial cache to maintain is the Cargo cache.  That's the real time-saver when it comes to rebuilding stuffs.  I've added some new perpetration commands for it to the grooming script.

Luap99:
Mhh the timings do not look right,
looking at a good PR (https://github.com/containers/netavark/pull/549) I except the unit and validate task to be much faster.

As I understand the build task should take longer since it has to compile from scratch but the validate and unit test task should use the build cache from the build task. This looks way to slow now.

cevich:
@Luap99 All tasks subsequent to `build` have readonly cache.  It's possible I'm grooming "too much".  For sure the cargo-cache is a much bigger compile time-saver than the target-cache.  Which is odd, since it's the target-cache we're having size problems with.  The [Most recent cold-cache run](https://cirrus-ci.com/task/5383630978023424) for cargo-cache resulted in a 3x increase in build time. Followup (warm-cache) [run is underway now.](https://cirrus-ci.com/task/5287118600142848)

In anycase, I'll take a look at the other tasks shortly, thanks for pointing them out.

Luap99:
Timings look much better now, although I wonder why aarch64 still looks much slower compared to the older PRs. But maybe this is just a one off.

Also I found https://github.com/rust-lang/cargo/issues/5026, which explains why target dir just keeps increasing. Basically every time as the rust toolchain is updated we need clean that dir in the future.

cevich:
> But maybe this is just a one off.

The way cache is setup now, it's keyed on a pr-by-pr basis.  For branches, it's keyed on the branch-name.  So the first run in a PR or new branch, cache will be cold.  Then subsequent re-runs (in the same PR) will benefit from a warm cache.  I did it this way to thwart accidental or intentional pollution/poisoning from one PR to another.

If you think it's safe for PRs to share cache, I can make that change.  I just defaulted to "most-safe" as a matter of habit.

> Also I found https://github.com/rust-lang/cargo/issues/5026,

Oh interesting, yep that's got to be the problem.  It also explains why the github action goes to such lengths as reading config files and checking mtimes.  I didn't know there was a `cargo clean` command.  I'll use that in my "butcher's knife" hack instead of trying to manually remove files...

cevich:
Okay, I think this PR is ready now.  I tried to document what's happening as best as I could.

Luap99 (approved):
LGTM

flouthoc (approved):
LGTM
/lgtm

PR #554, opened 2023-01-13
Fix UDS path

baude:
The const for default the netavark dhcp proxy UDS path was changed to a different module in its project. This is required to continue to build against newer versions of the proxy.


mheon:
LGTM

PR #564, opened 2023-01-21
Convert a few words to more inclusive ones

TomSweeneyRedHat:
To make the code base more inclusive, I've changed a few variable names from "master*" to "primary*".

[NO NEW TESTS NEEDED]

Signed-off-by: tomsweeneyredhat <tsweeney@redhat.com>

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -531,7 +531,7 @@ fn create_veth_pair(
     host: &mut netlink::Socket,
     netns: &mut netlink::Socket,
     data: &InternalData,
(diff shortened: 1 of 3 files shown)


Luap99 (changes requested):
I disagree with this change, the correct name is master. That is how it is defined in the kernel netlink API. Renaming just makes it harder to understand the code IMO.

Luap99 (approved):
Based on internal discussion, LGTM

flouthoc (approved):
LGTM

PR #568, opened 2023-01-27
Switch to explicit platform list for `cargo vendor-filterer`

cgwalters:
See https://github.com/coreos/cargo-vendor-filterer/issues/50 This fixes an error trying to vendor for m68k.

Also while we have the patient open, move the platform list to `Cargo.toml` in declarative format.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,13 @@ categories = ["containers", "networking", "podman"]
 exclude = ["/.cirrus.yml", "/.github/*", "/hack/*"]
 build = "build.rs"
 
+[package.metadata.vendor-filter]
+# This list is not exhaustive.
+platforms = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "powerpc64le-unknown-linux-gnu", 
+             "s390x-unknown-linux-gnu", "riscv64gc-unknown-linux-gnu",
+             "x86_64-unknown-linux-musl", "aarch64-unknown-linux-musl",
+             ]
+
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
 default = ["serde", "deps-serde"]
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ validate: $(CARGO_TARGET_DIR)
 .PHONY: vendor-tarball
 vendor-tarball: build install.cargo-vendor-filterer
 	VERSION=$(shell bin/netavark --version | cut -f2 -d" ") && \
-	cargo vendor-filterer '--platform=*-unknown-linux-*' --format=tar.gz --prefix vendor/ && \
+	cargo vendor-filterer --format=tar.gz --prefix vendor/ && \
 	mv vendor.tar.gz netavark-v$$VERSION-vendor.tar.gz && \
 	gzip -c bin/netavark > netavark.gz && \
 	sha256sum netavark.gz netavark-v$$VERSION-vendor.tar.gz > sha256sum


Cargo.toml
```diff
@@ -12,6 +12,10 @@ categories = ["containers", "networking", "podman"]
 exclude = ["/.cirrus.yml", "/.github/*", "/hack/*"]
 build = "build.rs"
 
+[package.metadata.vendor-filter]
+# This list is not exhaustive.
+platforms = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu", "powerpc64le-unknown-linux-gnu", "s390x-unknown-linux-gnu", "riscv64gc-unknown-linux-gnu"]
```
Luap99, line 17: I think we should try to include all linux platforms when possible. I do not know what packagers make use of this tarball and where netavark is packaged. I think this should continue to work for as much linux targets as possible. I don't want to cause regressions.

cgwalters: Not having a platform in this list doesn't mean it won't build.  

It means that if there is a dependency that *only* exists on that platform, then the build would fail.  This is really unlikely - there's very very very few crates which e.g. *only* exist for say s390x-unknown-linux-gnu or so.

In the case that happens, it's not a big deal - the platform can be added.

Luap99: I still would like to have at least the musl targets included, I think alpine depends on this.

cgwalters: OK done, though note that today that does not change the vendored set.  (It seems really unlikely for there to really ever be a crate which is only a dependency on musl targets)

Luap99: Fair enough, however I much rather have it in there just in case.

cgwalters:
BTW, I would also recommend using e.g. GH Actions to generate a vendor tar automatically on every commit - this helps ensure it's done in a reproducible environment.  Then you would have noticed that action failing long before a release.

We have https://github.com/coreos/cargo-vendor-filterer/blob/main/.github/workflows/autovendor.yml as a demonstration of this.

lsm5:
> BTW, I would also recommend using e.g. GH Actions to generate a vendor tar automatically on every commit - this helps ensure it's done in a reproducible environment. Then you would have noticed that action failing long before a release.
> 
> We have https://github.com/coreos/cargo-vendor-filterer/blob/main/.github/workflows/autovendor.yml as a demonstration of this.

ack, i'm working on enabling packit copr-builds for our repos which should also catch this (that's how I found out too today https://github.com/containers/aardvark-dns/pull/214 ), but it's got quite a ways to go before merge.. If GHA can get in sooner, fine by me. @edsantiago @cevich wdyt ?

I noticed the builds on the podman-next copr failing too but it doesn't complain loudly or maybe my notifications get filtered.

cgwalters:
> ack, i'm working on enabling packit copr-builds for our repos

See also https://github.com/containers/bootc/pull/36/commits/b274b315c24de933e67a538a5fccbcf904854af8

lsm5:
this still occurs on local and copr:

```
error: failed to run custom build command for `netavark_proxy v0.1.0 (https://github.com/containers/netavark-dhcp-proxy#d6286927)`

Caused by:
  process didn't exit successfully: `/home/lsm5/repositories/netavark/targets/release/build/netavark_proxy-91f59c5100d51df8/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-changed=proto/proxy.proto
  cargo:rerun-if-changed=proto

  --- stderr
  thread 'main' panicked at 'Could not find `protoc` installation and this build crate cannot proceed without
      this knowledge. If `protoc` is installed and this crate had trouble finding
      it, you can set the `PROTOC` environment variable with the specific path to your
      installed `protoc` binary.If you're on debian, try `apt-get install protobuf-compiler` or download it from https://github.com/protocolbuffers/protobuf/releases

  For more information: https://docs.rs/prost-build/#sourcing-protoc
  ', /home/lsm5/.cargo/registry/src/github.com-1ecc6299db9ec823/prost-build-0.11.4/src/lib.rs:1296:10
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
make: *** [Makefile:50: build] Error 101
```

see copr log at https://download.copr.fedorainfracloud.org/results/rhcontainerbot/podman-next/fedora-rawhide-x86_64/05361945-netavark/builder-live.log.gz 

fwiw, aardvark-dns builds fine on copr now

Luap99:
> Could not find `protoc` installation and this build crate cannot proceed without

@lsm5 You need to install `protoc`, this is a hard dependency from the new dhcp proxy.

Luap99:
https://github.com/containers/netavark-dhcp-proxy#build-dependencies

lsm5:
> https://github.com/containers/netavark-dhcp-proxy#build-dependencies

thanks!

PR #597, opened 2023-02-28
Support none parameter on NETAVARK_FW

oskarirauta:
Passing environment valuepair NETAVARK_FW=none disables all firewall/portmapper related features leaving configuration of firewall to user.

### DIFF
--- a/src/firewall/fwnone.rs
+++ b/src/firewall/fwnone.rs
@@ -1,0 +1,32 @@
+use crate::firewall;
+use crate::firewall::NetavarkResult;
+use crate::network::internal_types::{
+    PortForwardConfig, SetupNetwork, TearDownNetwork, TeardownPortForward,
+};
+
+// Iptables driver - uses direct iptables commands via the iptables crate.
+pub struct Fwnone {}
+
+pub fn new() -> NetavarkResult<Box<dyn firewall::FirewallDriver>> {
+    Ok(Box::new(Fwnone {}))
+}
+
+impl firewall::FirewallDriver for Fwnone {
+    fn setup_network(&self, _network_setup: SetupNetwork) -> NetavarkResult<()> {
+        Ok(())
+    }
+
+    // teardown_network should only be called in the case of
+    // a complete teardown.
+    fn teardown_network(&self, _tear: TearDownNetwork) -> NetavarkResult<()> {
+        Ok(())
+    }
+
+    fn setup_port_forward(&self, _setup_portfw: PortForwardConfig) -> NetavarkResult<()> {
+        Ok(())
+    }
+
+    fn teardown_port_forward(&self, _tear: TeardownPortForward) -> NetavarkResult<()> {
+        Ok(())
+    }
+}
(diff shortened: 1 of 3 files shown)


src/firewall/fwnone.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl firewall::FirewallDriver for Fwnone {
+    fn setup_network(&self, _network_setup: SetupNetwork) -> NetavarkResult<()> {
+        Ok(())
+    }
+
+    // teardown_network should only be called in the case of
+    // a complete teardown.
+    fn teardown_network(&self, _tear: TearDownNetwork) -> NetavarkResult<()> {
+        Result::Ok(())
```
Luap99, line 21:
```suggestion
       Ok(())
```

src/firewall/fwnone.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    fn setup_port_forward(&self, _setup_portfw: PortForwardConfig) -> NetavarkResult<()> {
+        Result::Ok(())
```
Luap99, line 25:
```suggestion
        Ok(())
```

src/firewall/fwnone.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    fn teardown_port_forward(&self, _tear: TeardownPortForward) -> NetavarkResult<()> {
+        Result::Ok(())
```
Luap99, line 29:
```suggestion
        Ok(())
```

Luap99 (commented):
Thanks for the PR, to pass CI you have to run `cargo fmt` to fix formatting issues.

Could you also add a new test file for the none driver in test/ directory?
Something like this should be enough to check https://github.com/containers/netavark/blob/70eee078a9d194d3f93cbd74666ceb687a6dac0c/test/100-bridge-iptables.bats#L10-L13
make sure you set NETAVARK_FW=none

oskarirauta:
@Luap99 
Check if it's ok now..

Luap99:
looks good, could you squash the commits please.

@mheon PTAL

oskarirauta:
@Luap99 

Unfortunately I do not know how to do it properly- I do know how to rebase, issue is that I should had made the PR some time ago, when we talked about this on issues- now I have synced it- so there's a merge commit in between where I synced it before doing this PR-

It looks a bit like this:
 - last mainstream commit before I started working
 - my work here
 - sync with origin/master
 - cargo fmt commit
 - remove unnecessary result::
 - add firewall none test

So I could squash them to cargo fmt commit, but if I squash to sync commit, or even to _my work here_-commit, it will be a problem... Any good ideas, except make a new branch? That would be easiest solution for me, but if you know a "magic trick" to make it happen, I am all ears..

mheon:
LGTM. I do start to wonder if we should add a way for Podman to request a specific driver so it can actually request this on a per-network basis.

oskarirauta:
@mheon I have already requested it for some time ago.. Until that, I'll go with a wrapper script.

oskarirauta:
PR has been rebased succesfully.

Luap99 (approved):
LGTM

PR #596, opened 2023-02-23
Merge nv-dhcp-proxy into nv

baude:
The proxy is nearing completion and having them in separate repos has become problematic geiven that they both "use" each other.

PR #603, opened 2023-03-03
spec.rpkg: list netavark-dhcp-proxy in files list

lsm5:
/usr/libexec/podman/netavark-dhcp-proxy is now installed so the spec file should mention it in the files list.


lsm5:
nvm, looks like it's handled in https://github.com/containers/netavark/pull/596/commits/fdc86933d2df6f27fcde430ad0986a678ad654ba

PR #611, opened 2023-03-08
packit: Update build release tag correctly

lsm5:
.packit.sh was searching for the wrong release string and as a result not replacing anything, resulting in all packit builds ending up with release tag `1`.

Ref: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/package/netavark/

This commit will correctly update the release tag.


Luap99 (commented):
LGTM

PR #617, opened 2023-03-10
update: should allow empty `network_dns_servers`

flouthoc:
`podman network update test --dns-drop 8.8.8.8` converges to a case when there are no nameservers for a network and netavark handles it correctly here: https://github.com/containers/netavark/blob/main/src/dns/aardvark.rs#L366

So allow users to update exisiting network with empty list of `network_dns_servers`


Luap99:
Please add a test for it.

flouthoc:
@Luap99 Done.

Luap99 (approved):
LGTM

flouthoc:
@mheon @baude PTAL

PR #619, opened 2023-03-10
combine dhcp-proxy and netavark into one binary

### LINKED ISSUE #608: compine netavark and dhcp-proxy into one binary
Now that we merged the repos I suggest that we merge both binaries into one, i. e. instead of calling `netavark-dhcp-proxy` we would call `netavark dhcp-proxy` and implement it as subcommand.

The main benefit would be reduced binary size. Because rust libs are statically linked and both binaries mostly use the same code they are duplicated in each binary. If we combine them it would reduce the overall size.

@baude @mheon @flouthoc WDYT?

Luap99:
The proxy will now be called as `netavark dhcp-proxy ...`. The new binary is about 18 MB big, this safes about 10 MB in space compared to both binaries together.

Fixes #608


baude:
LGTM

mheon:
LGTM

flouthoc (approved):
LGTM

PR #620, opened 2023-03-11
Include Systemd Unit Files in RPM Package

apsoyka:
Currently, the COPR package does not include the Systemd unit files under `contrib/systemd/system`, which are desirable for use with the new DHCP proxy functionality. This PR adds these unit files to the COPR RPM package file, as well as the Makefile install and uninstall tasks.

### DIFF
--- a/netavark.spec.rpkg
+++ b/netavark.spec.rpkg
@@ -34,6 +34,8 @@ BuildRequires: git-core
 BuildRequires: go-md2man
 BuildRequires: protobuf-c
 BuildRequires: protobuf-compiler
+%{?systemd_requires}
+BuildRequires: systemd
 Recommends: aardvark-dns
 Provides: container-network-stack
 %if 0%{?fedora}
@@ -58,12 +60,20 @@ popd
 %install
 %{__make} DESTDIR=%{buildroot} PREFIX=%{_prefix} install
 
+%preun
+%systemd_preun %{name}-dhcp-proxy.service
+
+%postun
+%systemd_postun %{name}-dhcp-proxy.service
+
 %files
 %license LICENSE
 %dir %{_libexecdir}/podman
 %{_libexecdir}/podman/%{name}
 %{_libexecdir}/podman/%{name}-dhcp-proxy
 %{_mandir}/man1/%{name}.1*
+%{_unitdir}/%{name}-dhcp-proxy.service
+%{_unitdir}/%{name}-dhcp-proxy.socket
 
 %changelog
 {{{ git_dir_changelog }}}
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ DESTDIR ?=
 PREFIX ?= /usr/local
 LIBEXECDIR ?= ${PREFIX}/libexec
 LIBEXECPODMAN ?= ${LIBEXECDIR}/podman
+SYSTEMDDIR ?= ${PREFIX}/lib/systemd/system
 
 SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)
 # Get crate version by parsing the line that starts with version.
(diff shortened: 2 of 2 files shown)


netavark.spec.rpkg
```diff
@@ -58,12 +61,27 @@ popd
...
+if [ $1 -eq 1 ]; then
+    %{_bindir}/systemctl enable --now %{name}-dhcp-proxy.service
+fi
+
+%preun
+%systemd_preun %{name}-dhcp-proxy.service
+
+%postun
+%systemd_postun %{name}-dhcp-proxy.service
+
```
Luap99, line 74: I am not fully understand what these macros do, looking at https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd does not really help me.

However I do not think we should enable it by default and it should not be restarted on upgrade otherwise the dhcp leases can fail.

netavark.spec.rpkg
```diff
@@ -58,12 +61,27 @@ popd
...
 %files
 %license LICENSE
 %dir %{_libexecdir}/podman
 %{_libexecdir}/podman/%{name}
 %{_libexecdir}/podman/%{name}-dhcp-proxy
 %{_mandir}/man1/%{name}.1*
+%{_unitdir}/%{name}.service
+%{_unitdir}/%{name}.socket
+%{_userunitdir}/%{name}.service
+%{_userunitdir}/%{name}.socket
```
Luap99, line 84: same here

Luap99 (changes requested):
Thanks for doing this, this was on my TODO list before the release.

@lsm5 PTAL

netavark.spec.rpkg
```diff
@@ -34,6 +34,9 @@ BuildRequires: git-core
 BuildRequires: go-md2man
 BuildRequires: protobuf-c
 BuildRequires: protobuf-compiler
+%{?systemd_requires}
+BuildRequires: systemd
+BuildRequires: systemd-devel
```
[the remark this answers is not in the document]
apsoyka: I don't think we need the `systemd-devel` package either, I have removed it from the RPKG spec file.

netavark.spec.rpkg
lsm5: @Luap99 try `rpm --eval "%systemd_post netavark-dhcp-proxy.service"` which will expand the macro. Similar for %systemd_preun and %systemd_postun / %systemd_postun_with_restart.

```
$ rpm --eval "%systemd_post netavark-dhcp-proxy.service"


if [ $1 -eq 1 ] && [ -x "/usr/lib/systemd/systemd-update-helper" ]; then
    # Initial installation
    /usr/lib/systemd/systemd-update-helper install-system-units netavark-dhcp-proxy.service || :
fi
```

If we don't want it to be started on initial install and also not restarted on upgrade, we can keep the `%systemd_preun` and `%systemd_postun` parts and only remove `%post`. Restart on upgrade would be handled by `%systemd_postun_with_restart` which isn't here anyway.

netavark.spec.rpkg
```diff
@@ -58,12 +60,25 @@ popd
...
+%systemd_postun %{name}-dhcp-proxy.service
+
 %files
 %license LICENSE
 %dir %{_libexecdir}/podman
 %{_libexecdir}/podman/%{name}
 %{_libexecdir}/podman/%{name}-dhcp-proxy
 %{_mandir}/man1/%{name}.1*
+%{_unitdir}/%{name}.service
+%{_unitdir}/%{name}.socket
```
lsm5, line 81: these need to be `%{name}-dhcp-proxy` instead of just `%{name}`. 

Right now the packit rpm builds are failing with:
```
RPM build errors:
    source_date_epoch_from_changelog set but %changelog is missing
    File not found: /builddir/build/BUILDROOT/netavark-1.6.0~dev-1.20230314174247870645.pr620.58.g722fa8a.fc39.x86_64/usr/lib/systemd/system/netavark.service
    File not found: /builddir/build/BUILDROOT/netavark-1.6.0~dev-1.20230314174247870645.pr620.58.g722fa8a.fc39.x86_64/usr/lib/systemd/system/netavark.socket
```

apsoyka: Thank you for catching this 🙂

netavark.spec.rpkg
```diff
@@ -58,12 +60,25 @@ popd
 %install
 %{__make} DESTDIR=%{buildroot} PREFIX=%{_prefix} install
 
+%post
+if [ $1 -eq 1 ]; then
+    %{_bindir}/systemctl enable --now %{name}-dhcp-proxy.service
+fi
```
Luap99, line 66: I still do not think we should enable it, also if we need would want to enable by default it should enable the socket not service.

Luap99 (commented):
Thanks, could you squash the commits one please?

apsoyka:
Okay, I squashed the commits, and the systemd service is no longer enabled by default.

Luap99 (approved):
LGTM

@baude @lsm5 PTAL

lsm5 (approved):
/approve
/lgtm

PR #639, opened 2023-03-22
Quit Leasing when Client disconnects

baude:
Attention:
This is a rebase of
https://github.com/containers/netavark-dhcp-proxy/pull/61 which was written by Jack Baude. Jack is tied up in school and I simply performed the rebase to netavark after ithe netavark and netavark-dhcp-proxy projects were merged.

In the case where a client asks the proxy server for a leased address, it is possible that the client would quit (hangup) before the server could respond.  Or more likely, the server is having difficulty obtaining a lease and the client quits.  In either case, when the client quits, the server needs to stop trying to aquire a lease.

* start using mozim 0.2.1 with async lease capabilities.
* fixed timeout type to match mozim 0.2.1
* detect client disconnection with tokio oneshot channel
* use async lease DORA

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,12 +56,12 @@ netlink-sys = "0.8.4"
 tokio = { version = "1.26", features = ["rt", "rt-multi-thread", "signal", "fs"] }
 tokio-stream = { version = "0.1", features = ["net"] }
 tonic = "0.8"
-mozim = "0.2"
+mozim = "0.2.1"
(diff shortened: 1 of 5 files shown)


Luap99:
lint is not happy

Luap99 (commented):
Is there any chance to add a test? Although I guess the dhcp test are already flaking a lot maybe better to not add another timing based test.

LGTM

PR #640, opened 2023-03-23
feat: added the --firewall-driver option

unknowndevQwQ:
Change the firewall driver with the command option

Signed-off-by: unknowndevQwQ <unknowndevQwQ@pm.me>

### DIFF
--- a/src/firewall/mod.rs
+++ b/src/firewall/mod.rs
@@ -3,7 +3,6 @@ use crate::network::internal_types::{
     PortForwardConfig, SetupNetwork, TearDownNetwork, TeardownPortForward,
 };
 use log::{debug, info};
-use std::env;
 use zbus::blocking::Connection;
 
 pub mod firewalld;
@@ -45,13 +44,9 @@ enum FirewallImpl {
 /// What firewall implementations does this system support?
 fn get_firewall_impl(driver_name: Option<String>) -> NetavarkResult<FirewallImpl> {
     // It respects "firewalld", "iptables", "nftables", "none".
-
-    // If not requested lookup in NETAVARK_FW env var as well.
-    let driver = driver_name.or_else(|| env::var("NETAVARK_FW").ok());
-
-    if let Some(var) = driver {
-        debug!("Forcibly using firewall driver {}", var);
-        match var.to_lowercase().as_str() {
+    if let Some(driver) = driver_name {
+        debug!("Forcibly using firewall driver {driver}");
+        match driver.to_lowercase().as_str() {
             FIREWALLD => {
                 let conn = match Connection::system() {
                     Ok(c) => c,
--- a/src/main.rs
+++ b/src/main.rs
@@ -15,6 +15,10 @@ struct Opts {
     /// Instead of reading from STDIN, read the configuration to be applied from the given file.
     #[clap(short, long)]
     file: Option<OsString>,
+    /// Select netavark's firewall driver
+    // There is no suitable short argument like -F, so there is no short argument.
+    #[clap(long, env = "NETAVARK_FW")]
+    firewall_driver: Option<String>,
     /// config directory for aardvark, usually path to a tmpfs.
     #[clap(short, long)]
     config: Option<OsString>,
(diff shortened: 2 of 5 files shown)


src/main.rs
```diff
@@ -12,6 +12,9 @@ struct Opts {
     /// Instead of reading from STDIN, read the configuration to be applied from the given file.
     #[clap(short, long)]
     file: Option<String>,
+    /// Select netavark's firewall driver
+    #[clap(short='i', long, env="NETAVARK_FW")]
```
Luap99, line 16: just remove the short option, we do not use the cli directly so it serves little to no purpose anyway.

Luap99 (commented):
Thanks for the contribution. 
We require a sign off as described here: https://github.com/containers/podman/blob/main/CONTRIBUTING.md#sign-your-prs
Also make sure to fix the validate issue from the CI logs, https://cirrus-ci.com/task/5515233758806016

src/main.rs
unknowndevQwQ: > just remove the short option, we do not use the cli directly so it serves little to no purpose anyway.

Currently netavark has no way of automatically completing parameters, short parameters may be useful in the future.

unknowndevQwQ:
@Luap99 Is it okay now?

Luap99:
Please squash your commits into one

src/main.rs
Luap99: Netavark is not intended to be ever executed manually on a CLI, I rather just have the long form. Using `-i` just fells random, who is going to remember what it means?

unknowndevQwQ: > Netavark is not intended to be ever executed manually on a CLI, I rather just have the long form. Using `-i` just fells random, who is going to remember what it means?

I decided to give up until I found the right short parameters.

unknowndevQwQ:
> Please squash your commits into one

I'm ready.

unknowndevQwQ:
@Luap99 RUOK?

maxxberg:
Any update?

mheon:
We're finally getting around to getting the other firewall drivers plumbed in (expect the nftables driver to appear in a PR sometime next week, and firewalld will be finalized sometime after that), so if this is rebased I would be OK with merging

unknowndevQwQ:
> We're finally getting around to getting the other firewall drivers plumbed in (expect the nftables driver to appear in a PR sometime next week, and firewalld will be finalized sometime after that), so if this is rebased I would be OK with merging

I need to spend some time generating an aesthetically pleasing gpg key, or rebase it later in github.dev

src/firewall/mod.rs
```diff
@@ -45,13 +44,9 @@ enum FirewallImpl {
...
-
-    // If not requested lookup in NETAVARK_FW env var as well.
-    let driver = driver_name.or_else(|| env::var("NETAVARK_FW").ok());
-
-    if let Some(var) = driver {
-        debug!("Forcibly using firewall driver {}", var);
-        match var.to_lowercase().as_str() {
+    if let Some(driver) = driver_name {
+        debug!("Forcibly using firewall driver {driver}");
+        match driver.to_lowercase().as_str() {
```
mheon, line 49: Can we retain the environment variable here if driver_name is is none?

mheon:
One change requested otherwise LGTM

src/firewall/mod.rs
unknowndevQwQ: > Can we retain the environment variable here if driver_name is is none?

I'm already getting the value from the environment variable via clap's env attribute, so it should be consistent with the original behavior when no arguments are specified.

mheon: Clearly I need more coffee

mheon:
@Luap99 PTAL, this should be ready

src/firewall/mod.rs
unknowndevQwQ: > Clearly I need more coffee

But you don't look like you've been up all night or anything.

Luap99 (approved):
/lgtm

Issue #641, opened 2023-03-23
[Rootful] When binding multiple ports, Netavark fails to create all necessary DNAT IPv6 rules


JCBird1012:
Netavark seems to fail to add some DNAT rules on the IPv6 side when binding multiple ports. I'm running Netavark built from @Luap99's `wildcard-port` branch on their forked repo, but this happens on v1.5.0 too (so unless this got fixed recently in mainline, it probably wasn't a regression due to that fix)...

####   This is using Netavark built from [Luap99\netavark](https://github.com/Luap99/netavark)'s `wildcard-port` branch - 
```
[root@scw-lucid-cori ~]# /usr/lib/podman/netavark -V
netavark 1.6.0-dev
```
---
1. Create custom IPv6 bridge network
```
[root@scw-lucid-cori ~]# podman network create ipv6 --ipv6
ipv6
```

2. Run a container with an app that listens on multiple ports... I chose `nginx` here.
```
[root@scw-lucid-cori nginx]# cat nginx.conf
worker_processes  auto;

error_log  /dev/null;
pid        /var/run/nginx.pid;

events {
    worker_connections  4096;
    multi_accept on;
    use epoll;
}


http {
    sendfile on;
    tcp_nopush on;
    tcp_nodelay on;

    gzip off;
    server {
      listen [::]:80 ipv6only=off;
      return 200 'Hello!';
    }

    server {
      listen [::]:443 ipv6only=off;
      return 200 'Hello!';
  }
}

[root@scw-lucid-cori nginx]# podman run -d --network=ipv6 -p 80:80 -p 443:443 -v './nginx.conf:/etc/nginx/nginx.conf' nginx
bef36518be2e69f1da2e3bd1987b8f87ac162f20413098f621d1058a4b79c0a1
```

3. Check the bound ports (I'm using the host's external IPs here)
```
[root@scw-lucid-cori nginx]# curl -v4 163.172.153.244:80
*   Trying 163.172.153.244:80...
* Connected to 163.172.153.244 (163.172.153.244) port 80 (#0)
> GET / HTTP/1.1
> Host: 163.172.153.244
> User-Agent: curl/8.0.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: nginx/1.23.3
< Date: Thu, 23 Mar 2023 01:25:12 GMT
< Content-Type: text/plain
< Content-Length: 6
< Connection: keep-alive
<
* Connection #0 to host 163.172.153.244 left intact
Hello!

[...]

PR #642, opened 2023-03-23
iptables: fix incorrect debug statement

Luap99:
We should only log that the rule exists when this is the case. I noticed this while debugging #641.


Luap99:
@mheon @baude @flouthoc PTAL

flouthoc (approved):
LGTM

PR #646, opened 2023-03-28
update,main: do not write empty space when no `network_dns_servers` is provided.

flouthoc:
Pass `network_dns_servers` empty when its needed (https://github.com/containers/netavark/blob/main/src/dns/aardvark.rs#L357 ) and not as a single element with empty string and don't write false config for `aardvark-dns`.


flouthoc:
Issue was initially reported by @ypu.

flouthoc:
@Luap99 @mheon @baude PTAL

flouthoc:
@Luap99 Done.

Luap99 (approved):
LGTM

PR #661, opened 2023-03-31
fix some systemd dhcp-proxy issues

Luap99:
Do not remove the socket when systemd socket activated.
Use correct binary path when installing the unit file with make install.
see commit for details


baude:
LGTM

Luap99:
@flouthoc PTAL

flouthoc (approved):
LGTM
/lgtm

PR #664, opened 2023-04-04
dhcp-proxy: use better error when connection fails

Luap99:
Wrap the tonic status errors in the netavark error type, this allows us to create better error messages.

Also a bit complicated but special case when the socket is not found and return a message to consider enabling the netavark-dhcp-proxy.socket. old error:
`unable to obtain lease: status: Internal, message: "transport error", details: [], metadata: MetadataMap { headers: {} }` new error:
`unable to obtain lease: socket "/run/podman/nv-proxy.sock": No such file or directory (os error 2), is the netavark-dhcp-proxy.socket enabled?`

### DIFF
--- a/src/dhcp_proxy/lib.rs
+++ b/src/dhcp_proxy/lib.rs
@@ -1,7 +1,7 @@
 extern crate core;
 
 use crate::dhcp_proxy::lib::g_rpc::{Lease, NetworkConfig};
-
+use crate::error::NetavarkError;
 use std::convert::TryFrom;
 use std::error::Error;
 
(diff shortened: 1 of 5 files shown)


baude:
yeah code LGTM

src/dhcp_proxy/lib.rs
```diff
@@ -174,27 +174,50 @@ impl NetworkConfig {
...
     ///
     /// ```
-    async fn get_client(p: String) -> Result<NetavarkProxyClient<Channel>, Status> {
+    async fn get_client(p: String) -> Result<NetavarkProxyClient<Channel>, NetavarkError> {
         // We do not know why the uds connections need to be done like this.  The
         // maintainer suggested it is part of the their API.
-        let endpoint = Endpoint::try_from("http://[::1]:10000")
-            .map_err(|e| Status::internal(e.to_string()))?;
+        // We know this is safe and if it ever fails test will catch it
+        let endpoint = Endpoint::try_from("http://[::1]").unwrap();
```
rlanting, line 188: This is missing the port from the current version, is that intended?

Luap99: yes

PR #665, opened 2023-04-04
dhcp-proxy: correctly exit > 0 on errors

Luap99:
If there is an error during setup we should return the error so that netavark can exit > 0 and systemd will report this as failure.


baude:
LGTM

PR #666, opened 2023-04-05
Releasev1.6.0

baude:
* Now supports a driver plugin module for user defined network drivers
* Initial MACVLAN DHCP support (additional unit file required for packagers)
* Dependency updates


baude:
i believe lsm is on PTO this week and we likely need packaging changes ... should we cut the release or hold it right here?

PR #677, opened 2023-04-20
aardvark,network-update: no error when aardvark config is not there

flouthoc:
`network update` maybe called by podman when no container is running in such case no aardvark config is present so netavark must return as-in without throwing error since podman database is still updated.

Closes:
BZ 2188524
BZ 2182898

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -329,6 +329,9 @@ impl Aardvark {
 
     // Modifies network dns_servers for a specific network and notifies aardvark-dns server
     // with the change.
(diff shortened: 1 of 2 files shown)


flouthoc:
@Luap99 PTAL

Luap99 (commented):
change looks but but please add a test and make clear in the `modify_network_dns_servers` function comment that it will return without doing anything when the file does not exists.

Luap99 (approved):
LGTM

flouthoc:
@baude @mheon Merge me pls.

PR #680, opened 2023-04-24
Static routes

Cydox:
This PR adds the ability to configure static routes. Discussion is in #678 

Supports both ipv4 and ipv6 routes including optional route metric. 

Example json:

```
"network_info": {
        "podman": {
            "dns_enabled": false,
            "driver": "bridge",
            "id": "53ce4390f2adb1681eb1a90ec8b48c49c015e0a8d336c197637e7f65e365fa9e",
            "internal": false,
            "ipv6_enabled": false,
            "name": "podman",
            "network_interface": "podman0",
            "subnets": [
                {
                    "gateway": "10.88.0.1",
                    "subnet": "10.88.0.0/16"
                }
            ],
            "routes": [
                {
                    "destination": "10.89.0.0/24",
                    "gateway": "10.88.0.2",
                    "metric": "120"
                },
                {
                    "destination": "10.90.0.0/24",
                    "gateway": "10.88.0.3"
                }
            ]
        }
    }
```

### DIFF
--- a/src/network/vlan.rs
+++ b/src/network/vlan.rs
@@ -223,6 +223,11 @@ impl driver::NetworkDriver for Vlan<'_> {
             )?
         }
 
+        let routes = core_utils::create_route_list(&self.info.network.routes)?;
+        for route in routes.iter() {
+            netlink_sockets.1.del_route(route)?;
+        }
+
         netlink_sockets.1.del_link(netlink::LinkID::Name(
             self.info.per_network_opts.interface_name.to_string(),
         ))?;
@@ -339,6 +344,11 @@ fn setup(
         .wrap(format!("set {} up", kind_data))?;
 
     core_utils::add_default_routes(netns, &data.ipam.gateway_addresses, data.metric)?;
+
+    // add static routes
+    for route in data.ipam.routes.iter() {
+        netns.add_route(route)?
+    }
 
     get_mac_address(dev.nlas)
 }
--- a/test/100-bridge-iptables.bats
+++ b/test/100-bridge-iptables.bats
@@ -111,6 +111,41 @@ fw_driver=iptables
     expected_rc=1 run_in_host_netns ip addr show podman0
 }
 
+@test "$fw_driver - bridge with static routes" {
+    # add second interface and routes through that interface to test proper teardown
+    run_in_container_netns ip link add type dummy
+    run_in_container_netns ip a add 10.91.0.10/24 dev dummy0
+    run_in_container_netns ip link set dummy0 up
+
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-staticroutes.json setup $(get_container_netns_path)
+
+    # check static routes
+    run_in_container_netns ip r
+    assert "$output" "=~" "10.89.0.0/24 via 10.88.0.2" "static route not set"
+    assert "$output" "=~" "10.90.0.0/24 via 10.88.0.3" "static route not set"
+    assert "$output" "=~" "10.92.0.0/24 via 10.91.0.1" "static route not set"
+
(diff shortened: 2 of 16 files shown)


Cydox:
The podman side of this PR is in [https://github.com/Cydox/podman/tree/static-routes](https://github.com/Cydox/podman/tree/static-routes). It's missing tests. Little tedious to get the tests to run and it's little late now, so I will do that tomorrow.

Cydox:
I have tests for the podman side now at [https://github.com/Cydox/podman/tree/static-routes](https://github.com/Cydox/podman/tree/static-routes). But obviously the tests will only work with a netavark version that includes this PR, so I will wait before I open that PR.

Cydox:
Routes are now removed before the link gets destroyed. If netlink can't remove a route, a message is printed and the rest of the teardown continues (somebody could have removed a route by hand. In that case the teardown should simply continue).

Also squashed the commits.

The list of routes to be added gets generated in get_ipam_addresses. So currently this only happens for HOST_LOCAL / None. Not sure if this is the best behavior, however maybe for DHCP something different should be done as that can contain route information on its own.

Cydox:
Tests, failed cause I accidentally pasted something into an error message... fixed.

Cydox:
This PR already works with the changes `no_auto_gateway` flag in containers/podman#18369

src/network/vlan.rs
```diff
@@ -223,6 +223,14 @@ impl driver::NetworkDriver for Vlan<'_> {
             )?
         }
 
+        let routes = core_utils::create_route_list(&self.info.network.routes)?;
+        for route in routes.iter() {
+            netlink_sockets
+                .1
+                .del_route(route)
+                .unwrap_or_else(|_| println!("unable to delete route {}, continuing", route))
```
Luap99, line 231: same here, do not print to stdout

test/100-bridge-iptables.bats
```diff
@@ -111,6 +111,93 @@ fw_driver=iptables
...
+    # check that the loopback adapter is up
+    run_in_container_netns ip addr show lo
+    assert "$output" =~ "127.0.0.1" "Loopback adapter is up (has address)"
+
+    run_in_host_netns ping -c 1 10.88.0.2
+
+    # check static routes
+    run_in_container_netns ip r
+    assert "$output" "=~" "10.89.0.0/24 via 10.88.0.2" "static route not set"
+    assert "$output" "=~" "10.90.0.0/24 via 10.88.0.3" "static route not set"
```
Luap99, line 151: this seems to be the main part of the test, please drop the iptables/ping check, etc... Looks like this is mostly copy pasted. It should already be covered by the other tests so only checking that you have the expected routes should be good enough IMO. Just make sure you also check that teardown removes the routes. Same for the other tests.

Luap99 (changes requested):
Sorry for the delay, main logic looks good. Just some error and test tweaking required.

Cydox:
Updated this PR with the requested changes.
1. No printing to stdout
2. changed error type of create_route_list to NetavarkError
3. simplified tests to just test static routes
4. tests now check teardown (including routes that would not get deleted by the kernel automatically)
5. added tests for `no_auto_gateway`

Luap99 (approved):
LGTM @mheon @baude @flouthoc PTAL

@Cydox The `no_auto_gateway` option isn't really used here. It just prevents podman or rather c/common/libnetwork from adding the gateway to the config, right?

flouthoc (approved):
LGTM

Cydox:
> @Cydox The `no_auto_gateway` option isn't really used here. It just prevents podman or rather c/common/libnetwork from adding the gateway to the config, right?

Correct. Still wanted to add a test to make sure that netavark does behave correctly with this option.

PR #682, opened 2023-04-25
aardvark-dns pid: return better errors

Luap99:
Ed reported a flake in podman where aardvark-dns pid was not found hen we tried to stop it. It is not clear why but the error report is far from optimal, we should never throw away the original error context. This is very important to understand what is going on.

Port most over to our netavark error type and wrap where needed. Also change the pid type to pid_t. It is i32 so it did not cause any issues before but this makes it more obvious that we have the correct type.

[1] https://github.com/containers/podman/issues/18325


flouthoc (approved):
LGTM

mheon:
Sure, LGTM
/lgtm

PR #685, opened 2023-04-28
simplify path printing

Luap99:
Just to remove some code that is not needed. The Debug trait will already print the path name with quotes so we can remove them from the format string.


baude:
LGTM

Luap99:
@mheon @flouthoc PTAL

PR #697, opened 2023-05-22
add no_default_gateway option

Cydox:
This PR adds the option to enable no_default_gateway in the options map. The result is that netavark will not add default routes to the network namespace.

This fixes the issue that DNS would not work in containers/common#1440 when using no_auto_gateway. The solution is to use this new no_default_gateway instead of no_auto_gateway.

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -16,7 +16,10 @@ use crate::{
 };
 
 use super::{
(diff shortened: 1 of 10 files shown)


Luap99 (approved):
LGTM, I think this is cleaner and better to document and understand for users as well.

@flouthoc @mheon PTAL

mheon:
Sure, LGTM

Cydox:
For clarification: The options is correctly called no_default_route as discussed on matrix. In the title of this PR it is incorrectly called no_default_gateway.

PR #703, opened 2023-05-29
feature: isolation among all netavark bridge 

yassi-github:
Currently, it appears that isolation is only enabled when both bridges are in isolate mode.  
This means that communication with networks that do not have isolate enabled is still possible.

What about using `isolate=strict` to deny communication with networks that do not have isolate enabled?  
Specifically, add a new chain of iptables and add a DROP rule to the chain for networks that do not have isolate enabled.

Here is a table showing the connectivity between networks before and after this change:

before:

```
|send\recv  |nonisol1   |nonisol1   |isol1      |isol2      |
|-----------|-----------|-----------|-----------|-----------|
|nonisol1   |     o     |     o     |     o     |     o     |
|nonisol2   |     o     |     o     |     o     |     o     |
|isol1      |     o     |     o     |     o     |     x     |
|isol2      |     o     |     o     |     x     |     o     |
```

after:

```
|send\recv  |nonisol1   |nonisol1   |isol1      |isol2      |strictisol1|strictisol2|
|-----------|-----------|-----------|-----------|-----------|-----------|-----------|
|nonisol1   |     o     |     o     |     o     |     o     |     x     |     x     |
|nonisol2   |     o     |     o     |     o     |     o     |     x     |     x     |
|isol1      |     o     |     o     |     o     |     x     |     x     |     x     |
|isol2      |     o     |     o     |     x     |     o     |     x     |     x     |
|strictisol1|     x     |     x     |     x     |     x     |     o     |     x     |
|strictisol2|     x     |     x     |     x     |     x     |     x     |     o     |
```

### DIFF
--- a/test/100-bridge-iptables.bats
+++ b/test/100-bridge-iptables.bats
@@ -630,46 +630,142 @@ EOF
 }
 
 @test "$fw_driver - isolate networks" {
+    # create container/networks with isolation
+
+    # isolate1: 10.89.0.2/24, fd90::2, isolate=true
     run_netavark --file ${TESTSDIR}/testfiles/isolate1.json setup $(get_container_netns_path)
 
+    # isolate2: 10.89.1.2/24, fd99::2, isolate=true
     create_container_ns
     run_netavark --file ${TESTSDIR}/testfiles/isolate2.json setup $(get_container_netns_path 1)
 
+    # isolate3: 10.89.2.2/24, fd92::2, isolate=strict
+    create_container_ns
+    run_netavark --file ${TESTSDIR}/testfiles/isolate3.json setup $(get_container_netns_path 2)
+
+    # isolate4: 10.89.3.2/24, fd93::2, isolate=strict
+    create_container_ns
+    run_netavark --file ${TESTSDIR}/testfiles/isolate4.json setup $(get_container_netns_path 3)
+
     # check iptables NETAVARK_ISOLATION_1 chain
     run_in_host_netns iptables -S NETAVARK_ISOLATION_1
-    assert "${lines[1]}" == "-A NETAVARK_ISOLATION_1 -i isolate2 ! -o isolate2 -j NETAVARK_ISOLATION_2"
-    assert "${lines[2]}" == "-A NETAVARK_ISOLATION_1 -i isolate1 ! -o isolate1 -j NETAVARK_ISOLATION_2"
+    assert "${lines[1]}" == "-A NETAVARK_ISOLATION_1 -i isolate4 ! -o isolate4 -j NETAVARK_ISOLATION_3"
+    assert "${lines[2]}" == "-A NETAVARK_ISOLATION_1 -i isolate3 ! -o isolate3 -j NETAVARK_ISOLATION_3"
+    assert "${lines[3]}" == "-A NETAVARK_ISOLATION_1 -i isolate2 ! -o isolate2 -j NETAVARK_ISOLATION_2"
+    assert "${lines[4]}" == "-A NETAVARK_ISOLATION_1 -i isolate1 ! -o isolate1 -j NETAVARK_ISOLATION_2"
 
     run_in_host_netns iptables -nvL FORWARD
     assert "${lines[2]}" =~ "NETAVARK_ISOLATION_1"
 
     # check iptables NETAVARK_ISOLATION_2 chain
     run_in_host_netns iptables -S NETAVARK_ISOLATION_2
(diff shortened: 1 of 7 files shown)


yassi-github:
@Luap99 @flouthoc 
Is there anything else left to do for the review?
I'm new to OSS so sorry if I'm doing something wrong.

Luap99:
@yassi-github No everything is good, I will try to take a look this week. I am a bit busy right now.

yassi-github:
@Luap99 OK, Thank you for your time. :smiley:

Luap99:
@mheon PTAL, can this be done with firewalld? I would like to avoid adding firewall driver specific options that we cannot support in other backends. Once we enable other backends we have to implement there as well. The problem is end users have no idea about what driver will be used and I really do not want critical security options only work with one implementation.

test/100-bridge-iptables.bats
```diff
@@ -635,41 +635,91 @@ EOF
...
+    expected_rc=1 run_in_container_netns ping -w 1 -c 1 10.89.3.2
     expected_rc=1 run_in_container_netns 1 ping -w 1 -c 1 10.89.0.2
+    expected_rc=1 run_in_container_netns 1 ping -w 1 -c 1 10.89.2.2
+    expected_rc=1 run_in_container_netns 1 ping -w 1 -c 1 10.89.3.2
+    expected_rc=1 run_in_container_netns 2 ping -w 1 -c 1 10.89.0.2
+    expected_rc=1 run_in_container_netns 2 ping -w 1 -c 1 10.89.1.2
+    expected_rc=1 run_in_container_netns 2 ping -w 1 -c 1 10.89.3.2
+    expected_rc=1 run_in_container_netns 3 ping -w 1 -c 1 10.89.0.2
+    expected_rc=1 run_in_container_netns 3 ping -w 1 -c 1 10.89.1.2
+    expected_rc=1 run_in_container_netns 3 ping -w 1 -c 1 10.89.2.2
```
Luap99, line 683: Please add comments here what checks what, i.e from network x to y. If one thing here fails it is a nightmare to debug.

test/100-bridge-iptables.bats
```diff
@@ -635,41 +635,91 @@ EOF
...
+    expected_rc=1 run_in_container_netns ping -w 1 -c 1 fd93::2
     expected_rc=1 run_in_container_netns 1 ping -w 1 -c 1 fd90::2
+    expected_rc=1 run_in_container_netns 1 ping -w 1 -c 1 fd92::2
+    expected_rc=1 run_in_container_netns 1 ping -w 1 -c 1 fd93::2
+    expected_rc=1 run_in_container_netns 2 ping -w 1 -c 1 fd90::2
+    expected_rc=1 run_in_container_netns 2 ping -w 1 -c 1 fd99::2
+    expected_rc=1 run_in_container_netns 2 ping -w 1 -c 1 fd93::2
+    expected_rc=1 run_in_container_netns 3 ping -w 1 -c 1 fd90::2
+    expected_rc=1 run_in_container_netns 3 ping -w 1 -c 1 fd99::2
+    expected_rc=1 run_in_container_netns 3 ping -w 1 -c 1 fd92::2
```
Luap99, line 702: same here

Luap99 (commented):
Also could you squash the commits please.

Luap99:
@mheon What is the state of firewalld implementation then? I guess this is already a problem with the current isolate=true/false behavior? I think this needs to be supported by firewalld if we ever want to enable firewalld. But given how long we have been pushing this I have doubts that this is going to happen soon.  

In any case I think this PR is good and solves a real problem and the firewalld future is something for us maintainers to figure out and should not block this PR from merging.

Luap99 (approved):
LGTM, thanks
/lgtm

In order to create networks with `podman network create --opt isolate=strict` we need to patch c/common to accept that new value.
Do you want to make a PR there to enable it? If not I can do it. 

ref: https://github.com/containers/common/blob/20def0054c6e1841e1b33fe4ecc45c0ab3c29e6f/libnetwork/netavark/config.go#L189-L195

yassi-github:
> LGTM, thanks /lgtm

Glad I could contribute!

> In order to create networks with `podman network create --opt isolate=strict` we need to patch c/common to accept that new value. Do you want to make a PR there to enable it? If not I can do it.
> 
> ref: https://github.com/containers/common/blob/20def0054c6e1841e1b33fe4ecc45c0ab3c29e6f/libnetwork/netavark/config.go#L189-L195

I'd like to try it.

PR #706, opened 2023-05-30
Document how to generate a code coverage report for netavark

jakecorrenti:
Documents how to generate a code coverage report using coverage-profiling in the Rust compiler and the `grcov` tool created by Mozilla. Generates an HTML version of the report that you can open in the browser and navigate.

Current status:

image


jakecorrenti:
@Luap99 PTAL

jakecorrenti:
> LGTM

Running into issues with reproducibility. Sometimes a correct report is generated (in my case), but with the same packages @Luap99 hasn't been able to generate the same report. Looking into it

jakecorrenti:
@Luap99 I made some super small changes, but I've been able to consistently reproduce a couple of times.

Luap99 (commented):
Sorry I still only get the build.rs coverage.

jakecorrenti:
The minor changes should be reproducible. Myself and Mohan were able to produce the same results.

baude:
@jakecorrenti mind rebasing and see if we can get this across the finish line ?  re-running the tests didnt work.

jakecorrenti:
looks like integration failed but integration_aarch64 is fine, maybe a restart would do the trick?

mheon:
Tests now green.
/lgtm

PR #714, opened 2023-06-06
dhcp implement release

Luap99:


src/dhcp_proxy/dhcp_service.rs
```diff
@@ -222,3 +99,33 @@ impl From<DhcpServiceError> for Status {
...
+                );
+                let lease = NetavarkLease::from(lease);
+                // get previous lease and check if ip addr changed, if not we do not have to do anything
+                if let Some(old_lease) = &client.previous_lease {
+                    if old_lease.yiaddr != lease.yiaddr
+                        || old_lease.gateways != lease.gateways
+                        || old_lease.subnet_mask != lease.subnet_mask
+                    {
+                        // ips do not match, remove old ones and assign new ones.
+                        log::error!("ips do not match, reassign not implemented")
```
Luap99, line 120: I still need to implement the actual ip/gateway changes here but the main work in order to actually request new leases is done and working. I am currently thinking about testing this and the problem is that dnsmasq sets 2 minutes as minimum lease time so if I write a bats test it would take a while but I see no way to avoid this.

@baude @mheon @flouthoc Any objections to having a signle test that takes over two minutes?

mheon: On the Podman repo I'd be a little more hesitant, but over here I have absolutely no problems with this

Luap99:
@baude @mheon @flouthoc PTAL
This is required to make dhcp actually work correctly for longer running containers.

flouthoc (approved):
LGTM

Luap99:
@mheon PTAL

PR #717, opened 2023-06-09
aardvark: fix missleading dns disabled log

Luap99:
We should only say dns is disabled because of the missing binary when dns was requested. If there are no dns entries we should not log this line. Also inlcude the path in the log so users know which path we want.


flouthoc (approved):
LGTM

PR #733, opened 2023-06-26
netavark: make --config required for dns

Luap99:
Using /tmp as default makes no sense, all callers that need dns should give us a proper path. Podman already does this so it should not cause any backwards compat problems.

I had to change the teardown logic a bit as we first need to confirm if we have network with dns before we should check if --config was given.

Also rework tests to always set --config to avoid code duplication.


Luap99:
reping, please let's get this and cut the release

flouthoc (approved):
LGTM

PR #738, opened 2023-06-28
[CI:BUILD] RPM: Fix ELN build and cleanup spec

lsm5:
- use go-md2man binary path as build dependency as package names are different on different envs.
- simplify copr conditional for rpm epoch values.


lsm5:
PTAL. good to go.

Luap99:
Is this required for netavark 1.7? The release commit is already merged so it will not be part of the tag!

lsm5:
> Is this required for netavark 1.7? The release commit is already merged so it will not be part of the tag!

Not a problem. If there's any issues noticed in koji, I can do a manual build.

lsm5:
There were some inconsistencies in Copr v/s Koji environments for ELN which I noticed only yesterday.

PR #740, opened 2023-07-03
iptables: improve error when ip6?tables commands are missing

martinetd:
Running podman on a system without iptables right now yields this error:
```
Error: netavark: No such file or directory (os error 2)
```

It's difficult to tell what is missing exactly; this commit makes the message print iptables or ip6tables before the ENOENT message so it should help diagnostics


flouthoc (approved):
LGTM

Luap99 (approved):
Thanks
/lgtm

PR #749, opened 2023-07-10
bump netlink deps

Luap99:


Luap99:
@mheon @baude @flouthoc PTAL

PR #751, opened 2023-07-14
[CI:BUILD] Packit: remove pre-sync action

lsm5:
It is currently not possible to reliably test the job-specific pre-sync action until the subsequent upstream release due to limitations in packit.

The lines being added by the action script to the downstream Fedora spec were only meant to keep Fedora happy. But given that they provide no tangible benefit as github notifies us of security issues in libraries, along with redhat prodsec's own magic for creating security alerts, there's absolutely no point to having the pre-sync action run and add a layer of uncertainty.

This commit removes the pre-sync action and
`rpm/update-spec-provides.sh`.

Ref: https://github.com/containers/podman/issues/19232

[NO NEW TESTS NEEDED]


Luap99 (approved):
/lgtm

PR #770, opened 2023-08-08
packit: Build PRs into default packit COPRs

martinpitt:
Building all PRs of all container projects into the same COPR does not properly isolate PRs from each other: E.g. a podman PR currently runs against whichever netavark PR was opened/updated last; in other words, sending a broken netavark PR will instantly break tests for all subsequent podman runs.

To avoid that, change the copr_build configuration to use the packit default COPRs, which are specific to the particular PR, and disappear after a few weeks. Depending projects like podman should only run against what landed in netavark/main i.e. the podman-next COPR.

Note that this does not preclude testing a podman PR against an netavark PR: This can be explicitly requested [1]. But most PRs don't change the API and thus should default to isolation.

[1] https://packit.dev/posts/testing-farm-triggering

----

This is exactly the same as https://github.com/containers/crun/pull/1260 ; please see the discussion there, this change needs to be applied to all container projects, and then all land together. Let's keep all the relevant discussions and tracking in  the crun PR please.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,31 +2,36 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
-# Build targets can be found at:
-# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
-# and
-# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/
-
 specfile_path: rpm/netavark.spec
 upstream_tag_template: v{version}
 
-jobs:
-  - &copr
-    job: copr_build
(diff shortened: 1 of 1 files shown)


Luap99:
@lsm5 PTAL

martinpitt:
[This "integration" test](https://github.com/containers/netavark/pull/770/checks?check_run_id=15704763169) failed with "Could not find a lease within the timeout limit". Retrying once to compare.

Luap99:
This is a know flake, feel free to ignore it

martinpitt:
Thanks @Luap99 ! This time it went green :green_circle:

PR #771, opened 2023-08-08
docs: Convert markdown with go-md2man instead of mandown

### LINKED ISSUE #768: Consider replacing mandown
The project appears to be unmaintained and doesn't ship a Cargo.lock, thus does not provide deterministic builds. An MR was opened to address this 3 months ago, but hasn't seen any activity from upstream, neither has the `master` branch of the repository seen any activity in over a year.

Fedora currently make use of `go-md2man` in their [spec file](https://src.fedoraproject.org/rpms/netavark/blob/f38/f/netavark.spec#_47), though I can only speculate as to their reasoning. Debian's [fork](https://salsa.debian.org/debian/netavark/-/tree/debian/sid/debian) for sid doesn't even seem to install man pages at this point, but I don't see them packaging `mandown` in it's current state.

benjamb:
Closes #768.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,16 +1,17 @@
 PREFIX ?= /usr/local
 DATADIR ?= ${PREFIX}/share
 MANDIR ?= $(DATADIR)/man
-MANDOWN ?= mandown
+GO ?= go
+GOMD2MAN ?= go-md2man
 
 docs: $(patsubst %.md,%,$(wildcard *.1.md))
 
 %.1: %.1.md
-	$(MANDOWN) $< > $@
+	$(GOMD2MAN) -in $^ -out $@
 
-.PHONY: .install.mandown
-.install.mandown:
-	cargo install mandown
+.PHONY: .install.md2man
+.install.md2man:
+	$(GO) install github.com/cpuguy83/go-md2man/v2@latest
 
 .PHONY: install
 install:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,7 +17,7 @@ env:
     # Save a little typing (path relative to $CIRRUS_WORKING_DIR)
     SCRIPT_BASE: "./contrib/cirrus"
     FEDORA_NAME: "fedora-38"
-    IMAGE_SUFFIX: "c20230807t144831z-f38f37d13"
+    IMAGE_SUFFIX: "c20230809t143240z-f38f37d13"
     FEDORA_NETAVARK_IMAGE: "fedora-netavark-${IMAGE_SUFFIX}"
     AARDVARK_DNS_BRANCH: "main"
     AARDVARK_DNS_URL: "https://api.cirrus-ci.com/v1/artifact/github/containers/aardvark-dns/success/binary.zip?branch=${AARDVARK_DNS_BRANCH}"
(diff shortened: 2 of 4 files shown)


docs/Makefile
```diff
@@ -1,16 +1,18 @@
...
 %.1: %.1.md
-	$(MANDOWN) $< > $@
+	$(GOMD2MAN) -in $^ -out $@
 
-.PHONY: .install.mandown
-.install.mandown:
-	cargo install mandown
+.PHONY: .install.md2man
+.install.md2man:
+	$(GO) install github.com/cpuguy83/go-md2man/v2@latest
```
benjamb, line 15: Wasn't sure whether to simply remove this and assume `go-md2man` was installed, or update this convenience target to install it. Happy to simply drop it if preferred.

cevich:
> @cevich would you be able to add the tool in question to the images used by validate?

Sure.

docs/Makefile
cevich: Doesn't this assume the caller has `$GOPATH/bin` in their `$PATH`?  Seeing as this is not a golang project, it might be nice to at least mention that in a comment.

benjamb: To an extent. If you look at [line 6](https://github.com/containers/netavark/pull/771/files#diff-cdec797dddf8fc387304501d6bd4f318e1e9067c6fae012dfa8b69fd85fd3248R6), you can see we use `command` to check for it in `$PATH`, falling back to `$(GOBIN)/go-md2man`. The line above sets a default for `$GOBIN` in case it's unset on the environment.

cevich: I don't see any default for `$GOPATH`, so `$GOBIN` will default to `/bin` which is probably wrong.

Luap99: I think we can just assume it to be in $PATH, on go dev envronment it is very likely that gobin is in $PATH anyway and for CI systems it should really be installed $PATH.

as for line 6 we can simply leave it as `go-md2man` like we did with mandown.

docs/Makefile
```diff
@@ -1,16 +1,18 @@
 PREFIX ?= /usr/local
 DATADIR ?= ${PREFIX}/share
 MANDIR ?= $(DATADIR)/man
-MANDOWN ?= mandown
+GO ?= go
+GOBIN ?= $(GOPATH)/bin
```
cevich, line 5:
```suggestion
GOBIN ?= $($(GO) env GOPATH)/bin
```

docs/Makefile
benjamb: @Luap99 Should I go ahead and remove `.install.md2man` then?

docs/Makefile
benjamb: I've dropped the setting of `$GOBIN` as per https://github.com/containers/netavark/pull/771/files#r1288638343.

Luap99 (approved):
LGTM

cevich:
*<BEEP>* new CI images are done: `c20230809t143240z-f38f37d13`.  Stick that in `.cirrus.yml` as the value for `IMAGE_SUFFIX` (near the top).  Commit and push.  If CI breaks and I don't notice, just poke me and I'll take a look.

benjamb:
Looks like we passed!

@cevich Thanks for the image!

Luap99 (approved):
/lgtm

Thanks for the PR

PR #774, opened 2023-08-11
iptables: drop invalid packages

Luap99:
As described in the bugzilla it is possible to leak the container source ip in a small race window where the connection is reset from both ends at the same time. To prevent leaking invalid packages we should just drop them. There is no benefit in sending such packages so we simply add a single rule to drop all invalid packages. I think this is better then adding a bunch of rules filtering on the container subnet.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2230144


Luap99:
@baude @mheon @flouthoc @cgiradkar PTAL

mheon:
LGTM

cgiradkar:
LGTM

PR #780, opened 2023-08-16
Add ACCEPT rule in firewall for bridge network with internal dns

cgiradkar:
When running container in non default podman network with dns enabled and default INPUT host firewall policy is DROP, name resolution does not work in container. To fix this, a firewall rule to allow dns traffic was added. 

Signed-off-by: Chetan Giradkar <cgiradka@redhat.com>

### DIFF
--- a/src/firewall/varktables/types.rs
+++ b/src/firewall/varktables/types.rs
@@ -15,7 +15,10 @@ const FILTER: &str = "filter";
 const POSTROUTING: &str = "POSTROUTING";
 const PREROUTING: &str = "PREROUTING";
 const NETAVARK_FORWARD: &str = "NETAVARK_FORWARD";
+const NETAVARK_FIREWALL_RULE_BUILDER: &str = "-m comment --comment 'netavark firewall rules' -j ";
+const NETAVARK_INPUT: &str = "NETAVARK_INPUT";
 const OUTPUT: &str = "OUTPUT";
+const INPUT: &str = "INPUT";
 const FORWARD: &str = "FORWARD";
 const ACCEPT: &str = "ACCEPT";
 const NETAVARK_HOSTPORT_DNAT: &str = "NETAVARK-HOSTPORT-DNAT";
@@ -196,6 +199,7 @@ pub fn get_network_chains<'a>(
     is_ipv6: bool,
     interface_name: String,
     isolation: IsolateOption,
+    dns_port: u16,
 ) -> Vec<VarkChain<'a>> {
     let mut chains = Vec::new();
     let prefixed_network_hash_name = format!("{}-{}", "NETAVARK", network_hash_name);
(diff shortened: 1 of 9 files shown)


Luap99 (commented):
This needs a test, take a look at `dual stack dns with alt port`.
You simple add the drop rules in the test with
```
run_in_host_netns iptables -P INPUT DROP
run_in_host_netns iptables -A INPUT -m state --state RELATED,ESTABLISHED -j ACCEPT
```
Then make sure dig works like done in the mentioned test.


In order for this to work it has to be part of the forward chain in the filter table. We currently do not have any rules there so it would be best to to create a new NETAVARK_INPUT chain for this so we do not clutter the main forward chain with all our rules.

Right now you set it in the nat table in prerouting so it will never match the packages we need.

src/firewall/varktables/types.rs
```diff
@@ -200,6 +201,8 @@ pub fn get_network_chains<'a>(
     let mut chains = Vec::new();
     let prefixed_network_hash_name = format!("{}-{}", "NETAVARK", network_hash_name);
 
+    let dns_port: i64 = 53;
```
Luap99, line 204: we should not hardocde that, you can add the dns_port as parameter to the function and the use that one instead as user could have changed it.

src/firewall/varktables/types.rs
```diff
@@ -234,7 +237,21 @@ pub fn get_network_chains<'a>(
...
+    // Always add ACCEPT rules in firewall for dns traffic from containers
+    // to gateway when using bridge network with internal dns.
+    input_chain.build_rule(VarkRule::new(
+        format!(
+            "-p {} -s {} --dport {} -j {}",
+            "udp", network, dns_port, ACCEPT
+        ),
+        Some(TeardownPolicy::OnComplete),
+    ));
+    chains.push(input_chain);
```
Luap99, line 254: we should create another chain in the INPUT chain called NETAVARK_INPUT, then put this rule into the NETVARK_INPUT chain. This makes sure to keep our rules better organised.

see the NETAVARK_FORWARD chain as an example

Luap99 (approved):
LGTM, don't forget to remove the draft status

@mheon @baude @flouthoc PTAL

src/firewall/varktables/types.rs
```diff
@@ -353,6 +361,33 @@ pub fn get_network_chains<'a>(
         VarkChain::new(conn, FILTER.to_string(), NETAVARK_FORWARD.to_string(), None);
     netavark_forward_chain.create = true;
 
+    // Add NETAVARK_INPUT chain to INPUT chain
+    input_chain.build_rule(VarkRule {
+        rule: format!(
+            "-m comment --comment 'netavark firewall plugin rules' -j {}",
```
flouthoc, line 367: This is the exact same string which is written here: https://github.com/containers/netavark/blob/main/src/firewall/varktables/types.rs#L343 , I think this should become a constant.

also should comment here just be called `netavark firewall rules` @Luap99 WDYT ?

flouthoc (approved):
One nit otherwise LGTM

cgiradkar:
@flouthoc can you have another look please?

cgiradkar:
@flouthoc can you reveiew and merge if everything apt. Thanks

flouthoc (approved):
LGTM @mheon PTAL

PR #807, opened 2023-09-13
rust io safety: convert RawFd to BorrowedFd<>

### LINKED ISSUE #806: io safety: convert RawFd to AsFd types
see https://rust-lang.github.io/rfcs/3128-io-safety.html for reasons

nix now accepts AsFd types so we can use the over the RawFd type, the current code is fine at the moment but easy to misuse because we can drop the file and still have the RawFd somewhere which would lead to weird bugs.

Luap99:
This is much safer because we no longer pass a single int around. With BorrowedFd<> the compiler will actually make sure the underlying file will not be closed before we use it.

I could have made every struct generic to accept the AsFd trait but that would have resulted in IMO more complex code without benefit.

Fixes #806

### DIFF
--- a/examples/host-device-plugin.rs
+++ b/examples/host-device-plugin.rs
@@ -3,6 +3,7 @@
 use std::{
     collections::HashMap,
     net::{Ipv4Addr, Ipv6Addr},
+    os::fd::AsFd,
 };
 
 use netavark::{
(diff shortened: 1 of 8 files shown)


flouthoc (approved):
LGTM

Luap99:
@baude @mheon PTAL

Issue #729, opened 2023-06-20
Add VRF support for bridge networks


MrSuicideParrot:
With native VRF support, we would be able to isolate traffic in a host that has multiple network interfaces without the need for complex firewall rules.

Previous discussion about this feature: https://github.com/containers/podman/discussions/18737

flouthoc:
I think this is implemented here: https://github.com/containers/netavark/pull/815 hence closing.

PR #815, opened 2023-09-24
Add vrf support for bridges

MrSuicideParrot:
This pull request adds support for adding a bridge created by netavark to a VRF (#729) .

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -19,7 +19,7 @@ use super::{
     constants::{
         ISOLATE_OPTION_FALSE, ISOLATE_OPTION_STRICT, ISOLATE_OPTION_TRUE,
         NO_CONTAINER_INTERFACE_ERROR, OPTION_ISOLATE, OPTION_METRIC, OPTION_MTU,
-        OPTION_NO_DEFAULT_ROUTE,
+        OPTION_NO_DEFAULT_ROUTE, OPTION_VRF,
     },
     core_utils::{self, get_ipam_addresses, join_netns, parse_option, CoreUtils},
     driver::{self, DriverInfo},
@@ -50,6 +50,8 @@ struct InternalData {
     metric: Option<u32>,
     /// if set, no default gateway will be added
     no_default_route: bool,
+    /// sef vrf for bridge
+    vrf: Option<String>,
     // TODO: add vlan
 }
 
@@ -81,6 +83,7 @@ impl driver::NetworkDriver for Bridge<'_> {
         let metric: u32 = parse_option(&self.info.network.options, OPTION_METRIC)?.unwrap_or(100);
         let no_default_route: bool =
             parse_option(&self.info.network.options, OPTION_NO_DEFAULT_ROUTE)?.unwrap_or(false);
+        let vrf: Option<String> = parse_option(&self.info.network.options, OPTION_VRF)?;
 
         let static_mac = match &self.info.per_network_opts.static_mac {
             Some(mac) => Some(CoreUtils::decode_address_from_hex(mac)?),
@@ -96,6 +99,7 @@ impl driver::NetworkDriver for Bridge<'_> {
             isolate,
             metric: Some(metric),
             no_default_route,
+            vrf,
         });
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


baude:
first glance, LGTM, thanks @MrSuicideParrot

src/network/bridge.rs
```diff
@@ -494,6 +498,17 @@ fn create_interfaces(
                     InfoKind::Bridge,
                 );
                 create_link_opts.mtu = data.mtu;
+
+                if data.vrf.is_some() {
+                    let vrf = match host.get_link(netlink::LinkID::Name(
+                        data.vrf.as_ref().unwrap().to_string(),
```
flouthoc, line 504: should this be handled `unwrap().to_string()` ? What would happen here is `unwrap()` fails ? There is no wrapping of error here.

src/network/netlink.rs
```diff
@@ -37,6 +37,7 @@ pub struct CreateLinkOptions<'fd> {
     pub link: u32,
     pub mac: Vec<u8>,
     pub netns: Option<BorrowedFd<'fd>>,
+    pub master: Option<u32>,
```
Luap99, line 40: this option already exists as `primary_index` so do not add this again

src/network/bridge.rs
Luap99: The best way to write this is `if let Some(vrf) = data.vrf {` then in the if block you can directly access the vrf var.

test/600-bridge-vrf.bats
```diff
@@ -0,0 +1,36 @@
...
+    run_in_host_netns ip link set dev test-vrf up
+
+    run_netavark --file ${TESTSDIR}/testfiles/simplebridge-vrf.json setup $(get_container_netns_path)
+    result="$output"
+    assert_json "$result" 'has("podman")' == "true" "object key exists"
+
+    mac=$(jq -r '.podman.interfaces.eth0.mac_address' <<<"$result")
+    # check that interface exists
+    run_in_container_netns ip -j --details link show eth0
+
```
Luap99, line 19: no need to check this

test/600-bridge-vrf.bats
```diff
@@ -0,0 +1,36 @@
...
+@test vrf - simple bridge {
+    run_netavark --file ${TESTSDIR}/testfiles/simplebridge.json setup $(get_container_netns_path)
+    run_in_host_netns ip -j --details link show podman0
+    result="$output"
+    assert_json "$result" ".[].linkinfo.info_slave_kind" "==" "null" "VRF is not set"
+}
+
+@test vrf - non existent vrf {
+    expected_rc=1 run_netavark --file ${TESTSDIR}/testfiles/simplebridge-vrf.json setup $(get_container_netns_path)
```
Luap99, line 34: IF you check for errors please also check for the errors message

test/600-bridge-vrf.bats
```diff
@@ -0,0 +1,36 @@
...
+    assert_json "$result" 'has("podman")' == "true" "object key exists"
+
+    mac=$(jq -r '.podman.interfaces.eth0.mac_address' <<<"$result")
+    # check that interface exists
+    run_in_container_netns ip -j --details link show eth0
+
+    # check if vrf exists
+    run_in_host_netns ip -j --details link show podman0
+    result="$output"
+    assert_json "$result" ".[].linkinfo.info_slave_kind" "==" "vrf" "Bridge has a vrf set"
```
Luap99, line 23: This should check the the vrf is in fact `test-vrf`

Luap99 (changes requested):

src/network/netlink.rs
MrSuicideParrot: I removed all the changes that I made to netlink.rs  and used instead  the `primary_index`.

test/600-bridge-vrf.bats
MrSuicideParrot: I added a check to verify if we receive the correct error message.

src/network/bridge.rs
MrSuicideParrot: Thank you very much for your input, I don't have much experience  with rust. I rewrote this part in my latest commit, could you check if it is better?

flouthoc (approved):
Thanks :) PR LGTM

Luap99 (approved):
LGTM

Luap99:
@MrSuicideParrot Thanks for the PR, FYI in order to add this option with `podman network create --opt` this needs to be wired into c/common https://github.com/containers/common/blob/e18cda8d7750746031de3d3d9305df059ac5f0ae/libnetwork/netavark/config.go#L174
And then a docs PR to podman to document this new option.

Let me know if you want to take care of that otherwise I will do it later.

PR #822, opened 2023-09-29
fix podman.io community link

Luap99:
This header no longer exists so we should not link to it, the top level page is just fine.


flouthoc:
LGTM
/lgtm
/approve

PR #833, opened 2023-10-30
bridge: force static mac on bridge interface

Luap99:
By default the kernel always uses the lowest mac address from all connected interfaces on the bridge. This means that while we start/stop containers the bridge can randomly change and that causes problems for existing ARP entries. ARP will still try the old mac and thus until the ARP entry is refreshed all packages are lost and for some reason that can take about 100s as seens in the linked issue.

To fix this we need to force a static mac on the bridge interface.

Fixes https://issues.redhat.com/browse/RHEL-14309


mheon:
Damn fine working tracking this one down

mheon:
Code LGTM, tests red

robbmanes:
Mind == Blown

...I banged my head against this for two days, and noticed the repeated MAC address assignments, but didn't think to correlate it to the lack of communication.  Excellent debugging and fix, thanks @Luap99.

src/network/bridge.rs
```diff
@@ -558,6 +577,7 @@ fn create_veth_pair<'fd>(
     netns: &mut netlink::Socket,
     data: &InternalData,
     primary_index: u32,
+    bridge_mac: Option<Vec<u8>>,
```
flouthoc, line 580: These arguments are too much as per clippy. Probably its time to move all these args to a struct.

Luap99: added `#[allow(clippy::too_many_arguments)]` I rather not rework most of this code here right now considering that we likely need to backport this.

PR #837, opened 2023-11-02
Update rust container build files

baude:
We need 1.67 and ubuntu:latest (which tracks LTS) still does not have it and there have been no updates to the image in 20 or more days.  Change from latest to rolling.

Update from fedora 38 to 39 and rename the Containerfile to exclude the fedora version.

when this PR merges, we need to update quay.io for the Fedora build.


Luap99 (approved):
LGTM

@flouthoc @mheon PTAL

PR #840, opened 2023-11-10
Add firewalld reload service

Luap99:
Add a new service command which listens on for the firewalld reload dbus event and then re-adds all firewall rules back.
The unit file is written so that it is only started and stopped with firewalld so it never runs unless needed. The new unit is called netavark-firewalld-reload.service.
Tests have been added to ensure it is working as expected.

This required more changes that I would have liked but it is important to only write the minimal firewall config to disk to not waste memory as the config dir is on tmpfs.

Fixes https://issues.redhat.com/browse/RHEL-3079

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,7 +191,7 @@ dependencies = [
  "cfg-if",
  "event-listener 3.0.0",
  "futures-lite",
- "rustix 0.38.14",
+ "rustix 0.38.21",
  "windows-sys",
 ]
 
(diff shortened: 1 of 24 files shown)


Luap99:
@mheon @baude @flouthoc PTAL

@mheon I expect this to cause quite a few conflicts with you nftables work. I am happy to rebase that for you because I think that may be faster as I know what changes I had to make.

baude:
LGTM

PR #844, opened 2023-11-13
[v1.7-rhel] bridge: force static mac on bridge interface

Luap99:
By default the kernel always uses the lowest mac address from all connected interfaces on the bridge. This means that while we start/stop containers the bridge can randomly change and that causes problems for existing ARP entries. ARP will still try the old mac and thus until the ARP entry is refreshed all packages are lost and for some reason that can take about 100s as seens in the linked issue.

To fix this we need to force a static mac on the bridge interface.

Fixes https://issues.redhat.com/browse/RHEL-14309

Backport for https://issues.redhat.com/browse/RHEL-16299 and https://issues.redhat.com/browse/RHEL-16300.


mheon:
LGTM

flouthoc (approved):
LGTM

PR #848, opened 2023-11-16
firewalld-reload: prevent race which could leak fw rules

Luap99:
There is a small possibility that the reload service could add fw rules after the container was teardown and thus leak them forever.

Consider this case if both processes run in parallel:

| netavark fw reload service  |  netavark teardown |
|-----------------------------|--------------------|
| read fw config              |                    |
|                             | remove fw config   |
|                             | remove fw rules    |
| add fw rules                |                    |

In order to prevent the race, this commit adds a lock file to ensure that it is not possible to add/remove fw configs while the reload service is doing it's thing.

### DIFF
--- a/src/firewall/state.rs
+++ b/src/firewall/state.rs
@@ -4,6 +4,7 @@ use std::{
     path::{Path, PathBuf},
 };
 
+use fs2::FileExt;
 use serde::de::DeserializeOwned;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


mheon:
Code LGTM

baude:
good work, nice and clean

/hold
/lgtm

PR #850, opened 2023-11-17
use OsString/Path over String for file paths

Luap99:
Rust String type has a pretty well documented difference to most string types in many other languages. A rust string must be valid utf-8. However linux itself does not care about the char encoding in paths, it is simply a non zero byte array. This means that technically netavark will not accept all valid paths. While it unlikely that anyone is using non utf-8 paths in the real world it is still a non great to deny that.

Rust actually offers the OsString/OsStr types that do not enforce the utf-8 requirement. And there is also PathBuf/Path types that futher abstract file paths. So to fix this just use these types instead.

See the added test which checks for it, without this patch we would see a `error: invalid UTF-8 was detected in one or more arguments` message.

### DIFF
--- a/src/commands/firewalld_reload.rs
+++ b/src/commands/firewalld_reload.rs
@@ -1,3 +1,8 @@
+use std::{
+    ffi::{OsStr, OsString},
+    path::Path,
+};
+
 use zbus::{blocking::Connection, dbus_proxy, CacheProperties};
 
 use crate::{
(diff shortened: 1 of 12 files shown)


Luap99:
@baude @mheon @flouthoc PTAL

mheon:
LGTM, but I quake in fear of the day I get a bug report with paths containing non-UTF8 characters.

Luap99:
Yeah I don't think it is likely that it will be hit but the nice thing about this is that the code does not have to validate if the args are valid utf8 so I guess there might be small speed up (likely not measurable though)

flouthoc (approved):
LGTM
/lgtm
/approve

PR #852, opened 2023-11-20
test: fix syntax problem in helpers.bash

Luap99:
firewalld_reload var needs to be defined.
This fixes the superfluous error message:
`helpers.bash: line 541: [: =: unary operator expected` which can be seen in the logs on failures.

Fixes b5d7510039 ("firewalld-reload: add integration tests")

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -407,6 +407,7 @@ function test_port_fw() {
     local container_port=""
     local range=1
     local connect_ip=""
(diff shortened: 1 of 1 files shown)


edsantiago:
LGTM. For a future maintenance PR, you might want to switch to using the double-bracket `[[ ... ]]` form. That does a better job of handling undefined/empty variables.

PR #856, opened 2023-11-30
aardvark: show error if process is in wrong netns

Luap99:
In the rootless netns case it is possible that aardvark-dns kept running
to an improper teardown (root cause unclear). Next time a new rootless
netns is created it means that aardvark-dns runs in the old wrong one
and no dns will work.

Technically it is possible for us to just kill it and start a new one
but if there are other old containers running based on that netns they
would loos dns as well. There is really no good way to fix this by
ourself as we do not know if other containers are still up and running.

So instead log a error with instructions on how to resolve the situation.

Fixes https://github.com/containers/podman/issues/20396

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -127,11 +127,42 @@ impl Aardvark {
         Ok(())
     }
 
+    fn check_netns(&self, pid: pid_t) -> Result<()> {
+        let cur_ns = fs::read_link("/proc/self/ns/net")?;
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

flouthoc (approved):
LGTM

Luap99:
Not needed IMO, this doesn't fix anything and just includes a error on how to recover. Besides this one user I have not seen other reports that would suggest that this is a more common error.

PR #862, opened 2023-12-06
ipam, network: implement `IPAM` management

flouthoc:
This PR intends to add IPAM Management code to netavark. Currently PR is in progress.

[NO NEW TESTS NEEDED]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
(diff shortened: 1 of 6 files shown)


flouthoc:
This PR has top level functions to create IPAM schema and make necessary queries needed for porting code from https://github.com/containers/common/blob/main/libnetwork/netavark/ipam.go to netavark.

Pending in this PR and still needs to be done.
* Ipam mod must port and implement logic from c/common and then remove logic/code from c/common. As a result IPAM management will be done by netavark itself and not podman.


Note: Netavark is already getting all the needed bits from podman in its config so till now I suspect no new config changes are needed for netavark end.

PS: Not working on this now but if possible I'd love to pick this up again in future and do it as per my convenience/bandwidth if this is not on high priority :smiley:

PR #865, opened 2023-12-07
netavark: dhcp_proxy: skip set gateway if missing

### LINKED ISSUE #861: macvlan crash when DHCP server doesn't send a gateway
I'm trying to use macvlan with DHCP, but my container will be on multiple networks and only one should provide a gateway.  If I remove `option routers 10.2.1.1` from my DHCP server so that no gateway is sent, podman crashes with:
```
Dec 05 00:16:32 fibrosis hass[436034]: thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', src/network/macvlan_dhcp.rs:67:38
Dec 05 00:16:32 fibrosis hass[436034]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

SilverBut:
Fixes #861

### DIFF
--- a/src/dhcp_proxy/dhcp_service.rs
+++ b/src/dhcp_proxy/dhcp_service.rs
@@ -68,19 +68,38 @@ impl DhcpV4Service {
     /// returns: Result<Lease, DhcpSearchError>. Either finds a lease successfully, finds no lease, or fails
     ///
     pub async fn get_lease(&mut self) -> Result<NetavarkLease, DhcpServiceError> {
-        if let Some(Ok(lease)) = self.client.next().await {
-            let mut netavark_lease = <NetavarkLease as From<MozimV4Lease>>::from(lease.clone());
-            netavark_lease.add_domain_name(&self.network_config.domain_name);
-            netavark_lease.add_mac_address(&self.network_config.container_mac_addr);
-
-            debug!(
-                "found a lease for {:?}, {:?}",
-                &self.network_config.container_mac_addr, &netavark_lease
-            );
-            self.previous_lease = Some(lease);
-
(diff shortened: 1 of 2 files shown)


src/dhcp_proxy/dhcp_service.rs
```diff
@@ -68,19 +68,29 @@ impl DhcpV4Service {
...
+                    return Ok(netavark_lease);
+                }
+                Err(err) => return Err(match err.kind() {
+                    mozim::ErrorKind::Timeout => DhcpServiceError::new(Timeout, err.to_string()),
+                    mozim::ErrorKind::InvalidArgument => DhcpServiceError::new(InvalidArgument, err.to_string()),
+                    mozim::ErrorKind::NoLease => DhcpServiceError::new(NoLease, err.to_string()),
+                    mozim::ErrorKind::Bug => DhcpServiceError::new(Bug, err.to_string()),
+                    _ => DhcpServiceError::new(Bug, err.to_string()),
+                })             
+            }
```
Luap99, line 91: This looks like a good change but properly worth to split in a separate commit as I do not see how this is related to the gw change.

Luap99 (commented):
Thanks for the fix, please make sure `make validate` passes and sign your commit, see https://github.com/containers/podman/blob/main/CONTRIBUTING.md#sign-your-prs

src/dhcp_proxy/dhcp_service.rs
SilverBut: Oh that was a mistake. Got it split into [4759152](https://github.com/containers/netavark/pull/865/commits/4759152125bc400bf970cd5c7a048b1c4639cc06).

PR #869, opened 2023-12-07
bridge: support DHCP ipam driver

### LINKED ISSUE #868: bridge doesn't work with DHCP
With podman version 4.7.2, netavark version 1.8.0:

`macvlan` driver works with DHCP:
```
# podman network create --disable-dns --driver macvlan --interface-name brint --ipam-driver dhcp lan
lan
# podman run -it --rm --network=lan alpine ip addr show eth0
2: eth0@if8: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP qlen 1000
    link/ether 4e:81:42:14:17:be brd ff:ff:ff:ff:ff:ff
    inet 10.0.103.174/16 brd 10.0.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::4c81:42ff:fe14:17be/64 scope link tentative 
       valid_lft forever preferred_lft forever
```
(note assigned 10.0.103.174 address there)

But `bridge` does not:
```
# podman network rm lan
# podman network create --disable-dns --driver bridge --interface-name brint --ipam-driver dhcp lan
lan
# podman run -it --rm --network=lan alpine ip addr show eth0
2: eth0@if28: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1500 qdisc noqueue state UP qlen 1000
    link/ether c6:9b:3d:ec:27:6c brd ff:ff:ff:ff:ff:ff
    inet6 fe80::c49b:3dff:feec:276c/64 scope link tentative 
       valid_lft forever preferred_lft forever
```

Is this configuration supportable?

(For my setup, I need `bridge` instead of `macvlan` so that the host can communicate with the container too.  I can work around it with `--cap-add NET_ADMIN,NET_RAW` and running `udhcpc` inside the container, but would prefer to avoid that if possible.)

jimparis:
Fixes #868

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -6,6 +6,7 @@ use netlink_packet_route::link::{
     InfoData, InfoKind, InfoVeth, LinkAttribute, LinkInfo, LinkMessage,
 };
 
+use crate::network::macvlan_dhcp::get_dhcp_lease;
 use crate::{
     dns::aardvark::AardvarkEntry,
     error::{ErrorWrap, NetavarkError, NetavarkErrorList, NetavarkResult},
@@ -150,9 +151,31 @@ impl driver::NetworkDriver for Bridge<'_> {
         // interfaces map, but we only ever expect one, for response
         let mut interfaces: HashMap<String, types::NetInterface> = HashMap::new();
 
+        // if dhcp is enabled, we need to call the dhcp proxy to perform
+        // a dhcp lease.  it will also perform the IP address assignment
+        // to the container interface.
+        let subnets = if data.ipam.dhcp_enabled {
+            let (subnets, dns_servers, domain_name) = get_dhcp_lease(
+                &data.bridge_interface_name,
+                &data.container_interface_name,
+                self.info.netns_path,
+                &container_veth_mac,
+            )?;
+            // do not overwrite dns servers set by dns podman flag
+            if !self.info.container_dns_servers.is_some() {
+                response.dns_server_ips = dns_servers;
+            }
+            if domain_name.is_some() {
+                response.dns_search_domains = domain_name;
+            }
+            subnets
+        } else {
+            data.ipam.net_addresses.clone()
+        };
+
         let interface = types::NetInterface {
             mac_address: container_veth_mac,
-            subnets: Option::from(data.ipam.net_addresses.clone()),
+            subnets: Option::from(subnets),
         };
         // Add interface to interfaces (part of StatusBlock)
         interfaces.insert(data.container_interface_name.clone(), interface);
@@ -250,6 +273,8 @@ impl driver::NetworkDriver for Bridge<'_> {
         let (host_sock, netns_sock) = netlink_sockets;
 
         let mut error_list = NetavarkErrorList::new();
+
+        core_utils::dhcp_teardown(&self.info, netns_sock)?;
 
         let routes = core_utils::create_route_list(&self.info.network.routes)?;
         for route in routes.iter() {
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -1,4 +1,5 @@
 use crate::error::{ErrorWrap, NetavarkError, NetavarkResult};
+use crate::network::macvlan_dhcp::release_dhcp_lease;
 use crate::network::{constants, internal_types, types};
 use crate::wrap;
 use ipnet::IpNet;
@@ -18,7 +19,9 @@ use std::os::unix::prelude::*;
 use std::str::FromStr;
 use sysctl::{Sysctl, SysctlError};
 
-use super::netlink;
+use super::{driver::DriverInfo, netlink};
+
+use netlink_packet_route::link::LinkAttribute;
 
 pub struct CoreUtils {
     pub networkns: String,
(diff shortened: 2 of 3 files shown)


Luap99 (commented):
Thanks for the contribution. The reason we do not support it right now is because we we do not really support layer 2 bridge setup. Our current code always assumes that we created the bridge adn we will also remove it once all interfaces connected to the bridge are removed.
I assume in order to use it you have your external interface connected to the bridge?

Also this assumption means we always create firewall rules to NAT the traffic which does not make sense for these setups usually as you want to talk directly to your gateway without having NAT involved. So I think we would need add proper layer two bridge support first before enabling DHCP for bridge networks.

---
As far as these changes the same code should not be duplicated for bridge, I suggest to split the common parts out into helper functions.

src/network/bridge.rs
```diff
@@ -69,6 +70,17 @@ impl<'a> Bridge<'a> {
     }
 }
 
+fn get_mac_address(v: Vec<Nla>) -> NetavarkResult<String> {
```
Luap99, line 73: move this function to core_utils.rs then make all coders import it and use it from there

src/network/bridge.rs
```diff
@@ -251,6 +277,33 @@ impl driver::NetworkDriver for Bridge<'_> {
...
+        let ipam = get_ipam_addresses(self.info.per_network_opts, self.info.network)?;
+        let if_name = self.info.per_network_opts.interface_name.clone();
+
+        // If we are using DHCP, we need to at least call to the proxy so that
+        // the proxy's cache can get updated and the current lease can be released.
+        if ipam.dhcp_enabled {
+            let dev = netns_sock
+                .get_link(netlink::LinkID::Name(if_name))
+                .wrap(format!(
+                    "get bridge interface {}",
```
Luap99, line 289: this is the container interface not the bridge, in fact the current error message for macvlan is also wrong

src/network/bridge.rs
```diff
@@ -251,6 +277,33 @@ impl driver::NetworkDriver for Bridge<'_> {
...
+                    .info
+                    .network
+                    .network_interface
+                    .clone()
+                    .unwrap_or_default(),
+                &self.info.per_network_opts.interface_name,
+                self.info.netns_path,
+                &container_mac_address,
+            )?
+        }
```
Luap99, line 305: this block should be moved to a function in core_utils.rs so both bridge and maclvna code can call it and not have to duplicate it.

baude:
iill approve as we wait for the comments to be addressed ...

/approve

jimparis:
Addressed the review comments

ccakes:
Just in case in helps, I'm trying to get the exact same use case working and this would solve it :+1:

jimparis:
> Just in case in helps, I'm trying to get the exact same use case working and this would solve it 👍

Thanks. Have you been able to test this?  It's been working fine for me and should hopefully be ready to get merged.

Luap99 (commented):
I think functionally this is fine, however I still have concerns that this is working based on implementation detail; rather than a well defined behaviour.

I think we should expose this as some specific layer2 setup instead, the reason is due our firewall setup and bridge removal. The firewall setup works today with iptables as it doesn't do anything if finds no subnet. However we also have a nftables and firewalld driver that might something in this case.

@mheon WDYT?

Luap99:
> All of our firewall code should be IP based (I don't think we can reasonably do NAT or other firewall operations without a source subnet). So I think it is a reasonable assumption that any firewall driver we include should only function in the case that a source subnet (v4 or v6, doesn't really matter, as long as there is at least 1) is set on the bridge. Anything lower-level than that isn't really a firewall anymore, but would arguably be an implementation detail of the lower-level bridge driver.

Yes the code is ip cased but it will still start to create to tables/chains that then are left empty due to no ips. I rather have a well defined behavior where we do not call into the firewall code at all in this case here so we know we won't mess anything up.

mheon:
Yeah, that's reasonable. There's no reason for the firewall code to be called at all if the interface we're configuring doesn't have an IP.

AkiraMoon305 (approved):

jimparis:
Rebased and re-tested on main.  Where do we stand on this PR?  It's still a feature I use and need for my setup.

WintersMichael:
I created https://github.com/containers/netavark/issues/1090 to address the underlying concerns and preferred approach, since those aren't DHCP-specific.

mikesbytes:
Hello, just throwing my hat in the ring for this being a desirable feature. This is pretty critical for achieving my desired setup of hosting podman containers alongside nspawn containers and VMs and reverse proxying them all at the host level while also maintaining direct connectivity inside LAN.

I see that #1090 is now merged, and this PR hasn't seen activity for a couple months. What's the status? I would offer to contribute myself but it looks like things are already basically at the finish line.

xhebox:
A usecase report from me.

I am using openwrt as a podman host for hosting apps. I'd like to let openwrt created an interface `podman0` and took over the management for better integration.

But this patch did not work as expected.

Because sending DHCP requests from `podman0` does not seem broadcast it correctly to `dnsmasq`. You can use `udhcpc` inside the container to get an ip address. But not with `dhcp-proxy`.

I have to setup another `veth` pair with one slave of `podman0` bridge. And let `mozim` dhcp requests on this interface instead of the bridge interface, to really forward BOOTP/DHCP requests.

jimparis:
> Yes this should be good to rebase and then add a check that DHCP is only supported with the unmanaged mode.

Sorry, I don't fully understand what #1090 changed and what new check this PR needs.  Any chance you could handle updating this PR with what you'd like to see?

xhebox:
> > Yes this should be good to rebase and then add a check that DHCP is only supported with the unmanaged mode.
> 
> Sorry, I don't fully understand what #1090 changed and what new check this PR needs. Any chance you could handle updating this PR with what you'd like to see?

I believe he means that `dhcp ipam driver for bridge`(this PR) can only be used when `{"options":{"mode":"unmanaged"}}`(#1090) is specified in the network definition.  In other words, guard the code of this PR by some if statements `BridgeMode::unmanaged == data.mode`.

#1090 basically assume that bridge is created externally and not managed by netavark, same to firewall rules, sysctl settings. netavark just pick up a bridge and attach veth pairs.

Luap99:
Yes I can take care of the rebase/update before we do the next release. My goal is to get this done for the 1.14 release which will be released in February most likely.

Luap99:
Replaced with https://github.com/containers/netavark/pull/1159

PR #873, opened 2023-12-14
Use tonic::transport::Uri instead of HTTP

baude:
With http-1.0.0, the use of HTTP::Uri is no longer valid for when a tonic::transport::Uri is required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1346,7 +1346,6 @@ dependencies = [
  "futures-channel",
  "futures-core",
  "futures-util",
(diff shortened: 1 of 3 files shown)


baude:
this replaces #847

Luap99 (changes requested):
This works but you just dropped the only import were http was used so the better solution just drop http as dependency and let our direct dependencies such as tonic handle the version of http they are using.

Luap99 (approved):
LGTM

PR #874, opened 2023-12-15
netavark: dhcp_proxy: use dns servers from dhcp lease

### LINKED ISSUE #866: DHCP proxy doesn't set advertised DNS servers
As per the current implementation, it seems that the nameservers in the container's `/etc/resolv.conf` are copied from the host's `/etc/resolv.conf`. It would be nice if `netavark-dhcp-proxy` somehow assigned to each container the DNS servers provided by their respective DHCP requests.

This would play really nice with scenarios where the host is in a different VLAN+subnet (e.g. `management` VLAN) than the containers (e.g. `services` VLAN) and each VLAN has its own DHCP server with different advertised DNS server.

agorgl:
Fixes #866

### DIFF
--- a/src/network/macvlan_dhcp.rs
+++ b/src/network/macvlan_dhcp.rs
@@ -6,6 +6,9 @@ use std::str::FromStr;
 
 use crate::dhcp_proxy::lib::g_rpc::NetworkConfig;
 use crate::dhcp_proxy::proxy_conf::DEFAULT_UDS_PATH;
+
+pub type DhcpLeaseInfo = (Vec<NetAddress>, Option<Vec<IpAddr>>, Option<Vec<String>>);
+
 /// dhcp performs the connection to the nv-proxy over grpc where it
 /// requests it to perform a lease via the host's network interface
 /// but passes it the network interface from the container netns.:w
@@ -30,7 +33,7 @@ pub fn get_dhcp_lease(
     container_network_interface: &str,
     ns_path: &str,
     container_macvlan_mac: &str,
-) -> NetavarkResult<Vec<NetAddress>> {
+) -> NetavarkResult<DhcpLeaseInfo> {
     let nvp_config = NetworkConfig {
         host_iface: host_network_interface.to_string(),
         // TODO add in domain name support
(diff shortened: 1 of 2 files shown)


src/network/macvlan_dhcp.rs
```diff
@@ -75,6 +75,25 @@ pub fn get_dhcp_lease(
...
+                        return Err(NetavarkError::msg(format!("bad dns address: {e}")));
+                    }
+                }
+            })
+            .collect::<Result<Vec<IpAddr>, NetavarkError>>()?;
+        Some(servers)
+    } else {
+        None
+    };
+    let domain_name = Some(lease.domain_name).filter(|s| !s.is_empty());
```
Luap99, line 95: It seems better to do something like this
```
let domain_name = if !lease.domain_name.is_empty() {
        Some(vec![lease.domain_name])
    } else {
        None
    };
```
Then you do not need to convert in the caller side again.

src/network/vlan.rs
```diff
@@ -181,12 +181,19 @@ impl driver::NetworkDriver for Vlan<'_> {
...
+            let (subnets, dns_servers, domain_name) = get_dhcp_lease(
                 &data.host_interface_name,
                 &data.container_interface_name,
                 self.info.netns_path,
                 &container_vlan_mac,
-            )?
+            )?;
+            if dns_servers.is_some() {
+                response.dns_server_ips = dns_servers;
+            }
```
Luap99, line 192: you do not need to check dns_servers.is_some() but you check that self.info.container_dns_servers is empty because these are set via podman run --dns and dhcp should not overwrite them.

Luap99 (commented):
Thanks for the contribution, you have to add a sign off to your commit, see: https://github.com/containers/podman/blob/main/CONTRIBUTING.md#sign-your-prs

agorgl:
Fixed as you proposed, thanks for the suggestions!

agorgl:
@Luap99 this should be good to go

src/network/macvlan_dhcp.rs
```diff
@@ -6,6 +6,9 @@ use std::str::FromStr;
 
 use crate::dhcp_proxy::lib::g_rpc::NetworkConfig;
 use crate::dhcp_proxy::proxy_conf::DEFAULT_UDS_PATH;
+
+pub type DhcpLeaseInfo = (Vec<NetAddress>, Option<Vec<IpAddr>>, Option<Vec<String>>);
```
Luap99, line 10: it may be better to define a struct so that the fields have names so it is easier for the reader to understand

But this is just a minor nit not blocking and I am fine to merge it like this.

Luap99 (approved):
LGTM
@mheon @baude PTAL

@agorgl Did you test this with podman if it adds the correct resolv.conf entry for the container? I guess it just works but haven't tested it.

agorgl:
I think I tested it back in December (but it did not merge then because I didn't fix the linter checks failing)

PR #883, opened 2024-01-04
Initial draft of nftables support

### LINKED ISSUE #816: Nftables support
As of now, [the nftables backend is not implemented](https://github.com/containers/netavark/blob/3806d9a97dc912ae30827621abe78676802710ff/src/firewall/mod.rs#L103).

Hence, it is not possible to run netavark on a Linux kernel with `nftables` modules and without `iptables`.

Altough the `CONFIG_NFT_MASQ` kernel config is enabled, the `/usr/sbin/iptables-nft` alternative iptables CLI cannot emulate the masquerade action and complains:
```
Error: netavark: unable to append rule '! -d 224.0.0.0/4 -j MASQUERADE' to table 'nat': code: 4, msg: Warning: Extension MASQUERADE revision 0 not supported, missing kernel module?
iptables v1.8.9 (nf_tables):  RULE_APPEND failed (No such file or directory): rule in chain NETAVARK-1700E948C4CCE
```

Supposedly, native nftables driver should fix the issue.

mheon:
This implementation lacks isolation support at present (that'll be a followon PR) and has only very minimal testing at the moment (including absolutely 0 testing for IPv6), but does handle all core tasks including forwarding and all types of port forwarding.

Fixes #816

### DIFF
--- a/src/firewall/nft.rs
+++ b/src/firewall/nft.rs
@@ -1,0 +1,1122 @@
+use crate::error::{NetavarkError, NetavarkResult};
+use crate::firewall;
+use crate::firewall::firewalld;
+use crate::network::internal_types;
+use crate::network::types::PortMapping;
+use ipnet::IpNet;
+use nftables::batch::Batch;
+use nftables::expr;
+use nftables::helper::{self};
+use nftables::schema;
+use nftables::stmt;
+use nftables::types;
+use std::collections::HashSet;
+use std::net::IpAddr;
+
+const TABLENAME: &str = "netavark";
+
+const INPUTCHAIN: &str = "INPUT";
+const FORWARDCHAIN: &str = "FORWARD";
+const POSTROUTINGCHAIN: &str = "POSTROUTING";
+const PREROUTINGCHAIN: &str = "PREROUTING";
+const OUTPUTCHAIN: &str = "OUTPUT";
+const DNATCHAIN: &str = "NETAVARK-HOSTPORT-DNAT";
+const MASKCHAIN: &str = "NETAVARK-HOSTPORT-SETMARK";
+
+const MASK: u32 = 0x2000;
+
+/// The dnat priority for chains
+/// This (and the below) are based on https://wiki.nftables.org/wiki-nftables/index.php/Netfilter_hooks#Priority_within_hook
+const DNATPRIO: i32 = -100;
+/// The srcnat priority for chains
+const SRCNATPRIO: i32 = 100;
+/// The filter priority for chains
+const FILTERPRIO: i32 = 0;
+
+pub struct Nftables {}
+
+pub fn new() -> Result<Box<dyn firewall::FirewallDriver>, NetavarkError> {
+    Ok(Box::new(Nftables {}))
+}
+
+impl firewall::FirewallDriver for Nftables {
+    fn driver_name(&self) -> &str {
+        firewall::NFTABLES
+    }
+
+    fn setup_network(&self, network_setup: internal_types::SetupNetwork) -> NetavarkResult<()> {
+        let mut batch = Batch::new();
+
+        // Overall table
+        batch.add(schema::NfListObject::Table(schema::Table::new(
+            types::NfFamily::INet,
+            TABLENAME.to_string(),
+        )));
+
+        // Five default chains, one for each hook we have to monitor
+        batch.add(make_complex_chain(
+            INPUTCHAIN,
+            types::NfChainType::Filter,
+            types::NfHook::Input,
+            FILTERPRIO,
+        ));
+        batch.add(make_complex_chain(
+            FORWARDCHAIN,
+            types::NfChainType::Filter,
+            types::NfHook::Forward,
+            FILTERPRIO,
+        ));
+        batch.add(make_complex_chain(
+            POSTROUTINGCHAIN,
+            types::NfChainType::NAT,
+            types::NfHook::Postrouting,
+            SRCNATPRIO,
+        ));
+        batch.add(make_complex_chain(
+            PREROUTINGCHAIN,
+            types::NfChainType::NAT,
+            types::NfHook::Prerouting,
+            DNATPRIO,
+        ));
+        batch.add(make_complex_chain(
+            OUTPUTCHAIN,
+            types::NfChainType::NAT,
+            types::NfHook::Output,
+            DNATPRIO,
+        ));
+
+        // dnat rules. Not used here, but need to be created first, because they have rules that must be first in their chains.
+        // A lot of these are thus conditional on if the rule already exists or not.
+        let existing_rules = helper::get_current_ruleset(None, None)?;
+
+        // Two extra chains, not hooked to anything, for our NAT pf rules
+        batch.add(make_basic_chain(DNATCHAIN));
+        batch.add(make_basic_chain(MASKCHAIN));
+
+        // Postrouting chain needs a single rule to masquerade if mask is set.
+        // But only one copy of that rule. So check if such a rule exists.
+        let match_meta_masq = |r: &schema::Rule| -> bool {
+            // Match on any rule that matches against 0x2000
+            for statement in &r.expr {
+                match statement {
+                    stmt::Statement::Match(m) => match &m.right {
+                        expr::Expression::Number(n) => {
+                            if *n == MASK {
+                                return true;
+                            }
+                        }
+                        _ => continue,
+                    },
+                    _ => continue,
+                }
+            }
+            false
+        };
+        if get_matching_rules_in_chain(&existing_rules, POSTROUTINGCHAIN, match_meta_masq)
+            .is_empty()
+        {
+            // Postrouting: meta mark & 0x2000 == 0x2000 masquerade
+            batch.add(make_rule(
+                POSTROUTINGCHAIN,
+                vec![
+                    stmt::Statement::Match(stmt::Match {
+                        left: expr::Expression::BinaryOperation(expr::BinaryOperation::AND(
+                            Box::new(expr::Expression::Named(expr::NamedExpression::Meta(
+                                expr::Meta {
+                                    key: expr::MetaKey::Mark,
+                                },
+                            ))),
+                            Box::new(expr::Expression::Number(MASK)),
+                        )),
+                        right: expr::Expression::Number(MASK),
+                        op: stmt::Operator::EQ,
+                    }),
+                    stmt::Statement::Masquerade(None),
+                ],
+            ));
+        }
+
+        // Mask chain needs a single rule to apply the mask.
+        // But only one copy of that rule. So check if such a rule exists.
+        let match_meta_mark = |r: &schema::Rule| -> bool {
+            // Match on any mangle rule.
+            for statement in &r.expr {
+                match statement {
+                    stmt::Statement::Mangle(_) => return true,
+                    _ => continue,
+                }
+            }
+            false
+        };
+        if get_matching_rules_in_chain(&existing_rules, MASKCHAIN, match_meta_mark).is_empty() {
+            // Mask chain: mark or 0x2000
+            batch.add(make_rule(
+                MASKCHAIN,
+                vec![stmt::Statement::Mangle(stmt::Mangle {
+                    key: expr::Expression::Named(expr::NamedExpression::Meta(expr::Meta {
+                        key: expr::MetaKey::Mark,
+                    })),
+                    value: expr::Expression::BinaryOperation(expr::BinaryOperation::OR(
+                        Box::new(expr::Expression::Named(expr::NamedExpression::Meta(
+                            expr::Meta {
+                                key: expr::MetaKey::Mark,
+                            },
+                        ))),
+                        Box::new(expr::Expression::Number(MASK)),
+                    )),
+                })],
+            ));
+        }
+
+        // We need rules in Prerouting and Output pointing to our dnat chain.
+        // But only if they do not exist.
+        let match_jump_dnat = get_rule_matcher_jump_to(DNATCHAIN.to_string());
+        // Prerouting: fib daddr type local jump <dnat_chain>
+        // Output: fib daddr type local jump <dnat_chain>
+        let mut rules_hash: HashSet<expr::FibFlag> = HashSet::new();
+        rules_hash.insert(expr::FibFlag::Daddr);
+        let base_conditions: Vec<stmt::Statement> = vec![
+            stmt::Statement::Match(stmt::Match {
+                left: expr::Expression::Named(expr::NamedExpression::Fib(expr::Fib {
+                    result: expr::FibResult::Type,
+                    flags: rules_hash,
+                })),
+                right: expr::Expression::String("local".to_string()),
+                op: stmt::Operator::EQ,
+            }),
+            get_jump_action(DNATCHAIN),
+        ];
+        if get_matching_rules_in_chain(&existing_rules, PREROUTINGCHAIN, &match_jump_dnat)
+            .is_empty()
+        {
+            batch.add(make_rule(PREROUTINGCHAIN, base_conditions.clone()));
+        }
+        if get_matching_rules_in_chain(&existing_rules, OUTPUTCHAIN, &match_jump_dnat).is_empty() {
+            batch.add(make_rule(OUTPUTCHAIN, base_conditions.clone()));
+        }
+
+        // Forward chain: ct state invalid drop
+        let match_deny = |r: &schema::Rule| -> bool {
+            for statement in &r.expr {
+                match statement {
+                    stmt::Statement::Drop(_) => return true,
+                    _ => continue,
+                }
+            }
+            false
+        };
+        if get_matching_rules_in_chain(&existing_rules, FORWARDCHAIN, match_deny).is_empty() {
+            batch.add(make_rule(
+                FORWARDCHAIN,
+                vec![
+                    stmt::Statement::Match(stmt::Match {
+                        left: expr::Expression::Named(expr::NamedExpression::CT(expr::CT {
+                            key: "state".to_string(),
+                            family: None,
+                            dir: None,
+                        })),
+                        right: expr::Expression::String("invalid".to_string()),
+                        op: stmt::Operator::IN,
+                    }),
+                    stmt::Statement::Drop(None),
+                ],
+            ));
+        }
+
+        // Basic forwarding for all subnets
+        if let Some(nets) = network_setup.subnets {
+            for subnet in nets {
+                let chain = get_subnet_chain_name(subnet, &network_setup.network_id, false);
+
+                // Add us to firewalld if necessary.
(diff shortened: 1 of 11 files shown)


mheon:
Stripping WIP. Passes testing I've thrown at it. Now needs integration tests.

src/firewall/nft.rs
```diff
@@ -0,0 +1,1078 @@
...
+        let rules = batch.to_nftables();
+
+        helper::apply_ruleset(&rules, None, None)?;
+
+        Ok(())
+    }
+}
+
+/// Convert a subnet into a chain name.
+fn get_subnet_chain_name(subnet: IpNet, dnat: bool) -> String {
```
Luap99, line 712: I am not a fan of using the subnet as chain name. What is wrong with the existing hash the used in iptables?
We could also just use the network ID I think?
Is there a reason to have one chain per subnet vs chain per network? One per subnet looks incorrect to me, because nft supports both ipv4 and ipv6 at the same time I would expect one chain that matches both the v4 and v6 subnet in there. More chains should also work but doesn't this just make things more complicated?

src/firewall/nft.rs
```diff
@@ -0,0 +1,1078 @@
...
+                                expr::Meta {
+                                    key: expr::MetaKey::Mark,
+                                },
+                            ))),
+                            Box::new(expr::Expression::Number(MASK)),
+                        )),
+                        right: expr::Expression::Number(MASK),
+                        op: stmt::Operator::EQ,
+                    }),
+                    stmt::Statement::Masquerade(None),
```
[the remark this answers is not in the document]
mheon: It is atrocious. But somehow better to use than the string API.

src/firewall/nft.rs
mheon: I like using subnets because it reads easier in nftables output (very obvious what each chain does), is completely unambiguous (no chance of a hash collision).

I could definitely combine v4 and v6 subnets into a single chain, but I think separating them is a bit more logical. It doesn't really need more rules, either (we need the same number of rules to actually forward traffic, and separate rules to forward traffic to the chain for v4 and v6 subnets).

src/firewall/nft.rs
```diff
@@ -0,0 +1,1078 @@
...
+    fn setup_network(&self, network_setup: internal_types::SetupNetwork) -> NetavarkResult<()> {
+        let mut batch = Batch::new();
+
+        // Overall table
+        batch.add(schema::NfListObject::Table(schema::Table::new(
+            types::NfFamily::INet,
+            TABLENAME.to_string(),
+        )));
+
+        // Five default chains, one for each hook we have to monitor
```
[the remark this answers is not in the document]
mheon: Sure, can do

src/firewall/nft.rs
Luap99: Yeah I need to dump me a `nft list ruleset` output to see what it actually looks like to better judge what I prefer.

mheon:
@Luap99 One last failing test, and it's a firewalld reload one. Any idea what might be going on?

Luap99:
Looks like you do not call into firewalld again to set the subnet trusted. (Do we even have to do that when using nftables?)
The VM is using f39 which has firewalld 2.0 which no longer flushes our custom fw rules by default, looking at your code you check that the nft rule exists then skip (continue) the subnet loop so the firewalld call is never done thus it is not trusted.

But because `nftables - port forwarding ipv4 - tcp with firewalld reload` passes I wonder if we even need to be trusted with nftables?

mheon:
The firewalld folks said it was safer to do so. I'll move the firewalld make-trusted call to the top of the loop.

mheon:
Alright, that should get tests passing. This is ready for review.

Luap99:
FYI CI is stucked likely due github outage.

Either way you can remove some of the clone() calls:

```diff
diff --git a/src/firewall/nft.rs b/src/firewall/nft.rs
index e362a0c..0b6e843 100644
--- a/src/firewall/nft.rs
+++ b/src/firewall/nft.rs
@@ -346,7 +346,7 @@ impl firewall::FirewallDriver for Nftables {
                         match statement {
                             stmt::Statement::Match(m) => match &m.right {
                                 expr::Expression::Named(expr::NamedExpression::Prefix(p)) => {
-                                    match *p.addr.clone() {
+                                    match p.addr.as_ref() {
                                         expr::Expression::String(s) => {
                                             if *s == subnet.addr().to_string() {
                                                 return true;
@@ -560,16 +560,16 @@ impl firewall::FirewallDriver for Nftables {
                                 _ => continue,
                             },
                             stmt::Statement::Jump(j) => {
-                                if let Some(v4) = dnat_chain_v4.clone() {
-                                    if j.target == v4 {
+                                if let Some(v4) = &dnat_chain_v4 {
+                                    if &j.target == v4 {
                                         if match_port {
                                             return true;
                                         }
                                         match_jump = true;
                                     }
                                 }
-                                if let Some(v6) = dnat_chain_v6.clone() {
-                                    if j.target == v6 {
+                                if let Some(v6) = &dnat_chain_v6 {
+                                    if &j.target == v6 {
                                         if match_port {
                                             return true;
                                         }
@@ -636,16 +636,16 @@ impl firewall::FirewallDriver for Nftables {
                     batch.delete(schema::NfListObject::Rule(rule));
                 }
 
-                if let Some(v4) = dnat_chain_v4.clone() {
+                if let Some(v4) = &dnat_chain_v4 {
                     for rule in
-                        get_matching_rules_in_chain(&existing_rules, &v4, match_all_ports_in_range)
+                        get_matching_rules_in_chain(&existing_rules, v4, match_all_ports_in_range)
                     {
                         batch.delete(schema::NfListObject::Rule(rule));
                     }
                 }
-                if let Some(v6) = dnat_chain_v6.clone() {
+                if let Some(v6) = &dnat_chain_v6 {
                     for rule in
-                        get_matching_rules_in_chain(&existing_rules, &v6, match_all_ports_in_range)
+                        get_matching_rules_in_chain(&existing_rules, v6, match_all_ports_in_range)
                     {
                         batch.delete(schema::NfListObject::Rule(rule));
                     }
@@ -941,7 +941,7 @@ fn get_dnat_rules_for_addr_family(
                 if let Some(stmt) = &daddr_cond {
                     localhost_jump_statements.push(stmt.clone());
                 }
-                localhost_jump_statements.push(dport_cond.clone());
+                localhost_jump_statements.push(dport_cond);
                 localhost_jump_statements.push(get_jump_action(MASKCHAIN));
                 rules.push(make_rule(&subnet_dnat_chain, localhost_jump_statements));
             }
```

src/firewall/nft.rs
```diff
@@ -0,0 +1,1114 @@
...
+
+        Ok(())
+    }
+
+    fn teardown_network(&self, tear: internal_types::TearDownNetwork) -> NetavarkResult<()> {
+        if !tear.complete_teardown {
+            log::info!("Nothing to tear down, network still in use");
+
+            return Ok(());
+        }
```
Luap99, line 333: This is odd, the same check is already done in the bridge code so teardown_network() is never called in this case.
Reading the iptables code confuses me because there is bunch of conditionals on that that IMO really make no sense.

Anyway I think you can just remove the condition here.

src/firewall/nft.rs
```diff
@@ -0,0 +1,1114 @@
...
+                    for statement in &r.expr {
+                        match statement {
+                            stmt::Statement::Match(m) => match &m.right {
+                                expr::Expression::Named(expr::NamedExpression::Prefix(p)) => {
+                                    match *p.addr.clone() {
+                                        expr::Expression::String(s) => {
+                                            if *s == subnet.addr().to_string() {
+                                                return true;
+                                            }
+                                        }
```
Luap99: The wording says it checks the subnet but it actually only matches the network address (subnet.addr().to_string()) and thus strips off the the netmask, is this intentional?

mheon:
Diff applied, TY

src/firewall/nft.rs
```diff
@@ -0,0 +1,1114 @@
...
+                    }),
+                    stmt::Statement::Drop(None),
+                ],
+            ));
+        }
+
+        // Basic forwarding for all subnets
+        if let Some(nets) = network_setup.subnets {
+            for subnet in nets {
+                let chain = get_subnet_chain_name(subnet, false);
```
Luap99, line 229: I wonder given nft support both ipv4 and v6 at once if doing this by address is the wrong approach? Maybe we should switch to using the bridge interface as filter (iifname oifname)?

Then we only need ip addresses for the port forwarding rules.

src/firewall/nft.rs
mheon: Should we just remove the `complete_teardown` bool entirely if this is the case?

src/firewall/nft.rs
mheon: It's mostly me being lazy and figuring that there will never be two networks with the same network address but different subnet masks (given we have guarantees around that elsewhere), so we can just check half of the match payload.

src/firewall/nft.rs
mheon: It feels like this might break IPv4/IPv6 selective enablement on the network (in that we'll have v6 on networks that otherwise might not, because they auto-assigned link-local addresses) but I'm not sure about that

src/firewall/nft.rs
mheon: Ah, TearDownNetwork is not unique to firewall, so we can't. OK

src/firewall/nft.rs
Luap99: ah good point, not sure about link local but I think you are right there could be a future where we want to NAT ipv4 but not NAT the ipv6 space at the same time so I guess doing this by subnet is better as it gives us some flexibility in that regards

src/firewall/nft.rs
Luap99: looks like p.len contains the subnet size
so something like `&& Into::<u32>::into(subnet.prefix_len()) == p.len` should match as well juts to be sure

mheon: Done

src/firewall/nft.rs
Luap99: So I really do not like the name like that, my main issue is that with the default `podman network create` we juts iterate trough the subnets to assign them, so it is normal to have 10.89.0.0/24, 10.89.1.0/24, 10.89.2.0/24  and so on. This makes it hard for me to follow the chains as only a single digit changes in the name. I have no problem with having the subnet as part of the name but I would prefer if lets say 12 chars of the network ID are added as suffix to make them more distinct to my eyes.

I know we don't really have to look at this often but when we get bug reports we have to and I think a more distinct name will make it easier to find problems (e.g. missing chains/rules)

Luap99:
I noticed these errors in the journal when testing this on my host:
```
Jan 10 16:44:31 pholzing-fedora /usr/bin/podman[42596]: time="2024-01-10T16:44:31+01:00" level=info msg="netavark: [DEBUG netavark::firewall::firewalld] Removing firewalld rules for IPs 10.88.0.0/16\n"
Jan 10 16:44:31 pholzing-fedora firewalld[2202]: ERROR: UNKNOWN_ERROR: rule ref count bug, missing ref count: rule_key '{"%%POLICY_SORT_KEY%%": [0, 1, "trusted", "10.88.0.0/16", "", 0, 1, "trusted", "10.88.0.0/16", "", 0, -1], "chain": "filter_FORWARD_POLICIES", "expr": [{"match": {"left": {"payload": {"field": "saddr", "protocol": "ip"}}, "op": "==", "right": {"prefix": {"addr": "10.88.0.0", "len": 16}}}}, {"match": {"left": {"payload": {"field": "daddr", "protocol": "ip"}}, "op": "==", "right": {"prefix": {"addr": "10.88.0.0", "len": 16}}}}, {"jump": {"target": "filter_FWD_policy_netavark_portfwd"}}], "family": "inet", "table": "firewalld"}'
Jan 10 16:44:31 pholzing-fedora firewalld[2202]: ERROR: COMMAND_FAILED: UNKNOWN_ERROR: rule ref count bug, missing ref count: rule_key '{"%%POLICY_SORT_KEY%%": [0, 1, "trusted", "10.88.0.0/16", "", 0, 1, "trusted", "10.88.0.0/16", "", 0, -1], "chain": "filter_FORWARD_POLICIES", "expr": [{"match": {"left": {"payload": {"field": "saddr", "protocol": "ip"}}, "op": "==", "right": {"prefix": {"addr": "10.88.0.0", "len": 16}}}}, {"match": {"left": {"payload": {"field": "daddr", "protocol": "ip"}}, "op": "==", "right": {"prefix": {"addr": "10.88.0.0", "len": 16}}}}, {"jump": {"target": "filter_FWD_policy_netavark_portfwd"}}], "family": "inet", "table": "firewalld"}'
Jan 10 16:44:31 pholzing-fedora /usr/bin/podman[42596]: time="2024-01-10T16:44:31+01:00" level=info msg="netavark: [WARN  netavark::firewall::firewalld] Error removing subnet 10.88.0.0/16 from firewalld trusted zone: org.fedoraproject.FirewallD1.Exception: COMMAND_FAILED: UNKNOWN_ERROR: rule ref count bug, missing ref count: rule_key '{\"%%POLICY_SORT_KEY%%\": [0, 1, \"trusted\", \"10.88.0.0/16\", \"\", 0, 1, \"trusted\", \"10.88.0.0/16\", \"\", 0, -1], \"chain\": \"filter_FORWARD_POLICIES\", \"expr\": [{\"match\": {\"left\": {\"payload\": {\"field\": \"saddr\", \"protocol\": \"ip\"}}, \"op\": \"==\", \"right\": {\"prefix\": {\"addr\": \"10.88.0.0\", \"len\": 16}}}}, {\"match\": {\"left\": {\"payload\": {\"field\": \"daddr\", \"protocol\": \"ip\"}}, \"op\": \"==\", \"right\": {\"prefix\": {\"addr\": \"10.88.0.0\", \"len\": 16}}}}, {\"jump\": {\"target\": \"filter_FWD_policy_netavark_portfwd\"}}], \"family\": \"inet\", \"table\": \"firewalld\"}'\n"
```
I don't think they are related to this PR but I have to retest with the main/fedora version. Do you see this as well?

mheon:
Nope. Don't see anything like that in journal (F38, fully updated, Podman from main, Netavark from my branch)

Luap99:
f39 ships with firewalld 2.0 and f38 with 1.X I think? So likely a firewalld change then. Anyway we can debug this if you get around to the firewalld stuff, doesn't seem related to this at all as I can reproduce using the fedora version with iptables.

Luap99:
One question though:
```
	chain NETAVARK-HOSTPORT-DNAT {
		tcp dport 80 jump nv_10_88_0_0_nm16_dnat
		tcp dport 81 jump nv_10_88_0_0_nm16_dnat
		tcp dport 82 jump nv_10_88_0_0_nm16_dnat
	}

	chain NETAVARK-HOSTPORT-SETMARK {
		meta mark set meta mark | 0x00002000
	}

	chain nv_10_88_0_0_nm16_dnat {
		ip saddr 10.88.0.0/16 tcp dport 80 jump NETAVARK-HOSTPORT-SETMARK
		ip saddr 127.0.0.1 tcp dport 80 jump NETAVARK-HOSTPORT-SETMARK
		tcp dport 80 dnat ip to 10.88.0.6:80
		ip saddr 10.88.0.0/16 tcp dport 81 jump NETAVARK-HOSTPORT-SETMARK
		ip saddr 127.0.0.1 tcp dport 81 jump NETAVARK-HOSTPORT-SETMARK
		tcp dport 81 dnat ip to 10.88.0.7:80
		ip saddr 10.88.0.0/16 tcp dport 82 jump NETAVARK-HOSTPORT-SETMARK
		ip saddr 127.0.0.1 tcp dport 82 jump NETAVARK-HOSTPORT-SETMARK
		tcp dport 82 dnat ip to 10.88.0.8:80
	}

```

What is the point of the dport match in NETAVARK-HOSTPORT-DNAT? The ..._dnat chain has to match the dport again anyway so just a direct jump should be easier?

mheon:
Comments addressed (added network ID to subnet chain names). Should be ready for merge.

I'm on the firewalld stuff now, so clearly I need to do a dist-upgrade to f39 and see how badly we're broken (and what fixing it will mean - I really don't want to have to conventionalize on firewalld version)

mheon:
I think that jump is mostly to prevent the NETAVARK-HOSTPORT-DNAT chain small. We could eliminate the `nv_10_88_0_0_nm16_dnat` chain entirely and stuff everything into `NETAVARK-HOSTPORT-DNAT` but that would get long and messy (especially if one container decides to do something like a 1000-port range forward). And so long as we have per-network DNAT chains, might as well only jump to them if we actually have to.

src/firewall/nft.rs
mheon: Done. Only 8 characters though (I don't want things to get ridiculously long with IPv6 addresses)

Luap99:
Ok I assumed a port range would actually only result is one rule not in one per port but seems like such feature (https://bugzilla.netfilter.org/show_bug.cgi?id=1501) is not yet implement or very new, not sure what the status is (finding out if patches on ML are merged is always such a chore)

mheon:
If it is merged, it's not in the JSON schema yet, so we can't make use of it. It's unfortunate because it's the only place ranges don't work.

RE: Firewalld - The release notes say 2.0 has no breaking changes. I'm looking deeper.

src/network/internal_types.rs
```diff
@@ -35,6 +37,8 @@ pub struct TearDownNetwork {
 pub struct PortForwardConfigGeneric<Ports, IpAddresses> {
     /// id of container
     pub container_id: String,
+    /// id of the network
+    pub id: String,
```
Luap99, line 41: Ah I just remembered the json parser is very strict and fields are required by default so the netavark firewalld reload service would fail to parse the config files created with the old version, not a huge deal but I think just adding `#[serde(default)]` to both fields might solve it. The string would be empty but given that iptables code does not use it that would be fine.

Also for consistency I think `network_id` is a better name

mheon: Fixed

Luap99 (approved):
LGTM assuming tests pass, nice work. I checked the simple bridge example with https://github.com/containers/netavark/pull/884 and see a 4-5x speedup with nftables compared to iptables.

/lgtm
/hold

src/firewall/nft.rs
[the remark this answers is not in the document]
Luap99: I think using macros that add better syntax sounds great, having something closer to the real nft syntax would be ideal.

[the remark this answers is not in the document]
mheon: I am a *massive* fan of this idea. The existing syntax can be very verbose, and I ended up with a lot of convenience functions to remove that boilerplate; adding macros on the library side would make using the library much more convenient.

And also strongly agree with the JSON API bit. My original attempt at this was using the Mullvad library, but it did not support everything we needed, and the description of that library as being based off reverse-engineered usage of libnftnl inside the nft binary was... concerning.

PR #885, opened 2024-01-15
Finalize firewalld port forwarding support

### LINKED ISSUE #881: Wrong firewalld rule generated when publishing ports on specified ip
When specifying `-p 10.52.0.2:1111:1111` to podman, my intention was to forward traffic coming from `10.52.0.2:1111` to `container:1111`. While when using firewalld as backend, the generated rule is 
```
port=1111:proto=tcp:toport=1111:toaddr=10.52.0.2
```
which will forward any traffic to port 1111 to `10.52.0.2:1111`

***

I think in this case, we should use rich rule in `netavark_portfwd` zone like
```
rule destination address="10.52.0.2" forward-port port=1111 protocol=tcp to-port=1111 to-addr="container-ip"
```
instead of the forward-ports rule.
***
```
rpm -q netavark podman
netavark-1.9.0-1.fc39.x86_64
podman-4.8.2-1.fc39.x86_64
```
***
reproduce
```
# podman run -p 10.52.0.2:1111:1111 -it --rm --log-level debug alpine
....
[DEBUG netavark::firewall::firewalld] Port is ("1111", "tcp", "1111", "10.52.0.2")
....
```

```
# firewall-cmd --info-policy=netavark_portfwd
netavark_portfwd (active)
  priority: -1
  target: CONTINUE
  ingress-zones: ANY
  egress-zones: ANY
  services: 
  ports: 
  protocols: 
  masquerade: no
  forward-ports: 
	port=1111:proto=tcp:toport=1111:toaddr=10.52.0.2
  source-ports: 
  icmp-blocks: 
  rich rules: 
```

mheon:
There are two major changes here.

Firstly, this adds proper support for port forwarding from localhost via a new policy accepting traffic from HOST. This is the last bit we were missing from the original port-forwarding implementation.

Secondly, this fixes a bug where we generated incorrect rules when port-forwarding from a single IP. Instead of doing standard port-forwarding rules, those need rich rules. This was reported as #881.

There are also some small code cleanups in how we handle setting up and tearing down port forwarding. It's still rather ugly, but at least a little better than it was before.

Fixes #881

### DIFF
--- a/test/200-bridge-firewalld.bats
+++ b/test/200-bridge-firewalld.bats
@@ -6,6 +6,7 @@
 load helpers
 
 fw_driver=firewalld
+export NETAVARK_FW=firewalld
 
 function setup() {
     basic_setup
@@ -13,7 +14,6 @@ function setup() {
 }
 
 @test "check firewalld driver is in use" {
-    skip "TODO: Firewalld driver swapped with iptables until firewalld 1.1.0"
     RUST_LOG=netavark=info run_netavark --file ${TESTSDIR}/testfiles/simplebridge.json setup $(get_container_netns_path)
     assert "${lines[0]}" "==" "[INFO  netavark::firewall] Using firewalld firewall driver" "firewalld driver is in use"
 }
@@ -156,11 +156,10 @@ function setup() {
 }
 
 @test "$fw_driver - dual stack dns with alt port" {
-    skip "FIXME (#846): firewalld 2.0 broken port redirect"
     # get a random port directly to avoid low ports e.g. 53 would not create iptables
     dns_port=$((RANDOM+10000))
 
-    NETAVARK_FW=firewalld NETAVARK_DNS_PORT="$dns_port" \
+    NETAVARK_DNS_PORT="$dns_port" \
         run_netavark --file ${TESTSDIR}/testfiles/dualstack-bridge.json \
         setup $(get_container_netns_path)
 
@@ -190,7 +189,7 @@ function setup() {
     run_in_container_netns dig +short "somename.dns.podman" @fd10:88:a::1
     assert "${lines[0]}" =~ "10.89.3.2" "ipv6 dns resolution works"
 
-    NETAVARK_FW=firewalld NETAVARK_DNS_PORT="$dns_port" \
+    NETAVARK_DNS_PORT="$dns_port" \
         run_netavark --file ${TESTSDIR}/testfiles/dualstack-bridge.json \
         teardown $(get_container_netns_path)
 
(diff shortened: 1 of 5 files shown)


mheon:
Tagging WIP as I'm still working on some bugs, seemingly related to firewalld 2.0

mheon:
Hm. Tests aren't even running firewalld, despite fw_driver=firewalld being set.

mheon:
Localhost port forwarding: still broken, being worked on.
Remote port forwarding: should be fully working after latest push.

mheon:
It is, though there were problems on the firewalld side that needed to be resolved before I went ahead. Those should be fixed, so I'm hoping to get back to this in January.

mheon:
@Luap99 After some further work, I have this working, with some serious caveats.
* We cannot hit access forwarded ports locally except via localhost due to a missing bit of firewalld functionality.
* Isolation driver had to be yanked out. Had to add a new firewalld policy to deal with localhost forwarding. Problem: every time we added a zone for isolation, that policy would have to be edited, and policy changes require a firewalld reload. Which wipes out all our non-permanent rules, breaking forwarding to all containers every time we want to add a container in a new zone.

Problem 2 I think we can ship with, I'll just add an error if isolation is requested. Problem 1 prevents a lot of testing we could do in CI. I can sort of work around this until firewalld adds the functionality we need by querying firewalld to make sure the right rules are in place but that doesn't seem like a complete solution.

mheon:
Dropping WIP. All port forwarding tests skipped until firewalld adds additional rich-rule syntax to let us do non-localhost port forwarding on the same machine.

test/200-bridge-firewalld.bats
```diff
@@ -156,7 +156,6 @@ function setup() {
 }
 
 @test "$fw_driver - dual stack dns with alt port" {
-    skip "FIXME (#846): firewalld 2.0 broken port redirect"
```
Luap99, line 159: please remove the `NETAVARK_FW=firewalld` calls below in this test

test/200-bridge-firewalld.bats
```diff
@@ -212,115 +211,165 @@ function setup() {
 }
 
 @test "$fw_driver - port forwarding ipv4 - tcp" {
+    skip "test requires firewalld same-machine port forwarding for non-localhost IP"
```
Luap99, line 214: I guess we have a testing cap in the other files as well, there is no 127.0.0.1 port forwarding test at ALL unless I am missing something.

I would assume we could have something like 
```
@test "$fw_driver - port forwarding with localhost - tcp" {
    test_port_fw hostip="127.0.0.1"
}
```

That should work with all drivers I guess

Luap99 (commented):
We should have more tests that validate the created rules with firewall-cmd commands like we do with iptables and nft...
However I would not consider this blocking as this is not fully working anyways and in case you update the rules again this would just mean you have to fix all the tests so I am good with that.

I have no tried to understand the firewalld rules as this would take me far to long

mheon:
Comments addressed

PR #886, opened 2024-01-16
Bump nftables-rs to latest commit

mheon:
Instead of vendoring by tag (~ a year old), grab the latest commit. This gets us a number of schema changes that are necessary to safely decode all possible JSON generated by nftables.

Solved an issue I ran into during demos today.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1456,8 +1456,7 @@ dependencies = [
 [[package]]
 name = "nftables"
 version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


baude:
equiv to a vendor update ... once lgtm enough

/lgtm

Luap99:
No problem with the change but this will require changes in regards to vendoring as we need to list the new source in the .cargo/config file, see https://github.com/containers/netavark/blob/main/DISTRO_PACKAGE.md#distro-packagers and compare the `cargo vendor` output.

So this will require changes to the rpm spec file and for other packagers that depend on it so this must be made clear in the  release notes as well.

Issue #709, opened 2023-06-02
Bridge network isolation blocks access to other containers published ports


shellt0pia:
I'm opening a new issue to address the same thing as #679 which was closed by its author because I'm encountering the problem myself.

I have this problem on all of my homelab hosts, I reproduced the issue in a Vagrant environment with the following Vagrantfile:

```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "fedora/38-cloud-base"
  config.vm.provider "virtualbox" do |vb|
    vb.cpus = "2"
    vb.memory = "2048"
  end
end
```

Some network/firewall information about the environment:

```
[root@localhost ~]# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 52:54:00:d6:6b:09 brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    inet 10.0.2.15/24 brd 10.0.2.255 scope global dynamic noprefixroute eth0
       valid_lft 85667sec preferred_lft 85667sec
    inet6 fe80::5487:511f:600:c001/64 scope link noprefixroute
       valid_lft forever preferred_lft forever
[root@localhost ~]# firewall-cmd --state
-bash: firewall-cmd: command not found
[root@localhost ~]# nft list ruleset
[root@localhost ~]#
[root@localhost ~]# iptables -S
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
-N NETAVARK_FORWARD
-N NETAVARK_ISOLATION_1
-N NETAVARK_ISOLATION_2
-A FORWARD -m comment --comment "netavark firewall plugin rules" -j NETAVARK_FORWARD
```

Here, iptables is used, but I have the same problem on my "production" hosts running firewalld with nftables as a backend.

If I try to create a container with a published port in one isolated network, I can't access the published port from a container in another isolated network which is something that does work with docker where networks are isolated by default (all the commands are executed as root):

```
# Create 2 networks with isolate=true
[...]

PR #888, opened 2024-01-16
Add support for isolation to the nftables driver

mheon:
Replicates the current iptables approach (3 isolation chains, with support for strict and normal isolation, or no isolation at all). No significant changes from iptables implementation.

### DIFF
--- a/src/firewall/nft.rs
+++ b/src/firewall/nft.rs
@@ -2,6 +2,7 @@ use crate::error::{NetavarkError, NetavarkResult};
 use crate::firewall;
 use crate::firewall::firewalld;
 use crate::network::internal_types;
+use crate::network::internal_types::IsolateOption;
 use crate::network::types::PortMapping;
 use ipnet::IpNet;
 use nftables::batch::Batch;
@@ -22,6 +23,9 @@ const PREROUTINGCHAIN: &str = "PREROUTING";
 const OUTPUTCHAIN: &str = "OUTPUT";
 const DNATCHAIN: &str = "NETAVARK-HOSTPORT-DNAT";
 const MASKCHAIN: &str = "NETAVARK-HOSTPORT-SETMARK";
+const ISOLATION1CHAIN: &str = "NETAVARK-ISOLATION-1";
+const ISOLATION2CHAIN: &str = "NETAVARK-ISOLATION-2";
+const ISOLATION3CHAIN: &str = "NETAVARK-ISOLATION-3";
 
 const MASK: u32 = 0x2000;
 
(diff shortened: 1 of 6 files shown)


mheon:
Still needs tests, so WIP

mheon:
Enabled the test. Seems to work in local testing.

mheon:
@Luap99 PTAL when you get back

src/firewall/nft.rs
```diff
@@ -1058,6 +1206,26 @@ fn get_rule_matcher_jump_to(jump_target: String) -> Box<dyn Fn(&schema::Rule) ->
     })
 }
 
+/// Make a closure that matches any rule that tests for a match to a given bridge interface.
+fn get_rule_matcher_bridge(bridge: String) -> Box<dyn Fn(&schema::Rule) -> bool> {
+    Box::new(move |r: &schema::Rule| -> bool {
```
Luap99, line 1211: we do not need the box here and we can also get rid of the bridge String with the use of a lifetime.
```diff
diff --git a/src/firewall/nft.rs b/src/firewall/nft.rs
index b123a2a..9c75380 100644
--- a/src/firewall/nft.rs
+++ b/src/firewall/nft.rs
@@ -246,7 +246,7 @@ impl firewall::FirewallDriver for Nftables {
             ));
         }
 
-        let match_our_bridge = get_rule_matcher_bridge(network_setup.bridge_name.clone());
+        let match_our_bridge = get_rule_matcher_bridge(&network_setup.bridge_name);
 
         // If and only if isolation is enabled: add isolation chains.
         // Some isolation rules are shared. Other rules are specific to one type
@@ -506,7 +506,7 @@ impl firewall::FirewallDriver for Nftables {
             }
         }
 
-        let match_our_bridge = get_rule_matcher_bridge(tear.config.bridge_name.clone());
+        let match_our_bridge = get_rule_matcher_bridge(&tear.config.bridge_name);
 
         let mut isolation_rules: Vec<schema::Rule> = Vec::new();
         isolation_rules.append(&mut get_matching_rules_in_chain(
@@ -1207,13 +1207,13 @@ fn get_rule_matcher_jump_to(jump_target: String) -> Box<dyn Fn(&schema::Rule) ->
 }
 
 /// Make a closure that matches any rule that tests for a match to a given bridge interface.
-fn get_rule_matcher_bridge(bridge: String) -> Box<dyn Fn(&schema::Rule) -> bool> {
-    Box::new(move |r: &schema::Rule| -> bool {
+fn get_rule_matcher_bridge(bridge: &str) -> impl '_ + Fn(&schema::Rule) -> bool {
+    move |r: &schema::Rule| -> bool {
         for statement in &r.expr {
             match statement {
                 stmt::Statement::Match(m) => match &m.right {
                     expr::Expression::String(s) => {
-                        if *s == bridge {
+                        if s == bridge {
                             return true;
                         }
                     }
@@ -1223,7 +1223,7 @@ fn get_rule_matcher_bridge(bridge: String) -> Box<dyn Fn(&schema::Rule) -> bool>
             }
         }
         false
-    })
+    }
 }
 
 /// Find all rules in the given chain which match the given closure (true == include).
```

Luap99 (commented):
mostly LGTM, one comment which should make the code a bit simpler IMO

I also assume this has the same problem as described in https://github.com/containers/netavark/issues/709 but that is not a blocker given it would be consistent with iptables and we can fix it at a later point

src/firewall/nft.rs
mheon: Changes pushed, TY

Luap99 (approved):
/lgtm
/hold

PR #894, opened 2024-01-24
RPM: update .cargo/config before building

lsm5:


lsm5:
@Luap99 @baude PTAL

lsm5:
Let's get this in before #893

Luap99 (commented):
https://github.com/containers/netavark/blob/main/DISTRO_PACKAGE.md#distro-packagers needs to be updated as well

I really hate that it includes the rev so we must update this every time we update the nftables dep which is ugly.

lsm5:
> https://github.com/containers/netavark/blob/main/DISTRO_PACKAGE.md#distro-packagers needs to be updated as well
> 
> I really hate that it includes the rev so we must update this every time we update the nftables dep which is ugly.

done.

Luap99 (commented):
LGTM

PR #896, opened 2024-01-25
DISTRO_PACKAGE: fix incorrect vendored tar archive URL

danishprakash:
Given there's a vendored tarball as part of releases, does this mean this is preferred over vendoring at build time `cargo update && cargo vendor` before building?

Signed-off-by: Danish Prakash <danish.prakash@suse.com>

### DIFF
--- a/DISTRO_PACKAGE.md
+++ b/DISTRO_PACKAGE.md
@@ -40,7 +40,7 @@ the official packages.
 The vendored sources for netavark will be attached to each netavark release as
 a tarball. You can download them with the following:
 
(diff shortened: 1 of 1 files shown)


Luap99 (approved):
/lgtm

Luap99:
> Given there's a vendored tarball as part of releases, does this mean this is preferred over vendoring at build time cargo update && cargo vendor before building?

I don't think it matters much, the result should be more or less the same. The only thing to consider is that if you change the dependencies versions we obviously did not tests that and you could run into problems.

PR #905, opened 2024-01-27
Fix impossible dep on aardvark-dns 1.10.1

AdamWill:
netavark 1.10.1 was released alongside aardvark 1.10.0. This tied dependency is impossible to satisfy at present for netavark 1.10.1.

### DIFF
--- a/rpm/netavark.spec
+++ b/rpm/netavark.spec
@@ -41,8 +41,11 @@ Source1: %{url}/releases/download/v%{version}/%{name}-v%{version}-vendor.tar.gz
 BuildRequires: cargo
 BuildRequires: %{_bindir}/go-md2man
 # aardvark-dns and %%{name} are usually released in sync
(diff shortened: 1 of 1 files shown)


AdamWill:
I have used provenpackager privs to do a downstream build with this change also, because this error is breaking openQA tests for all Rawhide updates.

mheon:
@lsm5 PTAL

Luap99 (commented):
Given this is the upstream spec file I rather see a "proper" fix, we should keep the same minor.major version in sync but the patch releases can vary of course.
Is it possible to extract only the X.Y versions from the rpm version macro and use this as minimum number?

lsm5:
@AdamWill thanks for the PR. I'll be switching the dep to depend on major.minor and also switch to manual bodhi updates for netavark and aardvark-dns in packit. Closing this..

PR #910, opened 2024-01-29
ignore Read-only file system error on sysctl set

### LINKED ISSUE #825: RFC: consider read only sysctl errors as non fatal
When running inside of unprivileged containers /proc is normally mounted read only.
Now if a users tries to run netavark it will fail hard if we cannot set all the sysctl's. Most of them are needed for routing or to disable some ipv6 options but general communication may still be possible.

We should consider not treating read only errors as fatal and just log them as warning. The biggest problem is likely the ip_forward sysctl, without it no external communication would be possible. However this could already be set by the outer container manager in which case I would expect it to mostly work fine. 

see https://github.com/containers/podman/issues/19991

haircommander:
fixes https://github.com/containers/netavark/issues/825

Needed for running podman in a kubernetes pod. I am investigating if we can not set Proc to unmasked and get away with enough podman operations.

### DIFF
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -282,7 +282,18 @@ impl CoreUtils {
             }
             Err(e) => return Err(e),
         }
-        ctl.set_value_string(val)
+        let result = ctl.set_value_string(val);
+        // if we have a read only /proc we ignore it as well
+        match result {
+            Ok(_) => result,
+            Err(SysctlError::IoError(e)) => {
+                if e.raw_os_error() == Some(libc::EROFS) {
+                    return Ok(String::from(""));
+                }
+                return Err(e.into());
+            },
+            Err(err) => return Err(err),
+        }
     }
 }
 
@@ -432,9 +443,6 @@ pub fn disable_ipv6_autoconf(if_name: &str) -> NetavarkResult<()> {
                 // if the sysctl is not found we likely run on a system without ipv6
                 // just ignore that case
             }
-
-            // if we have a read only /proc we ignore it as well
-            SysctlError::IoError(ref e) if e.raw_os_error() == Some(libc::EROFS) => {}
 
             _ => {
                 return Err(NetavarkError::wrap(


mheon:
This is probably going to break certain functionality (internal networks, definitely, probably also IPv6 will act strange since we can't configure neighbor advertisements properly).

haircommander:
> This is probably going to break certain functionality (internal networks, definitely, probably also IPv6 will act strange since we can't configure neighbor advertisements properly).

though in those cases, they were broken to begin with. This is just making a fatal error non fatal. If netavark truly needed to set the sysctls then something will break down the line I think

Luap99:
I would at least have a warning printed (although I guess that may defeat the purpose if every single container start-up prints these).
The main issue is that some things simple don't work. If routing is not enabled then there will be no external network connectivity, `route_localnet` would be needed to allow port forwarding via 127.0.0.1. There may be some ipv6 systcl that are not that important but I would need to double check which ones we really need.
Also internal networks will not work at all which will be a major issue as people may depend on the security that a container cannot connect to the internet,etc...

Ignoring these problems will turn into a nightmare of random support requests why xyz isn't working in their env and network issues are already a pain to deal with due to often lacking reproducer and/or race conditions. I don't what to deal with more of these.

I think we should still make the caller enforce that the sysctls are already set to the right values, i.e. it should be possible to set `net.ipv4.conf.default....` for the outer container/pod so that the values are already correct when we create interfaces. In this case the current logic already works as we first read the value and to not try to write if it is correct.

haircommander:
> I think we should still make the caller enforce that the sysctls are already set to the right values

so the code already does that. if the sysctl is set to the right value it skips. the problem is when setting values like `net.ipv4.eth0` or something, which does happen in the default flow. eth0 doesn't exist at container start, so we can't prepopulate the sysctls.. 

I am open to logging something if we skip, or passing an option to skip from podman. TBH I have hit other issues when testing with this patch so I'm not even sure if we should go down this route...

Luap99:
> the problem is when setting values like net.ipv4.eth0 or something, which does happen in the default flow. eth0 doesn't exist at container start, so we can't prepopulate the sysctls..

That is why I suggested the use of the special interface name `default` in the sysctl. All newly created interfaces will inherit the values from the default one and you can set the default sysctl to the correct values in the parent container because it will always exist.

Anyhow I tend to agree this all is less then ideal and would not be future proof, i.e. if we start setting a new sysctl then the parent containers config would need to be updated with it as well. It would be hard to maintain as each update could result in breaking changes for users going that route.

andsens:
> That is why I suggested the use of the special interface name default in the sysctl. All newly created interfaces will inherit the values from the default one and you can set the default sysctl to the correct values in the parent container because it will always exist.

Not always. As an example take podman running rootless in a kubernetes pod. In that case `/proc` is masked, and you need to run with
```
[containers]
volumes = ["/proc:/proc"]
```
This, combined with the root NS being a pod, means that there are truly no sysctls you can change. In our specific usecase [this patch](https://github.com/orbit-online/netavark/commit/c8548b1602b5aed269601b8f7191332a6ca31715) is enough to make things work. DNS resolution works, and containers can communicate.  
It is far from ideal, and I'm sure those sysctls weren't just set for fun, but there are some scenarios where you are *really* constrained.

PR #912, opened 2024-01-30
Do not perform network namespace detection on AV update

### LINKED ISSUE #911: 1.10: rootless: aardvark-dns runs in a different netns, dns will not work
podman e2e tests fail rootless, both [f39](https://api.cirrus-ci.com/v1/artifact/task/6167869470277632/html/int-podman-fedora-39-rootless-host-sqlite.log.html#t--Podman-run-networking-podman-verify-network-scoped-DNS-server-and-also-verify-updating-network-dns-server--1) and [rawhide](https://api.cirrus-ci.com/v1/artifact/task/6730819423698944/html/int-podman-rawhide-rootless-host-sqlite.log.html#t--Podman-run-networking-podman-verify-network-scoped-DNS-server-and-also-verify-updating-network-dns-server--1):
```
$ podman [options] network update IntTest207af0ff0c --dns-add 127.0.0.255
[ERROR netavark::dns::aardvark] aardvark-dns runs in a different netns, dns will not work for this container.  \
      To resolve please stop all containers, kill the aardvark-dns process, \
      remove the /tmp/podman_test3080652420/runroot/networks/aardvark-dns directory \
      and then start the containers again
IntTest207af0ff0c
```
See: https://github.com/containers/podman/pull/21366#issuecomment-1917359468

Given all the problems with *vark 1.10, I recommend pulling them until someone can actually test them.

mheon:
Update is not required to be in the same network namespace as everything else, because we're just restarting an existing server with an updated config. Make the check for Aardvark's netns only happen if we are not updating as such.

Fixes #911

### DIFF
--- a/test/100-bridge-iptables.bats
+++ b/test/100-bridge-iptables.bats
@@ -150,8 +150,11 @@ fw_driver=iptables
     run_helper ps "$aardvark_pid"
     assert "${lines[1]}" =~ ".*aardvark-dns --config $NETAVARK_TMPDIR/config/aardvark-dns -p $dns_port run" "aardvark not running or bad options"
 
-    NETAVARK_DNS_PORT="$dns_port" run_netavark --file ${TESTSDIR}/testfiles/dualstack-bridge-network-container-dns-server.json \
+    # Use run_helper instead of run_netavark here to check network namespace detection logic.
+    # See https://github.com/containers/netavark/issues/911 for details.
+    NETAVARK_DNS_PORT="$dns_port" run_helper $NETAVARK --config "$NETAVARK_TMPDIR/config" --rootless "$rootless" --file ${TESTSDIR}/testfiles/dualstack-bridge-network-container-dns-server.json \
         update podman1 --network-dns-servers 8.8.8.8
+    assert "$output" = ""
 
     # check aardvark config and running
     run_helper cat "$NETAVARK_TMPDIR/config/aardvark-dns/podman1"
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -146,23 +146,27 @@ impl Aardvark {
         Ok(())
     }
 
-    pub fn notify(&self, start: bool) -> NetavarkResult<()> {
+    pub fn notify(&self, start: bool, is_update: bool) -> NetavarkResult<()> {
         match self.get_aardvark_pid() {
             Ok(pid) => {
                 match signal::kill(Pid::from_raw(pid), Signal::SIGHUP) {
-                    Ok(_) => match self.check_netns(pid) {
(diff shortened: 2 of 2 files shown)


mheon:
@Luap99 Do you think we can test this in CI? The netns stuff we're doing in Netavark testing confuses me a bit

Luap99 (commented):
code LGTM, for tests you can copy paste the netavark command in `run_netavark` and run it with `run_helper` instead then we will not join the same netns so it should trigger this warning. Note that this is only printed and not a error exit code so you have to check the full output.

test/100-bridge-iptables.bats
```diff
@@ -150,7 +150,7 @@ fw_driver=iptables
     run_helper ps "$aardvark_pid"
     assert "${lines[1]}" =~ ".*aardvark-dns --config $NETAVARK_TMPDIR/config/aardvark-dns -p $dns_port run" "aardvark not running or bad options"
 
-    NETAVARK_DNS_PORT="$dns_port" run_netavark --file ${TESTSDIR}/testfiles/dualstack-bridge-network-container-dns-server.json \
+    NETAVARK_DNS_PORT="$dns_port" run_helper --file ${TESTSDIR}/testfiles/dualstack-bridge-network-container-dns-server.json \
         update podman1 --network-dns-servers 8.8.8.8
```
edsantiago, line 154: I think you need to add `assert "$output" = ""` or something similar. Remember, the message does not cause a program failure or error exit status.

test/100-bridge-iptables.bats
```diff
@@ -150,8 +150,9 @@ fw_driver=iptables
     run_helper ps "$aardvark_pid"
     assert "${lines[1]}" =~ ".*aardvark-dns --config $NETAVARK_TMPDIR/config/aardvark-dns -p $dns_port run" "aardvark not running or bad options"
 
-    NETAVARK_DNS_PORT="$dns_port" run_netavark --file ${TESTSDIR}/testfiles/dualstack-bridge-network-container-dns-server.json \
+    NETAVARK_DNS_PORT="$dns_port" run_helper $NETAVARK --config "$NETAVARK_TMPDIR/config" --rootless "$rootless" --file ${TESTSDIR}/testfiles/dualstack-bridge-network-container-dns-server.json \
```
Luap99, line 153: a comment on what we do this here may be appropriate as it may be forgotten otherwise

mheon: Done

edsantiago:
Thank you! @mheon @lsm5 any idea when we can expect a rawhide bodhi?

mheon:
We can get a version cut this afternoon and work with that

PR #922, opened 2024-02-07
fix netavark update to not start a new aardvark-dns

Luap99:
While commit 45f158118d fixed one problem it added another one as well. The condition of !is_update got mixed up and now never returned in the good case thus the code went to the end of the function and started a new aardvark-dns conflicting with an existing running arrdvark-dns.


Luap99:
@mheon @baude PTAL, we need a new release with this. Issue reported in https://github.com/containers/automation_images/pull/327#issuecomment-1931973876
cc @edsantiago

Luap99:
I locally verified it against the failing `podman verify network scoped DNS server and also verify updating network dns server` test that this fix makes it work again.

PR #950, opened 2024-03-22
fix Total success task name

Luap99:
Ever since the merge bot change[1] the bot is blocking on the task name, it was set to "Total Success" while this repo had "Total success" so I think this is the reason the bot is no longer merging things here.

[1] https://github.com/openshift/release/commit/74ec1922d6d53e0d2cc25358afff869d258e485d

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -251,7 +251,7 @@ centos9_build_task:
 
 
 success_task:
-  name: "Total success"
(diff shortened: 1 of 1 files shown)


Luap99:
Actually lets just rename this in the prow config, otherwise we need to backport this change + change github setting for all our older branches as well.

PR #956, opened 2024-04-03
fix aardvark-dns netns check

Luap99:
Right now there is a race condition where we return errors even in cases where they should be ignored. When we send SIGHUP to aardvark on teardown it might exit when all containers are removed. This means the check afterwards might read the netns path at a weird time while the process is being removed from the kernel structures. I assumed the only error can be ENOENT but I was wrong, in CI we also see EACCES and in my reproducer I also saw ESRCH. Given the check is a nice to have do ignore all errors there.

Fixes containers/podman#22103

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -127,9 +127,21 @@ impl Aardvark {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

Issue #942, opened 2024-03-08
`nftables error: got invalid json: data did not match any variant of untagged enum NfObject`


g00g1:
<details><summary>podman info</summary>
<p>

```
admin@flopster ~ $ sudo podman info
host:
  arch: amd64
  buildahVersion: 1.35.0
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - misc
  cgroupManager: cgroupfs
  cgroupVersion: v2
  conmon:
    package: app-containers/conmon-2.1.8
    path: /usr/libexec/podman/conmon
    version: 'conmon version 2.1.8, commit: 00e08f4a9ca5420de733bf542b930ad58e1a7e7d'
  cpuUtilization:
    idlePercent: 92.28
    systemPercent: 0.82
    userPercent: 6.9
  cpus: 12
  databaseBackend: sqlite
  distribution:
    distribution: gentoo
    version: "2.14"
  eventLogger: file
  freeLocks: 2043
  hostname: flopster.at.encryp.ch
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 6.6.16
  linkmode: dynamic
  logDriver: k8s-file
  memFree: 44018987008
  memTotal: 67366854656
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    dns:
      package: app-containers/aardvark-dns-1.10.0
      path: /usr/libexec/podman/aardvark-dns
      version: aardvark-dns 1.10.0
    package: app-containers/netavark-1.10.3
    path: /usr/libexec/podman/netavark
    version: netavark 1.10.3
  ociRuntime:
    name: crun
    package: app-containers/crun-1.8.4
    path: /usr/bin/crun
    version: |-
      crun version 1.8.4
      commit: 5a8fa99a5e41facba2eda4af12fa26313918805b
      rundir: /run/crun
      spec: 1.0.0
      +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  os: linux
  pasta:
    executable: ""
    package: ""
    version: ""
  remoteSocket:
    exists: true
    path: /run/podman/podman.sock
  security:
    apparmorEnabled: false
[...]

PR #959, opened 2024-04-03
update to nftables 0.4 and only read netavark table rules

Luap99:
This contains some breaking changes to the payload type[1] so fix them.

[1] https://github.com/namib-project/nftables-rs/commit/0deaf7d86429b79b857a25f3cd1fb7fdeef8b056

nftables: only dump netavark table rules

This makes things easier as we the lib no longer needs to parse
unrelated rules, we only need to touch/care about the netavark table.

One big issue is that the rust nftables lib is not feature complete and
all rules that it doesn't undersatnd cause the json deserialization to
fail hard and thus easily break netavark. With this fix we only dump our
rules and because we the rules via this lib we should never encounter
unspported rules and thus it should always work.

I assume this also results in a nice speed up if there are many rules in
other tables on the host as they will never be processed by the paring
logic.

Unfortunately this implementation has one big issue, nft will fail when
the table does not exists. This is normal on the first run after boot of
course. To fix this we need to add some special error handling to read
the stderr and we get the ENOENT error message on stderr so we know when
this is the case and can ignore silently.

Fixes https://github.com/containers/netavark/issues/942

### DIFF
--- a/src/firewall/nft.rs
+++ b/src/firewall/nft.rs
@@ -381,10 +381,10 @@ impl firewall::FirewallDriver for Nftables {
                         get_subnet_match(&subnet, "saddr", stmt::Operator::EQ),
                         stmt::Statement::Match(stmt::Match {
                             left: expr::Expression::Named(expr::NamedExpression::Payload(
-                                expr::Payload {
+                                expr::Payload::PayloadField(expr::PayloadField {
                                     protocol: "udp".to_string(),
                                     field: "dport".to_string(),
-                                },
+                                }),
                             )),
                             right: expr::Expression::Number(53),
                             op: stmt::Operator::EQ,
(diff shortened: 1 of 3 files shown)


mheon:
LGTM

Luap99:
@baude PTAL

src/firewall/nft.rs
```diff
@@ -1288,3 +1292,28 @@ fn get_chain(base_rules: &schema::Nftables, chain: &str) -> Option<schema::Chain
 
     None
 }
+
+fn get_netavark_rules() -> Result<schema::Nftables, helper::NftablesError> {
+    match helper::get_current_ruleset(None, Some(vec!["list", "table", "inet", TABLENAME])) {
```
baude, line 1297: any value in filing an issue upstream with nftables to see if they can give us a better API? I suppose we could also contribute one.

baude:
small consideration, merging

/lgtm

PR #964, opened 2024-04-16
renovate: stop rebasing PRs automatically

Luap99:
This is not really helpful, as it makes it impossible to merge many PRs in a row. The risk of conflicts is rather low IMO so we should not require this by default. We can still rebase manually via the button if needed.


Luap99:
@mheon @baude @cevich PTAL

PR #965, opened 2024-04-23
Add suffix to Aardvark internal network filenames

mheon:
Use the %int suffix for internal network files passed to AV so we can properly trigger DNS handling (or, more accurately, lack of DNS handling) over there. Also, to handle version migrations, if there is a previous file for the network named without the suffix, delete that.

This will unblock [1].

[1] https://github.com/containers/aardvark-dns/pull/447

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -29,6 +29,7 @@ pub struct AardvarkEntry<'a> {
     pub container_ips_v6: Vec<Ipv6Addr>,
     pub container_names: Vec<String>,
     pub container_dns_servers: &'a Option<Vec<IpAddr>>,
+    pub is_internal: bool,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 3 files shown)


mheon:
CI test failures round 1 are expected. I will disable those tests until https://github.com/containers/aardvark-dns/pull/447 merges

mheon:
@Luap99 PTAL

src/dns/aardvark.rs
```diff
@@ -220,7 +221,11 @@ impl Aardvark {
         }
 
         for entry in &entries {
-            let path = Path::new(&self.config).join(entry.network_name);
+            let mut path = Path::new(&self.config).join(entry.network_name);
+            if entry.is_internal {
+                let _ = std::fs::remove_file(&path);
+                path = Path::new(&self.config).join(entry.network_name.to_owned() + "%int");
+            }
```
Luap99, line 228: This is breaking backwards compat.  If there are already running containers then you just delete all info from them.
The proper approach must be to rename the old path when it exists IMO. That way we migrate it automatically.

Luap99 (commented):
You also must adapt delete_entry() to use both paths, first check for new, then old

mheon:
Tests are... passing. That's unexpected.

baude:
heya @mheon wanna push this through ?

mheon:
I think we're good to merge.

Luap99:
Let me do some actual manual test on the update path to be sure the logic does what we intend, if it works I will merge

Luap99 (approved):
/lgtm

PR #978, opened 2024-05-03
fix ncat sctp test and update CI image 

Luap99:
see commits

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -621,17 +621,18 @@ function run_nc_test() {
     local host_port=$5
 
     local nc_common_args=""
-    local stdin=/dev/null
+    exec {stdin}<>/dev/null
 
     case $proto in
     tcp) ;; # nothing to do (default)
     udp) nc_common_args=--udp ;;
     sctp)
         nc_common_args=--sctp
-        # for some reason we have to attach STDIN to the server only for the sctp proto
-        # otherwise it will just exit for unknown reasons. However we must not add STDIN
-        # to udp and tcp otherwise those tests will fail.
-        stdin=/dev/zero
+        # For some reason we have to attach a empty STDIN (not /dev/null and not something with data in it)
+        # to the server only for the sctp proto otherwise it will just exit for weird reasons.
+        # As such create a empty anonymous pipe to work around that.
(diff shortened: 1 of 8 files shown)


Luap99:
@edsantiago I hope this unblock us hopefully

mheon:
LGTM once it passes CI

Luap99:
@edsantiago @mheon PTAL again

test/helpers.bash
```diff
@@ -644,7 +645,7 @@ function run_nc_test() {
     fi
 
     nsenter -n -t "${CONTAINER_NS_PIDS[$container_ns]}" timeout --foreground -v --kill=10 5 \
-        nc $nc_common_args -l -p $container_port &>"$NETAVARK_TMPDIR/nc-out" <$stdin &
+        nc $nc_common_args -l -p $container_port &>"$NETAVARK_TMPDIR/nc-out" <&$stdin &
```
edsantiago, line 648: Although CI is passing, it's not for the reasons described in your commit: `$stdin` is, at this point, undefined. I expected bash to treat this as an error, but it doesn't, and I'm completely lost as to what exactly is happening.

I *think* the answer here is to remove `<&$stdin` entirely, given that you've set up the `exec`, but I'm sorry I can't approve this as it is.

edsantiago (changes requested):
Very weird behavior, see inline. 

And your third commit (clippy) doesn't say what the clippy warnings actually are, and I'm assuming that "ipva" is a typo for "ipv6" maybe? Not a blocker from my perspective but I mention it for you to decide whether a future maintainer will care about that or not.

test/helpers.bash
Luap99: sorry I don't follow why do you think $stdin is undefined?

Luap99: I will try to remove it, if that makes it I happily do that as it makes the test code simpler

edsantiago: no wait hold on; I'm refreshing my memory on `{varname}`

edsantiago (approved):

edsantiago:
My bad.

/lgtm

PR #979, opened 2024-05-03
[skip-ci] Packit: enable c10s downstream sync

lsm5:
This commit will enable downstream syncing to CentOS Stream 10. The centos maintainer will need to manually run `packit propose-downstream` and `centpkg build` until better centos integration is in place.

This commit also builds both rhel9 and centos9 copr rpms so we can check for things like differences in rust compiler.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,8 +2,17 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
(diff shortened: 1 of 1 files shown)


lsm5:
@mheon @Luap99 PTAL

Luap99 (commented):
LGTM

PR #987, opened 2024-05-14
[skip-ci] RPM: switch default firewall to nftables on EL10+

lsm5:
set NETAVARK_DEFAULT_FW to nftables on EL10+.

### DIFF
--- a/rpm/netavark.spec
+++ b/rpm/netavark.spec
@@ -22,6 +22,15 @@
 # Minimum X.Y dep for aardvark-dns
 %define major_minor %((v=%{version}; echo ${v%.*}))
 
+# Set default firewall to nftables on CentOS Stream 10+, RHEL 10+
+# and default to iptables on all other environments
+# The `rhel` macro is defined on CentOS Stream, RHEL as well as Fedora ELN.
+%if %{defined rhel} && 0%{?rhel} >= 10
+%define default_fw nftables
+%else
+%define default_fw iptables
+%endif
+
 Name: netavark
 # Set a different Epoch for copr builds
 %if %{defined copr_username}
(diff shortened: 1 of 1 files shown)


mheon:
LGTM

lsm5:
ELN failures can be ignored for now. They are fixing ELN to default to dnf5 as we speak.

lsm5:
```
[root@7629c3fd2a93 /]# rpm -q netavark
netavark-1.10.1-1.20240514125744608979.pr987.119.ge7ea9b7.el10.x86_64
[root@7629c3fd2a93 /]# /usr/libexec/podman/netavark version
{
  "version": "1.11.0-dev",
  "commit": "14044b1e19f238939aa53f10d9acf02e0da11c49",
  "build_time": "2024-05-14T00:00:00+00:00",
  "target": "x86_64-unknown-linux-gnu",
  "default_fw_driver": "nftables"
}
```

Tested on centos stream 10 container image.

@Luap99 @mheon @jnovy PTAL

rpm/netavark.spec
```diff
@@ -22,6 +22,12 @@
 # Minimum X.Y dep for aardvark-dns
 %define major_minor %((v=%{version}; echo ${v%.*}))
 
+%if %{defined rhel} && 0%{?rhel} >= 10
```
Luap99, line 25: just to confirm the rhel macro is also defined on centos stream?

lsm5: yes, %{rhel} is also on centos as well as Fedora ELN. But %{centos} is only on centos stream

lsm5: i can add a comment in spec if it helps

lsm5:
Re-pushing with a comment.

rpm/netavark.spec
lsm5: done. PTAL

Luap99 (approved):
/lgtm

PR #988, opened 2024-05-14
[skip-ci] RPM: use default __cargo macro across all envs

lsm5:
On C9S and RHEL9, we've been specifying our own cargo macro which fails to create debuginfo packages with rust 1.77.2. Ref:
https://blog.rust-lang.org/2024/03/21/Rust-1.77.0.html#enable-strip-in-release-profiles-by-default

This can be remedied by simply using the default __cargo macro provided on build environments.

### DIFF
--- a/rpm/netavark.spec
+++ b/rpm/netavark.spec
@@ -2,14 +2,6 @@
 # Latest upstream rtnetlink frequently required
 # sha2, zbus, zvariant are currently out of date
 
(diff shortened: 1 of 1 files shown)


lsm5:
@Luap99 @mheon c9s builds are fixed now. PTAL

Luap99 (approved):
/lgtm

PR #994, opened 2024-05-20
test: Move firewalld tests into one file

Mingli-Yu:
Moving firewalld related tests into one file should be convenient for the user who doesn't care firewalld.


test/200-bridge-firewalld.bats
```diff
@@ -329,3 +329,71 @@ function setup() {
     expected_rc=1 run_netavark -f ${TESTSDIR}/testfiles/invalid-port.json setup $(get_container_netns_path)
     assert_json ".error" "invalid host ip \"abcd\" provided for port 8080" "host ip error"
 }
+
+@test "$fw_driver - test firewalld reload" {
```
Luap99, line 333: you need to fix the test name to use the real driver and make the test use `NETAVARK_FW=iptables run_netavark ...` to set the correct driver.

test/200-bridge-firewalld.bats
```diff
@@ -329,3 +329,71 @@ function setup() {
...
+    setup_firewalld
```
Luap99, line 334: this must be dropped now that it is part of the firewalld file

Mingli-Yu: Thanks! The change is in the latest PR now.

Mingli-Yu:
ping.

Thanks!

Luap99:
@Mingli-Yu the tests are failing so there isn't much point for me to review this. It is up to you to fix that.
Also you have not fixed my earlier review comment with the `$fw_driver` in the test name which is not correct

Mingli-Yu:
@Luap99  Many thanks for your respond! I actually didn't fix any test cases right now, just try to group all firewalld related cases into one file. And latest PR did have the problem with the $fw_driver in the test name. Do you plan to group all firewalld related cases into one file? If yes, I will send a new PR then.

PR #1001, opened 2024-05-30
Release1.11.0

baude:
* Do not perform namespace detection for aardvark-dns updates as it is not needed
* Fixed condition where ignored errors were being returned as real
* With nftables, only dump netavark table rules
* Fix port forward with strict RPF and multi-networks


mheon:
LGTM

Luap99 (approved):
/lgtm

PR #1012, opened 2024-06-25
test: add macvlan metric test

Luap99:
Make sure macvlan supports the metric option, the code did so for a while but there never was a metric test added here and it didn't work in podman[1]. So make sure we test it works and do not regress.

[1] https://github.com/containers/common/issues/2051


baude:
LGTM

flouthoc (approved):
LGTM

PR #1013, opened 2024-06-27
Fix Epoch so upgrade path is preserved from Fedora/RHEL.

jnovy:
The Epoch in spec file needs to be bumped to preserve upgrade path in c10s.

### DIFF
--- a/rpm/netavark.spec
+++ b/rpm/netavark.spec
@@ -28,7 +28,7 @@ Name: netavark
 %if %{defined copr_username}
 Epoch: 102
 %else
-Epoch: 0
+Epoch: 2
 %endif
 Version: 0
 Release: %autorelease


Luap99 (approved):
LGTM
@lsm5 PTAL

lsm5 (commented):
@jnovy I think we should fix Epoch in aardvark-dns first because this change in netavark at line #50 expects aardvark-dns with Epoch 2.

EDIT: I think both changes can be merged together. Could you please create same PR for aardvark-dns as well ?

jnovy:
Sure think @lsm5 : https://github.com/containers/aardvark-dns/pull/475

And thanks for noticing this!

lsm5 (approved):
/lgtm
/approve

PR #1015, opened 2024-06-28
Cirrus: remove centos9 build job

lsm5:
Packit ensures c9s buildability, so there's no need for this job.

Issue #775, opened 2023-08-11
replace ubuntu/centos container image build tests


Luap99:
Currently we do a centos9 and ubuntu 2004 build test in containers, as @cevich noted cirrus CI will start charging for that because we are over the free limit which liekly mean failing tests soon.

So I think we should we move the test to a different platform (github action or our existing VMs).

However while we do this I think the current container based build are not that great to test different rust versions.
First we do not have explicit control over the rust version so we actually do not even know what we are building with. Then the quay.io auto builder for these images keeps failing from time to time so manually intervention is needed to update these images so we can get a newer rust.

Because netavark does not really depend on the OS I think it would make more sense to have an explicitly defined Minimum Support Rust Version (MSRV) and then make sure to test this version. Updates to the MSRV will be explicit so we always know what versions we support. Of course we should keep in mind what rust version RHEL uses and not depend on newer versions to not cause downstream build failures.

@baude @mheon @flouthoc @cevich WDYT? Also the problem same applies to the aardvark-dns repo.

baude:
yeah agree

cevich:
Thanks for opening this issue Paul, I appreciate it.

> So I think we should we move the test to a different platform (github action or our existing VMs).

Converting to GHA would likely only be feasible for Ubuntu.  If you'd like to try, please feel free to go for it.  [You may find some of the changes and notes in my buildah experiment helpful](https://github.com/containers/buildah/pull/4966).

General caveat of note:  GHA workflows always source their instructions from either a PR or the `main` branch.  Cirrus always uses the cloned copy of it's YAML.  We execute runs on release branches daily via cirrus-cron.  This means every release branch will also need the cirrus tasks disabled, otherwise they'll continue to use up minutes.

> Then the quay.io auto builder for these images keeps failing

Yeah that thing sucks, and the failure notification sucks too.  If it would help, this repo. is already setup to e-mail `podman-monitor` if any cirrus-cron task fails.  So, you could (for example) add a container image build & push cirrus-cron task to run daily.  Here's a (way more complex than this would need) [example of this in skopeo](https://github.com/containers/skopeo/blob/eca8382a55c65de4643745e0595eb81f33ad542c/.cirrus.yml#L210-L250).

> Because netavark does not really depend on the OS I think it would make more sense to have an explicitly defined Minimum Support Rust Version (MSRV) and then make sure to test this version.

[...]

PR #1025, opened 2024-07-12
Update to zbus 4

nc7s:
Mainly adapting to API changes.

Especially note the error messages, I'm certainly not knowledgeable enough to properly write them, plus there's one with TODO.

There is an apparent repeating pattern (below), which I feel suboptimal, but can't yet find a better alternative. Please enlighten me if you can.

```rust
if let Value::Variant(val) = raw {
    Some(val)
} else {
    None
}
.and_then(|val| val.try_clone().ok())
.ok_or_else(|| NetavarkError::msg("..."))?
```

### DIFF
--- a/src/firewall/firewalld.rs
+++ b/src/firewall/firewalld.rs
@@ -1,8 +1,8 @@
 use crate::error::{NetavarkError, NetavarkResult};
-use crate::firewall;
 use crate::network::internal_types;
 use crate::network::internal_types::{PortForwardConfig, TearDownNetwork, TeardownPortForward};
 use crate::network::types::PortMapping;
+use crate::{firewall, wrap};
 use core::convert::TryFrom;
 use log::{debug, info, warn};
 use std::collections::HashMap;
@@ -128,28 +128,23 @@ impl firewall::FirewallDriver for FirewallD {
             "getPolicySettings",
             &(PORTPOLICYNAME),
         )?;
-        let policy_config: HashMap<&str, Value> = match policy_config_msg.body() {
-            Ok(m) => m,
-            Err(e) => {
-                return Err(NetavarkError::wrap(
-                    format!(
-                        "Error decoding DBus message for policy {PORTPOLICYNAME} configuration"
-                    ),
-                    e.into(),
-                ))
-            }
-        };
+        let body = policy_config_msg.body();
+        let mut policy_config: HashMap<&str, Value> = wrap!(
+            body.deserialize(),
+            format!("Error decoding DBus message for policy {PORTPOLICYNAME} configuration")
+        )?;
 
         let mut port_forwarding_rules: Array;
-        match policy_config.get("forward_ports") {
-            Some(a) => match a {
-                Value::Array(arr) => port_forwarding_rules = arr.clone(),
-                _ => {
+        match policy_config.remove("forward_ports") {
+            Some(a) => {
+                port_forwarding_rules = if let Value::Array(arr) = a {
+                    arr
+                } else {
                     return Err(NetavarkError::msg(
                         "forward-port in firewalld policy object has a bad type",
-                    ))
-                }
-            },
+                    ));
+                };
+            }
             None => {
                 // No existing rules
                 // Make us a new array.
@@ -199,15 +194,16 @@ impl firewall::FirewallDriver for FirewallD {
         let mut rich_rules_option: Option<Array> = None;
         if setup_portfw.dns_port != 53 && !setup_portfw.dns_server_ips.is_empty() {
             let mut rich_rules: Array;
-            match policy_config.get("rich_rules") {
-                Some(a) => match a {
-                    Value::Array(arr) => rich_rules = arr.clone(),
-                    _ => {
+            match policy_config.remove("rich_rules") {
+                Some(a) => {
+                    rich_rules = if let Value::Array(arr) = a {
+                        arr
+                    } else {
                         return Err(NetavarkError::msg(
                             "forward-port in firewalld policy object has a bad type",
-                        ))
-                    }
-                },
+                        ));
+                    };
+                }
                 None => {
                     // No existing rules
                     // Make us a new array.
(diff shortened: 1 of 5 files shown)


Luap99:
Note you have validate failures (make validate) that have to be fixed.

@mheon PTAL

mheon:
I'm still thinking we replace zbus entirely with a simpler library at some point, but I have no objection to merging this in the interim.

mheon:
Changes look fine once you can get CI to pass.

src/firewall/firewalld.rs
```diff
@@ -333,19 +335,29 @@ impl firewall::FirewallDriver for FirewallD {
...
-                            Value::Str(s) => s.as_str().to_string(),
+                        let port_ip = match fields[3].try_clone() {
+                            Ok(Value::Str(s)) => s.as_str().to_string(),
                             _ => return Err(NetavarkError::msg("Port forwarding tuples must contain only strings, encountered a non-string object")),
                         };
                         debug!("IP string from firewalld is {}", port_ip);
                         if port_ip != ipv4 && port_ip != ipv6 {
-                            port_forwarding_rules.append(port_tuple.clone())?;
+                            port_forwarding_rules.append(port_tuple.try_clone().map_err(
+                                |e| NetavarkError::wrap("TODO: write a proper message", e.into()),
```
nc7s, line 359: @Luap99 @mheon please help decide how this should be written; I don't really have an idea.

mheon: "Unable to clone port forwarding tuple" maybe?

nc7s: Sure, I'll wait for all checks to finish then push it.

nc7s:
centos9_build and ubuntu20_build failed because the zbus & zvariant 4 series require rustc 1.75; if you don't want to lift build requirements that's a hard no I guess.

src/firewall/firewalld.rs
Luap99: You can remove the clones above here actually

```diff
diff --git a/src/firewall/firewalld.rs b/src/firewall/firewalld.rs
index ab2dcaf..e6e4c2b 100644
--- a/src/firewall/firewalld.rs
+++ b/src/firewall/firewalld.rs
@@ -1,8 +1,8 @@
 use crate::error::{NetavarkError, NetavarkResult};
-use crate::firewall;
 use crate::network::internal_types;
 use crate::network::internal_types::{PortForwardConfig, TearDownNetwork, TeardownPortForward};
 use crate::network::types::PortMapping;
+use crate::{firewall, wrap};
 use core::convert::TryFrom;
 use log::{debug, info, warn};
 use std::collections::HashMap;
@@ -335,28 +335,21 @@ impl firewall::FirewallDriver for FirewallD {
             for port_tuple in old_port_forwarding_rules.iter() {
                 match port_tuple {
                     Value::Structure(s) => {
-                        let fields = s
-                            .try_clone()
-                            .map_err(|e| {
-                                NetavarkError::wrap(
-                                    "Invalid port forwarding rule encountered",
-                                    e.into(),
-                                )
-                            })?
-                            .into_fields();
+                        let fields = s.fields();
                         if fields.len() != 4 {
                             return Err(NetavarkError::msg(
                                 "Port forwarding rule that was not a 4-tuple encountered",
                             ));
                         }
-                        let port_ip = match fields[3].try_clone() {
-                            Ok(Value::Str(s)) => s.as_str().to_string(),
+                        let port_ip = match &fields[3] {
+                            Value::Str(s) => s.as_str().to_string(),
                             _ => return Err(NetavarkError::msg("Port forwarding tuples must contain only strings, encountered a non-string object")),
                         };
                         debug!("IP string from firewalld is {}", port_ip);
                         if port_ip != ipv4 && port_ip != ipv6 {
-                            port_forwarding_rules.append(port_tuple.try_clone().map_err(
-                                |e| NetavarkError::wrap("TODO: write a proper message", e.into()),
+                            port_forwarding_rules.append(wrap!(
+                                port_tuple.try_clone(),
+                                "Unable to clone port forwarding tuple "
                             )?)?;
                         }
                     }
```

Also note the use of the wrap!() macro which we have, that makes adding context to the error much easier.

nc7s: (It occurs to me that `wrap!` should be named `wrap_err!` or something)

Luap99:
> centos9_build and ubuntu20_build failed because the zbus & zvariant 4 series require rustc 1.75; if you don't want to lift build requirements that's a hard no I guess.

That should not a problem, they are ancient and mostly useless anyway as nobody takes care of these build images
ref: https://github.com/containers/netavark/issues/775

We need to merge https://github.com/containers/netavark/pull/1015 and then I guess we should just drop ubuntu job as well

Luap99:
Also 
```diff
diff --git a/src/firewall/firewalld.rs b/src/firewall/firewalld.rs
index ab2dcaf..501eb4f 100644
--- a/src/firewall/firewalld.rs
+++ b/src/firewall/firewalld.rs
@@ -129,7 +129,7 @@ impl firewall::FirewallDriver for FirewallD {
             &(PORTPOLICYNAME),
         )?;
         let body = policy_config_msg.body();
-        let policy_config: HashMap<&str, Value> = body.deserialize().map_err(|e| {
+        let mut policy_config: HashMap<&str, Value> = body.deserialize().map_err(|e| {
             NetavarkError::wrap(
                 format!("Error decoding DBus message for policy {PORTPOLICYNAME} configuration"),
                 e.into(),
@@ -137,17 +137,15 @@ impl firewall::FirewallDriver for FirewallD {
         })?;
 
         let mut port_forwarding_rules: Array;
-        match policy_config.get("forward_ports") {
+        match policy_config.remove("forward_ports") {
             Some(a) => {
                 port_forwarding_rules = if let Value::Array(arr) = a {
-                    Some(arr)
+                    arr
                 } else {
-                    None
-                }
-                .and_then(|arr| arr.try_clone().ok())
-                .ok_or_else(|| {
-                    NetavarkError::msg("forward-port in firewalld policy object has a bad type")
-                })?
+                    return Err(NetavarkError::msg(
+                        "forward-port in firewalld policy object has a bad type",
+                    ));
+                };
             }
             None => {
                 // No existing rules
```

You can do the same for the other policy_config.get() call 

I am sure there are more tricks to avoid some more clones.

src/firewall/firewalld.rs
Luap99: Maybe but renaming is out of scope for this PR

nc7s:
Updated with @Luap99's suggestions; I think I'll leave further refactoring out of this PR ;)

PR #1032, opened 2024-07-24
fix aardvark-dns error handling

Luap99:
Big ops, we didn't handle any error when starting aardvark-dns. While we waited for the fork to happen we never checked the exit status of the command thus all errors were silently ignored. Also the command is supposed to capture the error message so we can return it as proper error string to podman. Before we should wrote to stderr which will not show up for podman-remote users.

Of course right now aardvark-dns doesn't actually report many errors so we never cought this but I am trying to make bind failures visible to users so we need this.

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -116,16 +116,35 @@ impl Aardvark {
         // After https://github.com/containers/aardvark-dns/pull/148 this command
         // will block till aardvark-dns's parent process returns back and let
         // aardvark inherit all the fds.
-        Command::new(aardvark_args[0])
+        let out = Command::new(aardvark_args[0])
             .args(&aardvark_args[1..])
-            .stdin(Stdio::inherit())
-            .stdout(Stdio::inherit())
-            .stderr(Stdio::inherit())
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
             // set RUST_LOG for aardvark
             .env("RUST_LOG", log::max_level().as_str())
             .output()?;
 
-        Ok(())
+        if out.status.success() {
+            return Ok(());
+        }
+        if out.stderr.is_empty() {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "aardvark-dns exited unexpectedly without error message",
+            ));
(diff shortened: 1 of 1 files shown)


Luap99:
@mheon PTAL

src/dns/aardvark.rs
```diff
@@ -116,16 +116,35 @@ impl Aardvark {
         // After https://github.com/containers/aardvark-dns/pull/148 this command
         // will block till aardvark-dns's parent process returns back and let
         // aardvark inherit all the fds.
-        Command::new(aardvark_args[0])
+        let out = Command::new(aardvark_args[0])
             .args(&aardvark_args[1..])
-            .stdin(Stdio::inherit())
-            .stdout(Stdio::inherit())
-            .stderr(Stdio::inherit())
+            .stdin(Stdio::null())
```
mheon, line 121: So... Stdio::null() forces it to use a local buffer? That's rather unintuitive

mheon:
One comment, nonblocking
/lgtm

src/dns/aardvark.rs
Luap99: no, it does not it pus /dev/null on it, we obviously don't want to send anything on stdin so this is the right thing, we also do not care about stdout so there as well.

output() causes the default to be `Stdio::piped()` (write to a buffer) so that is why i didn't set stderr here

mheon: Should we not clear Stdout then, and print it if there's an error? Not that I expect there to be much there, but it might be good policy.

Luap99: sorry I don't understand your question, clear what exactly?

stdin -> dev/null
stdout -> dev/null (aardvark-dns never writes there)
sdterr -> buffered (when aardvark exits with error read all and return message in error string)

mheon: Basically saying we should buffer stdout in case we decide to write to it in the future. But I won't argue too loudly since we've not actually using it now.

Luap99: I will complain loudly if I ever see someone printing on stdout in aardvark-dns :)

PR #1035, opened 2024-07-26
aardvark-dns: trim whitespaces from error text

Luap99:
Noticed this while testing my aardvark-dns changes. As we capture the full stderr it includes the final newline from the error text which will then be returned like this to podman. However because all podman always adds a newline to its error output on the cli it resulted in two newlines being printed. Not a bug deal but still unnecessary and confusing.


Luap99:
@mheon @baude PTAL

mheon:
Sure, LGTM

PR #1048, opened 2024-08-05
silence new rust 1.80 warnings

Luap99:
A new lint check was added[1] that makes sure all cfg values are checked. As such we must tell the compiler which values are valid so it doesn't flag them as unexpected.

[1] https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html#cargorustc-check-cfg-for-buildrsbuild-script

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -75,6 +75,7 @@ fn main() {
         "none" => "none",
         inv => panic!("Invalid default firewall driver {}", inv),
     };
+    println!("cargo:rustc-check-cfg=cfg(default_fw, values(\"nftables\", \"iptables\", \"none\"))");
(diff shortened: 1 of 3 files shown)


Luap99:
@mheon @baude PTAL

mheon:
LGTM

baude:
centos build is failing ... i'll lgtm it when we figure out how we're going to deal with that.

Luap99:
> centos build is failing ... i'll lgtm it when we figure out how we're going to deal with that.

They are unmaintained images that break over and over and over again bc they contain super outdated versions. We have packit build jobs now that should be much better so I say just remove it https://github.com/containers/netavark/pull/1015

Anyhow in this case here I think I have to switch the syntax to the old one to keep working with some older versions

PR #1052, opened 2024-08-11
chore: remove superfluous formatters from debug stmt

flouthoc:
Following debug lines don't need any formatters hence removing them.


Luap99 (approved):
/lgtm

Although IMO none of these serve any useful purpose so IMO we could drop these logs entirely.

PR #1059, opened 2024-08-14
nftables/iptables: add dns rule for tcp as well

Luap99:


mheon:
LGTM once this passes tests

Luap99:
should we backport this into a 1.12.2?

mheon:
Sounds like a good idea to me, we're loosely planning a Podman 5.2.2 anyways to get more Sigstore work landed so another NV (just NV? I don't think we need another AV) would not be out of the question.

Luap99:
I found another issue that we leak container entries for aardvark-dns when we fail to start (i.e. port in use), now that netavark reports an error to podman the ctr is not started but the entries stay leaked.

I open another PR for that.

Luap99:
@baude @flouthoc  PTAL

flouthoc (approved):
LGTM

PR #1060, opened 2024-08-14
aardvark: on start failure delete entries again

Luap99:
If we fail to start make sure to remove the entries again to not leak them in case following starts might work.

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -211,7 +211,7 @@ impl Aardvark {
         Ok(())
     }
 
-    pub fn commit_entries(&self, entries: Vec<AardvarkEntry>) -> Result<()> {
+    pub fn commit_entries(&self, entries: &[AardvarkEntry]) -> Result<()> {
         // Acquire fs lock to ensure other instance of aardvark cannot commit
         // or start aardvark instance till already running instance has not
         // completed its `commit` phase.
@@ -240,7 +240,7 @@ impl Aardvark {
             ));
         }
 
-        for entry in &entries {
+        for entry in entries {
             let mut path = Path::new(&self.config).join(entry.network_name);
             if entry.is_internal {
                 let new_path = Path::new(&self.config).join(entry.network_name.to_owned() + "%int");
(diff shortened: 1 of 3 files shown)


mheon:
LGTM

Luap99:
@baude PTAL

src/dns/aardvark.rs
```diff
@@ -344,8 +344,18 @@ impl Aardvark {
 
     pub fn commit_netavark_entries(&self, entries: Vec<AardvarkEntry>) -> NetavarkResult<()> {
         if !entries.is_empty() {
-            self.commit_entries(entries)?;
-            self.notify(true, false)?;
+            self.commit_entries(&entries)?;
```
flouthoc, line 347: Nit and non-blocking:

Not sure can this be simplified using `map_err` ? Although current code has better readability.

```diff
diff --git a/src/dns/aardvark.rs b/src/dns/aardvark.rs
index c1a967e..ad97a55 100644
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -345,17 +345,12 @@ impl Aardvark {
     pub fn commit_netavark_entries(&self, entries: Vec<AardvarkEntry>) -> NetavarkResult<()> {
         if !entries.is_empty() {
             self.commit_entries(&entries)?;
-            match self.notify(true, false) {
-                Ok(_) => (),
-                Err(e) => {
-                    if let Err(err) = self.delete_from_netavark_entries(&entries) {
-                        log::warn!(
-                            "Failed to delete aardvark-dns entries after failed start: {err}"
-                        );
-                    };
-                    return Err(e);
+            self.notify(true, false).map_err(|e| {
+                if let Err(err) = self.delete_from_netavark_entries(&entries) {
+                    log::warn!("Failed to delete aardvark-dns entries after failed start: {err}");
                 }
-            };
+                e
+            })?;
         }
         Ok(())
     }
```

flouthoc (approved):
LGTM

src/dns/aardvark.rs
Luap99: I think map_err is best used only when you modify the error and not add extra logic so I think the match case makes it more clear what is going on.

flouthoc: Fair enough.

PR #1071, opened 2024-09-02
[skip-ci] Packit: add sidetag to release with aardvark-dns

lsm5:
Packit now supports Fedora sidetags to ship multiple package updates in a single bodhi.
Ref: https://packit.dev/docs/fedora-releases-guide/releasing-multiple-packages

The targets for bodhi update should be `fedora-all` and not
fedora-branched as the update is with builds from a sidetag and not the
default tag.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -88,13 +88,14 @@ jobs:
 
   - job: koji_build
     trigger: commit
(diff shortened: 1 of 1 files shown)


Luap99 (approved):
/lgtm

PR #1075, opened 2024-09-05
nft: remove port forwarding rules correctly

### LINKED ISSUE #1068: Firewall rule leakage in nft backend when /udp and /tcp specified
```
# installed package
$ rpm -q netavark podman 
netavark-1.12.2-1.fc40.x86_64
podman-5.2.1-1.fc40.x86_64

# using nftables
$ cat /etc/containers/containers.conf 
[containers]
userns = "auto"

[network]
firewall_driver = "nftables"

# no containers ATM
$ sudo podman ps --all                                                                               
CONTAINER ID  IMAGE       COMMAND     CREATED     STATUS      PORTS       NAMES

# current tables: clear
$ sudo nft list ruleset|grep 10222


$ sudo podman run -it --rm -p 10.0.1.10:10222:10222/tcp -p 10.0.1.10:10222:10222/udp alpine:latest sh
/ # exit

# check the leaked rule
$ sudo nft list ruleset|grep 10222
		udp dport 10222 jump nv_2f259bab_10_88_0_0_nm16_dnat
		tcp dport 10222 jump nv_2f259bab_10_88_0_0_nm16_dnat
		ip saddr 10.88.0.0/16 ip daddr 10.0.1.10 udp dport 10222 jump NETAVARK-HOSTPORT-SETMARK
		ip saddr 127.0.0.1 ip daddr 10.0.1.10 udp dport 10222 jump NETAVARK-HOSTPORT-SETMARK
		ip daddr 10.0.1.10 udp dport 10222 dnat ip to 10.88.0.10:10222
		ip saddr 10.88.0.0/16 ip daddr 10.0.1.10 tcp dport 10222 jump NETAVARK-HOSTPORT-SETMARK
		ip saddr 127.0.0.1 ip daddr 10.0.1.10 tcp dport 10222 jump NETAVARK-HOSTPORT-SETMARK
		ip daddr 10.0.1.10 tcp dport 10222 dnat ip to 10.88.0.10:10222
```

The key of reproducing the bug is to specify `/tcp` and `/udp` for a same port.

Luap99:
The problem with these match statements to check the rules that they are
hard to keep track of and match 1 to 1 to the rules we generate on
setup. This approach here generates the same rules as setup and uses this
to compare the existing rules against.

Fixes #1068

### DIFF
--- a/src/firewall/nft.rs
+++ b/src/firewall/nft.rs
@@ -12,7 +12,7 @@ use nftables::schema;
 use nftables::stmt;
 use nftables::types;
 use std::collections::HashSet;
-use std::net::IpAddr;
+use std::net::{IpAddr, Ipv4Addr};
 
 const TABLENAME: &str = "netavark";
 
@@ -36,6 +36,8 @@ const DNATPRIO: i32 = -100;
 const SRCNATPRIO: i32 = 100;
 /// The filter priority for chains
 const FILTERPRIO: i32 = 0;
+
+const IPV4_LOCALHOST: IpAddr = IpAddr::V4(Ipv4Addr::LOCALHOST);
 
 pub struct Nftables {}
 
@@ -1110,7 +1112,7 @@ fn get_dnat_rules_for_addr_family(
                 // Container dnat chain: ip saddr 127.0.0.1 ip daddr <host IP> <proto> dport <port(s)> jump SETMARKCHAIN
                 let mut localhost_jump_statements: Vec<stmt::Statement> = Vec::new();
                 localhost_jump_statements.push(get_ip_match(
-                    &("127.0.0.1".parse()?),
+                    &IPV4_LOCALHOST,
                     "saddr",
                     stmt::Operator::EQ,
                 ));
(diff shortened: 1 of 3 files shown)


Luap99:
@mheon PTAL

src/firewall/nft.rs
```diff
@@ -847,6 +710,55 @@ impl firewall::FirewallDriver for Nftables {
     }
 }
 
+// compare two rules, we only check the chain name and expr,
+// while we can do rule1 == rule2 it will not work how we like.
+// As we use this to compare rules from nft against rules created
+// by us in memory it means the handle id and index can never match.
+fn cmp_rules(rule1: &schema::Rule, rule2: &schema::Rule) -> bool {
+    if rule1.chain == rule2.chain && rule1.expr == rule2.expr {
```
mheon, line 718: This is so much easier than I thought it would have been. Would have made life so much easier when I wrote the code if I knew.

mheon:
LGTM

src/firewall/nft.rs
Luap99: yeah it also opens up the door for some other potential problems, we can never alter the rules we create as teardown would fail to removes rules created on previous versions. But well that is already the case on iptables and if we ever have to do that kind of thing we can add some extra rule matchers to handle old rules as well.

lsm5:
are we planning a new release soon? RE: https://bugzilla.redhat.com/show_bug.cgi?id=2322021

Issue #1079, opened 2024-09-18
iptables: DNAT rule for aardvark-dns on non-standard port is never reached if a container has a port forward for 53


dev-zero:
1. Start a container with `-p 53:53/udp`
2. Create a non-internal bridge network (non-internal due to #1051)
3. Create a container on that network

The result is this:

```
Chain NETAVARK-HOSTPORT-DNAT (2 references)
 pkts bytes target     prot opt in     out     source               destination
    3   192 NETAVARK-DN-1D8721804F16F  6    --  *      *       0.0.0.0/0            0.0.0.0/0            tcp dpt:53 /* dnat name: podman id: db612e4d0603704b2e6a8aa76f3b69c0f49af293928a952f4a04ec65d58441d6 */
 1447  101K NETAVARK-DN-1D8721804F16F  17   --  *      *       0.0.0.0/0            0.0.0.0/0            udp dpt:53 /* dnat name: podman id: db612e4d0603704b2e6a8aa76f3b69c0f49af293928a952f4a04ec65d58441d6 */
    0     0 DNAT       17   --  *      *       0.0.0.0/0            10.89.1.1            udp dpt:53 to:10.89.1.1:54
 ```
 
 This last rule will never be looked at since all packages will already be DNAT'ed to the first container providing the DNS in the chain `NETAVARK-DN-1D8721804F16F`.
 
 And DNS resolution between containers in the same network is therefore broken for all subsequent containers.

PR #1080, opened 2024-09-24
iptables/nftables: add dns dnat rule first

Luap99:
When a container has port 53 forwarded as well we also add dnst rules
for it. This means depending on the order the container dns traffic
might go to another container not aardvark-dns breaking container name
resolution.

The fix to make sure to insert the rule always first.

Fixes https://github.com/containers/netavark/issues/1079

### DIFF
--- a/src/firewall/nft.rs
+++ b/src/firewall/nft.rs
@@ -592,12 +592,22 @@ impl firewall::FirewallDriver for Nftables {
                 match ip {
                     IpAddr::V4(_) => {
                         if setup_portfw.container_ip_v4.is_some() {
-                            batch.add(make_dns_dnat_rule(ip, setup_portfw.dns_port));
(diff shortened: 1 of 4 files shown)


src/firewall/nft.rs
```diff
@@ -1083,7 +1091,12 @@ fn make_dns_dnat_rule(dns_ip: &IpAddr, dns_port: u16) -> schema::NfListObject {
                 flags: None,
             })),
         ],
-    )
+    );
+
+    // rule should be first so it is ordered before the normal contianer DNAT
+    rule.handle = Some(0);
```
mheon, line 1097: Is this still necessary since you're using INSERT instead of ADD above?

Luap99: Maybe not, I can test tomorrow. You might be right that insert defaults to the first position

baude:
LGTM, up to you now @mheon

PR #1081, opened 2024-09-27
Update ci image

Luap99:


Luap99:
Now this is fun, the linter warning makes me us something that was added only in rust 1.76 and it seems epel 9 at least is using something older.
Also the integration test are still flaky/broken with the new images https://github.com/containers/netavark/pull/1069

PR #1095, opened 2024-10-08
[skip-ci] Packit: constrain koji and bodhi to the fedora package

lsm5:
This helps to avoid dup downstream jobs. Doesn't affect upstream.


lsm5:
ah this one's a 400 error

Luap99 (approved):
/lgtm

PR #1098, opened 2024-10-14
define MSRV and check for it in CI

Luap99:
Define 1.76 as our MSRV currently and add a new container image that we use in CI to test that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,15 @@ Netavark is a tool for configuring networking for Linux containers. Its features
 - [Podman](https://podman.io/docs) 4.0+
 - [protoc](https://grpc.io/docs/protoc-installation/)
 
+## MSRV (Minimum Supported Rust Version)
+
+v1.76
+
+We test that Netavark can be build on this Rust version and on some newer versions.
+All newer versions should also build, and if they do not, the issue should be
+reported and will be fixed. Older versions are not guaranteed to build and issues
+will not be fixed.
+
 ## Build
 
 ```console
--- a/contrib/container_images/build_and_publish_rust_image.sh
+++ b/contrib/container_images/build_and_publish_rust_image.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -e
+
+PODMAN=${PODMAN:-podman}
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+DOCKERFILE="$(dirname "${BASH_SOURCE[0]}")/Dockerfile.Rust"
+
+# get the tag from the Dockerfile so we do not duplicate it
+TAG=$(awk -F':' '/FROM/{print $NF}' $DOCKERFILE)
+if [[ -z "$TAG" ]]; then
+    echo "Empty tag in $DOCKERFILE; the tag must specify the rust version to use" >&2
+    exit 1
+fi
+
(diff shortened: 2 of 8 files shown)


Luap99:
@mheon @baude PTAL

README.md
```diff
@@ -22,6 +22,12 @@ Netavark is a tool for configuring networking for Linux containers. Its features
 - [Podman](https://podman.io/docs) 4.0+
 - [protoc](https://grpc.io/docs/protoc-installation/)
 
+## MSRV (Minimum Supported Rust Version)
+
```
mheon, line 26: Add a sentence here: "Netavark is guaranteed to build on this Rust version and later." or similar

contrib/container_images/build_and_publish_rust_image.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+PODMAN=${PODMAN:-podman}
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+DOCKERFILE="$(dirname "${BASH_SOURCE[0]}")/Dockerfile.Rust"
+
+# get the tag from the Dockerfile so we do not duplicate it
+TAG=$(awk -F':' '/FROM/{print $NF}' $DOCKERFILE)
```
edsantiago, line 10: What would you think of `if [[ -z "$TAG" ]]; then echo "FOO! empty tag"; exit 1; fi` or equivalent?

README.md
Luap99: Given rust weird stance on backwards compatibility I find guaranteed to be a very strong word, i.e. it is not so recent that I think rust 1.80 failed to compile older version of the time crate
https://github.com/rust-lang/rust/issues/127343#issuecomment-2218261296

i.e. it is impossible to build aardvark-dns 1.10.0 with rust v1.80 or newer. To their credit they provide a helpful error but still
```
error[E0282]: type annotations needed for `Box<_>`
  --> /home/pholzing/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.30/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
   = note: this is an inference error on crate `time` caused by an API change in Rust 1.80.0; update `time` to version `>=0.3.35` by calling `cargo update`
```

And given we to not test all rust versions there might be some weird stuff in between so I think I rather `s/guaranteed/supposed` in your sentence

contrib/container_images/build_and_publish_rust_image.sh
Luap99: yes sure

README.md
mheon: "We test that Netavark will build on the following Rust version. Newer versions should also build, and if they do not, the issue should be reported and will be fixed. Older versions are not guaranteed to build and issues will not be fixed"?

Luap99: sounds good

mheon:
LGTM

baude:
this is awesome

edsantiago:
...and you even remembered to update the descriptions on quay, for this and the other two images. Thank you.

PR #1099, opened 2024-10-15
rust 1.81 lint fixes and remove NOP dhcp test

Luap99:
Stuff needed for CI image update but that seems to have other issues that need to debugged so I just submit the fixes alone https://github.com/containers/netavark/pull/1081


mheon:
LGTM

Luap99:
Oh good, the test are failing here too, wtf... So the NOP test might not so NOP after all???????

PR #1107, opened 2024-10-26
fix(deps): update rust crate nftables to 0.5.0 (with fixes)

jwhb:
This PR updates [nftables crate](https://github.com/namib-project/nftables-rs) from 0.4.1 to 0.5.0.

`nft.rs` was modified to deal with breaking changes in nftables 0.5.0 (mostly introducing implementation of `Default` that replaces `new()` functions).

Replaces https://github.com/containers/netavark/pull/1106


Luap99 (approved):
LGTM, thank you
@mheon PTAL

lsm5:
> /packit rebuild-failed

LGTM. rawhide builds are successful. And we can ignore epel builds for now given centos builds were successful in the last run.

Luap99:
@lsm5 the epel build roots contains a to old rust so I think we should drop all epel targets, as long as we can build for centos stream I think this is good enough?

PR #1108, opened 2024-10-28
[skip-ci] Packit: Remove epel targets

lsm5:
EPEL targets, which are essentially RHEL + EPEL repos, often contain very old builds of rust, golang and other important tools leading to a lot more failures compared to centos stream.


Luap99 (approved):
LGTM

PR #1111, opened 2024-10-29
Release v1.13.0

baude:
* Fixed bug where port forwarding rules might not be removed correctly on nftables
* Add DNS DNAT rules first with nftables


Luap99 (approved):
LGTM
@mheon PTAL

PR #1115, opened 2024-11-01
[skip-ci] RPM: cleanup changelog conditionals

lsm5:
rpmautospec is now present on all active Fedora and CentOS Stream envs so we don't need such changelog manipulation anymore.


Luap99 (approved):
/lgtm

PR #1123, opened 2024-11-11
setup: on av errors cleanup again

### LINKED ISSUE #1121: network interface/firewall rules leaked after aardvark-dns start error
### Issue Description

Podman version 5.2.3

The issue I'm seeing is identical to containers/podman#14365 which was closed and locked due to inactivity, but it seems was never resolved and was affecting at least a few people.

I recently updated my server from Fedora 39 to Fedora 40 following which a pihole container which was working perfectly before the upgrade stopped functioning, or rather as it turns out port forward for that container stopped working and in a rather interesting way.

I was forwarding port 53 (tcp/udp) on the host to port 53 on the container, I was also forwarding port 8888 on the host to port 80 on the container for pihole's admin interface. After the upgrade port forwarding broke for both ports.

I've played around with different caps, I disabled selinux enforcement on the host and disabled the firewall (although it was correctly configured). I've checked and checked the container configuration, and even managed to prove that it working as expected except for the port forwarding issue (see below).

To cut a very long story short, here's what I discovered after hours of trying to get things working again. I was able to reach both ports through the container IP, thus demonstrating that the container was functioning correctly. When I changed the ports used for forwarding 8888 > 8765 and 53 > 54 port forward worked! Therefore the issue is specific to certain ports, in my experience 53 as in the original ticket but also others including 8888.

A half dozen other containers, all with port forwards are unaffected by this issue. 

I can't see an obvious connection between port 53 and 8888 however maybe those two ports share something in common that triggers a thought for someone. 

1. Run a container forwarding port 53 (or 8888) from host

Port forwarding for some ports is resulting in traffic just disappearing into the void.

Traffic forwarded from the host on mapped ports to reach the container

### podman info output

Podman version 5.2.3
Fedora 40
x86_64

### Podman in a container

No

### Privileged Or Rootless

Privileged

### Upstream Latest Release

No

Additional environment details

### Additional information

Additional information like issue happens only occasionally or issue happens with a particular architecture or on a particular setting

Luap99:
When aardvark-dns error out at the end we already configured interfaces
+ firewall rules for the driver but because we return a error podman considers this a failure and teardown is never called.

Netavark should always cleanup on its own. So on errors make sure to tear down the drivers again.

Fixes #1121

### DIFF
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -3,8 +3,8 @@ use crate::commands::get_config_dir;
 use crate::dns::aardvark::Aardvark;
 use crate::error::{NetavarkError, NetavarkResult};
 use crate::firewall;
-use crate::network::driver::{get_network_driver, DriverInfo};
-use crate::network::netlink::LinkID;
+use crate::network::driver::{get_network_driver, DriverInfo, NetworkDriver};
+use crate::network::netlink::{self, LinkID};
 use crate::network::{self};
 use crate::network::{core_utils, types};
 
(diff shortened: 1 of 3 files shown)


Luap99:
@baude @mheon PTAL

src/commands/setup.rs
```diff
@@ -170,3 +161,18 @@ impl Setup {
         Ok(())
     }
 }
+
+fn teardown_drivers<'a, I>(drivers: I, host: &mut netlink::Socket, netns: &mut netlink::Socket)
+where
+    I: Iterator<Item = &'a Box<dyn NetworkDriver + 'a>>,
```
mheon, line 167: Ew. Unavoidable, probably, but ew.

mheon:
LGTM

Luap99:
Added lgtm again as I just rebased to fix the merge conflict

PR #1125, opened 2024-11-11
Support changing host veth name

M1cha:
Depends on:
https://github.com/containers/common/pull/2245
https://github.com/containers/podman/pull/24535

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -20,8 +20,8 @@ use crate::{
 use super::{
     constants::{
         ISOLATE_OPTION_FALSE, ISOLATE_OPTION_STRICT, ISOLATE_OPTION_TRUE,
-        NO_CONTAINER_INTERFACE_ERROR, OPTION_ISOLATE, OPTION_METRIC, OPTION_MTU,
-        OPTION_NO_DEFAULT_ROUTE, OPTION_VRF,
+        NO_CONTAINER_INTERFACE_ERROR, OPTION_HOST_INTERFACE_NAME, OPTION_ISOLATE, OPTION_METRIC,
+        OPTION_MTU, OPTION_NO_DEFAULT_ROUTE, OPTION_VRF,
     },
     core_utils::{self, get_ipam_addresses, join_netns, parse_option, CoreUtils},
     driver::{self, DriverInfo},
@@ -38,6 +38,8 @@ const NO_BRIDGE_NAME_ERROR: &str = "no bridge interface name given";
 struct InternalData {
     /// interface name of the veth pair inside the container netns
     container_interface_name: String,
+    /// interace name of the veth pair in the host netns
+    host_interface_name: String,
     /// interface name of the bridge for on the host
     bridge_interface_name: String,
     /// static mac address
@@ -86,6 +88,11 @@ impl driver::NetworkDriver for Bridge<'_> {
         let no_default_route: bool =
             parse_option(&self.info.network.options, OPTION_NO_DEFAULT_ROUTE)?.unwrap_or(false);
         let vrf: Option<String> = parse_option(&self.info.network.options, OPTION_VRF)?;
+        let host_interface_name = parse_option(
+            &self.info.per_network_opts.options,
+            OPTION_HOST_INTERFACE_NAME,
+        )?
+        .unwrap_or_else(|| "".to_string());
 
         let static_mac = match &self.info.per_network_opts.static_mac {
             Some(mac) => Some(CoreUtils::decode_address_from_hex(mac)?),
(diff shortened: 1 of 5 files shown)


M1cha:
I don't know why, but I can't get the issue linked properly.

src/network/types.rs
```diff
@@ -112,6 +112,10 @@ pub struct PerNetworkOptions {
     /// MAC address for the container interface.
     #[serde(rename = "static_mac")]
     pub static_mac: Option<String>,
+
+    /// Driver-specific options for this container.
+    #[serde(rename = "options", default)]
+    pub options: HashMap<String, String>,
```
Luap99, line 118: instead of using default I think it is cleaner to wrap this in an Option<>. While that makes accessing values a step more I think it carries the intend better that there can be empty.

src/network/bridge.rs
```diff
@@ -95,6 +97,13 @@ impl driver::NetworkDriver for Bridge<'_> {
         self.data = Some(InternalData {
             bridge_interface_name: bridge_name,
             container_interface_name: self.info.per_network_opts.interface_name.clone(),
+            host_interface_name: self
+                .info
+                .per_network_opts
+                .options
+                .get("host_interface_name")
+                .cloned()
+                .unwrap_or_else(|| "".to_string()),
```
Luap99, line 106: please use parse_option() as split out host_interface_name as constant like the other options

I know parsing a String cannot fail so the error handling is not needed technically but I think it better to use it consistently.

Luap99:
And also this needs tests, please have a look at test/100-bridge-iptables.bats
This should have a test cases where it fails if the veth name exists on the host and one where it works fine and you check the correct interface was created.

Luap99:
```
    host.create_link(host_veth).map_err(|err| match err {
        NetavarkError::Netlink(ref e) if -e.raw_code() == libc::EEXIST => NetavarkError::wrap(
            format!(
                "create veth pair: interface {} already exists on container namespace",
                data.container_interface_name
            ),
            err,
        ),
        _ => NetavarkError::wrap("create veth pair", err),
    })?;
```

This will likely return a wrong/misleading error now. Given both interface are created in one call it would be impossible to know which name conflicted. Maybe we can check if veth_name not empty then chnage the message to hint at both possibilities

src/network/types.rs
M1cha: Agree, fixed.

src/network/bridge.rs
M1cha: Fixed. Do we need to make clear which options come from the network and which from the per-network-config? With the current code, the messages and variables look the same for both kinds of options.

M1cha:
Does this change cause the integration test failures? because they seem unrelated at first glance.

Luap99:
```
not ok 108 nftables - check error message from netns thread
# (from function `assert' in file test/helpers.bash, line 348,
#  from function `assert_json' in file test/helpers.bash, line 382,
#  in test file test/250-bridge-nftables.bats, line 359)
#   `assert_json ".error" "create veth pair: interface eth0 already exists on container namespace: Netlink error: File exists (os error 17)" "interface exists on netns"' failed
#  nsenter -n -m -w -t 23168 ip link set lo up
#  nsenter -n -m -w -t 23170 ip link add eth0 type dummy
#  nsenter -n -m -w -t 23168 ./bin/netavark --rootless false --config /tmp/netavark_bats.zFocvy/config --file /var/tmp/netavark/test/testfiles/simplebridge.json setup /proc/23170/ns/net
# {"error":"create veth pair: interface eth0 already exists on container namespace or  exists on host namespace: Netlink error: File exists (os error 17)"}
# [ rc=1 (expected) ]
#  jq -r .error
# create veth pair: interface eth0 already exists on container namespace or  exists on host namespace: Netlink error: File exists (os error 17)
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #|     FAIL: interface exists on netns
# #| expected: 'create veth pair: interface eth0 already exists on container namespace: Netlink error: File exists (os error 17)'
# #|   actual: 'create veth pair: interface eth0 already exists on container namespace or  exists on host namespace: Netlink error: File exists (os error 17)'
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Yes it is your new error message, it seems your condition in the code is inverted.

src/network/bridge.rs
```diff
@@ -647,17 +655,25 @@ fn create_veth_pair<'fd>(
...
+                format!(
+                    "create veth pair: interface {} already exists on container namespace or {} exists on host namespace",
+                    data.container_interface_name, data.host_interface_name,
+                )
+            } else {
+                format!(
+                    "create veth pair: interface {} already exists on container namespace",
+                    data.container_interface_name
+                )
+            },
```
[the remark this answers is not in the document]
M1cha: agree, fixed

mheon:
LGTM

Luap99 (approved):
/lgtm

PR #1126, opened 2024-11-13
network: bridge: add support for unmanaged mode

### LINKED ISSUE #1090: [RFE] Support bridges created outside of netavark
Some users want to connect containers to a bridge that they have created outside of podman/netavark , e.g. so that VMs and containers can talk to each other.  According to the comments in https://github.com/containers/netavark/issues/868 we do not currently support this configuration:

> Our current code always assumes that we created the bridge and we will also remove it once all interfaces connected to the bridge are removed.

The patch in that PR does work, but only coincidentally due to implementation details, and it leaves some cruft behind such as tables/chains that are left empty due to no IPs.  We instead should explicitly support this configuration and not call the firewall code at all if the interface has no netavark-assigned IP.

M1cha:
While Linux doesn't support modes on bridges, we use this concept to let the user tell us if they want podman/netavark to own the bridge or not. L3 behaves the same way as before this commit. L2 requires the bridge to exist already, will not setup any sysctls or firewall rules on the host and will not delete the bridge once all containers left.

Fixes #1090

Related: https://github.com/containers/common/pull/2247

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -21,7 +21,7 @@ use super::{
     constants::{
         ISOLATE_OPTION_FALSE, ISOLATE_OPTION_STRICT, ISOLATE_OPTION_TRUE,
         NO_CONTAINER_INTERFACE_ERROR, OPTION_HOST_INTERFACE_NAME, OPTION_ISOLATE, OPTION_METRIC,
-        OPTION_MTU, OPTION_NO_DEFAULT_ROUTE, OPTION_VRF,
+        OPTION_MODE, OPTION_MTU, OPTION_NO_DEFAULT_ROUTE, OPTION_VRF,
     },
     core_utils::{self, get_ipam_addresses, join_netns, parse_option, CoreUtils},
     driver::{self, DriverInfo},
@@ -34,6 +34,14 @@ use super::{
 };
 
 const NO_BRIDGE_NAME_ERROR: &str = "no bridge interface name given";
+
+#[derive(Clone, Copy, PartialEq)]
+enum BridgeMode {
+    /// The bridge is managed by netavark.
+    Managed,
+    /// The bridge was created externally and we only add/remove veths.
+    Unmanaged,
+}
 
 struct InternalData {
     /// interface name of the veth pair inside the container netns
@@ -52,6 +60,8 @@ struct InternalData {
     isolate: IsolateOption,
     /// Route metric for any default routes added for the network
     metric: Option<u32>,
+    /// Management mode of the bridge.
+    mode: BridgeMode,
     /// if set, no default gateway will be added
     no_default_route: bool,
     /// sef vrf for bridge
@@ -82,6 +92,7 @@ impl driver::NetworkDriver for Bridge<'_> {
         }
         let ipam = get_ipam_addresses(self.info.per_network_opts, self.info.network)?;
 
+        let mode: Option<String> = parse_option(&self.info.network.options, OPTION_MODE)?;
         let mtu: u32 = parse_option(&self.info.network.options, OPTION_MTU)?.unwrap_or(0);
         let isolate: IsolateOption = get_isolate_option(&self.info.network.options)?;
         let metric: u32 = parse_option(&self.info.network.options, OPTION_METRIC)?.unwrap_or(100);
@@ -108,6 +119,7 @@ impl driver::NetworkDriver for Bridge<'_> {
             mtu,
             isolate,
             metric: Some(metric),
+            mode: get_bridge_mode_from_string(mode.as_deref())?,
             no_default_route,
             vrf,
         });
@@ -133,9 +145,11 @@ impl driver::NetworkDriver for Bridge<'_> {
             data.bridge_interface_name, data.ipam.gateway_addresses
         );
 
-        setup_ipv4_fw_sysctl()?;
-        if data.ipam.ipv6_enabled {
-            setup_ipv6_fw_sysctl()?;
+        if let BridgeMode::Managed = data.mode {
+            setup_ipv4_fw_sysctl()?;
+            if data.ipam.ipv6_enabled {
+                setup_ipv6_fw_sysctl()?;
+            }
         }
 
         let (host_sock, netns_sock) = netlink_sockets;
@@ -224,29 +238,29 @@ impl driver::NetworkDriver for Bridge<'_> {
             None
         };
 
-        // if the network is internal block routing and do not setup firewall rules
-        if self.info.network.internal {
-            CoreUtils::apply_sysctl_value(
-                format!(
-                    "/proc/sys/net/ipv4/conf/{}/forwarding",
-                    data.bridge_interface_name
-                ),
-                "0",
-            )?;
-            if data.ipam.ipv6_enabled {
+        if let BridgeMode::Managed = data.mode {
+            // if the network is internal block routing and do not setup firewall rules
+            if self.info.network.internal {
                 CoreUtils::apply_sysctl_value(
                     format!(
-                        "/proc/sys/net/ipv6/conf/{}/forwarding",
+                        "/proc/sys/net/ipv4/conf/{}/forwarding",
                         data.bridge_interface_name
                     ),
                     "0",
                 )?;
+                if data.ipam.ipv6_enabled {
+                    CoreUtils::apply_sysctl_value(
+                        format!(
+                            "/proc/sys/net/ipv6/conf/{}/forwarding",
+                            data.bridge_interface_name
+                        ),
+                        "0",
+                    )?;
+                }
+            } else {
+                self.setup_firewall(data)?
             }
-            // return here to skip setting up firewall rules
-            return Ok((response, aardvark_entry));
         }
-
-        self.setup_firewall(data)?;
 
         Ok((response, aardvark_entry))
     }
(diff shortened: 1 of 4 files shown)


src/network/bridge.rs
```diff
@@ -225,7 +237,7 @@ impl driver::NetworkDriver for Bridge<'_> {
         };
 
         // if the network is internal block routing and do not setup firewall rules
-        if self.info.network.internal {
+        if self.info.network.internal && data.mode == BridgeMode::L3 {
```
Luap99, line 240: if we wrap this entire part in `if let BridgeMode::L3 = data.mode {` up until the setup_firewall call we would only need to match once

src/network/bridge.rs
```diff
@@ -288,12 +305,14 @@ impl driver::NetworkDriver for Bridge<'_> {
             return Ok(());
         }
 
-        match self.teardown_firewall(complete_teardown, bridge_name) {
-            Ok(_) => {}
-            Err(err) => {
-                error_list.push(err);
-            }
-        };
+        if let BridgeMode::L3 = mode {
```
Luap99, line 308: Can we simplify this together with the self.info.network.internal condition into one and then remove the early return from internal

src/network/bridge.rs
```diff
@@ -550,6 +569,11 @@ fn create_interfaces(
                     // for all other errors we want to return the error
                     return Err(err).wrap("get bridge interface");
                 }
+
+                if let BridgeMode::L2 = data.mode {
+                    return Err(err).wrap("l2 bridge interface not found");
+                }
```
Luap99, line 575: I wonder if user will be running into https://github.com/containers/netavark/commit/730e1bffb6e0bf0fb1f597195454e581ea9804be if we do not create the bridge. There seems to be bad things going on while/adding removing new interfaces.

I think the error here should say l2 requires the bridge to already exist on the host

src/network/bridge.rs
```diff
@@ -703,41 +727,44 @@ fn create_veth_pair<'fd>(
...
+            if data.ipam.ipv6_enabled {
+                //  Disable dad inside the container too
+                let disable_dad_in_container = format!(
+                    "/proc/sys/net/ipv6/conf/{}/accept_dad",
+                    &data.container_interface_name
+                );
+                core_utils::CoreUtils::apply_sysctl_value(disable_dad_in_container, "0")?;
+            }
+            let enable_arp_notify = format!(
+                "/proc/sys/net/ipv4/conf/{}/arp_notify",
```
Luap99, line 742: I think we may want to keep arp_notify even on l2?

Luap99 (commented):
So this is something I did not thought about. What does it mean to be a internal network with l2? There seems to be fair bit of overlap between internal l3 and l2 (except for a few sysctls and the delete the bridge thing).

My understanding is that routing would not be able to blocked without using the firewall in the l2 case normally you would have your l2 ethX interface connected to the bridge and your LAN so there is no ip routing happing on the host system as all the packages will be send to the routed normally vie the LAN.

I guess there is no good reason to support this but I wonder if we should make podman (rather c/common) error out if internal and l2 is used

M1cha:
> So this is something I did not thought about. What does it mean to be a internal network with l2? There seems to be fair bit of overlap between internal l3 and l2 (except for a few sysctls and the delete the bridge thing).

It may just be me, but I always had trouble understanding what `--internal` does by just reading the documentation (or even just by it's name). The code is very simple though: Disable forwarding via sysctls, don't setup a firewall and don't add a default route. `(1)`

> I guess there is no good reason to support this but I wonder if we should make podman (rather c/common) error out if internal and l2 is used

No matter if podman "manages" the bridge or not, having podman configure addresses and routes can be useful. You can compare that with static IPs and routes inside the network-config of a physical host.

Also, podmans firewall is currently external to the container, so it can be compared to a managed l3 switch with an integrated firewall. That is definitely needed for both managed and external bridges, the question is just if there is a need to have podman setup the firewall for external bridges. My guess is: no. Either way, podman currently does not have an option to disable the firewall for a single network, that's why I disabled it for L2 in this PR.

> My understanding is that routing would not be able to blocked without using the firewall in the l2 case normally you would have your l2 ethX interface connected to the bridge and your LAN so there is no ip routing happing on the host system as all the packages will be send to the routed normally vie the LAN.

Without a firewall, the only way to block traffic going out the bridge to a different interface is by setting the forwarding sysctls to `0`, yes. However, you could control traffic between ports on a bridge using an nftables bridge filter(or ebtables). This is like a managed L2-switch with a firewall. Currently, netavark does not support this and it's one of many reasons why I needed configurable veth names, so I can do this outside of podman. It might make sense to add this in future though, especially since this is kinda what the `isolate` option sounds like - just more fine-grained.

Also, there doesn't always have to be a physical port on an external bridge. For example, you could simply want to put both containers and libvirt VMs on the same bridge so they can communicate directly, but you don't want them to access the internet or the physical local network. Or your containers are running on the router itself and you have to forward packets to the wan interface instead of l2-switching them through the bridge to a different port.

Given my arguments above, implementing the external bridge feature as l2 vs l3 mode might be confusing. In the end it's an API choice and will not affect the functionality, but you might want to reconsider it.

`(1)` Not setting up a firewall might be wrong if netavark/podman would ever support firewalling between containers on the bridge.

Luap99:
Yeah I am to sure about the naming either. My think was l3 is routed while 2 is local LAN but l2 is just whatever the user configured so you are right that that naming does not really reflect what is going on.

Would something `mode=managed/unmanaged` or `type=managed/unmanaged` better? Where managed is what we to today and unmanaged is what you do here. Then it is easy to documented that unmanaged means no firewall and sysctl settings AND ONLY veth pair + ip assignment to the container interface.

M1cha:
`mode=managed/unmanaged` sounds good to me, yes. We're already using `mode` for both macvlan and ipvlan (with different possible values) so reusing it makes sense to me. I also can't think of anything else we'd want to use `mode` for in future.

src/network/bridge.rs
```diff
@@ -133,9 +143,11 @@ impl driver::NetworkDriver for Bridge<'_> {
             data.bridge_interface_name, data.ipam.gateway_addresses
         );
 
-        setup_ipv4_fw_sysctl()?;
-        if data.ipam.ipv6_enabled {
-            setup_ipv6_fw_sysctl()?;
+        if let BridgeMode::L3 = data.mode {
```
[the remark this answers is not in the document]
M1cha: If podman passed an unsupported mode it would fail in `get_bridge_mode_from_string`. In case you mean "in case we add support for the new mode, but forget to handle it in all places", we need to use `match` everywhere. It might make some conditions more complicated but it would cause compile-errors.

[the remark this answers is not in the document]
Luap99: yes I meant we forget to handle it. However if we go unmanaged/managed for now I think it is safe enough that we do not add new modes often enough. And if we do  we need to check what mode should do what anyway so let's keep the readable form.

src/network/bridge.rs
M1cha: I don't know, all of these can actually be useful for both  managed and unmanaged mode. The question is just if we want to enforce our own defaults or rely on the linux kernel defaults. If we want to make this decision differently for each sysctl, we have to document why.

src/network/bridge.rs
```diff
@@ -35,6 +35,12 @@ use super::{
 
 const NO_BRIDGE_NAME_ERROR: &str = "no bridge interface name given";
 
+#[derive(Clone, Copy, PartialEq)]
+enum BridgeMode {
+    L2,
+    L3,
```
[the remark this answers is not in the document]
M1cha: done

src/network/bridge.rs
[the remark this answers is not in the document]
M1cha: makes sense. 

> Also now that I look at this we should really not set the global sysctl for internal networks here (feel free to fix this in a second commit or I put it on my list)

done

src/network/bridge.rs
M1cha: good idea, done

src/network/bridge.rs
M1cha: good idea, done

test/620-bridge-mode.bats
```diff
@@ -0,0 +1,42 @@
...
+    assert_json "$link_info" '.[].flags[] | select(.=="UP")' == "UP" "Host bridge interface is up"
+
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-managed.json teardown $(get_container_netns_path)
+
+    # check if the interface gets removed
+    expected_rc=1 run_in_host_netns ip -j --details link show podman0
+    assert "$output" "==" 'Device "podman0" does not exist.'
+}
+
+@test bridge - l2 mode {
```
Luap99, line 22: test need to be changed to unmanged

test/620-bridge-mode.bats
```diff
@@ -0,0 +1,42 @@
...
+
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-unmanaged.json setup $(get_container_netns_path)
+
+    run_in_host_netns ip -j --details link show brtest0
+    link_info="$output"
+    assert_json "$link_info" '.[].flags[] | select(.=="UP")' == "UP" "Host bridge interface is up"
+
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-unmanaged.json teardown $(get_container_netns_path)
+
+    # check if the interface gets removed
```
Luap99, line 37: this is a bit confusion, it should say that we ensure the bridge is not removed.

Luap99 (commented):
code seems good, two minor nits on the test then LGTM

M1cha:
thanks, fixed

Luap99 (approved):
LGTM

Luap99:
@mheon PTAL

Issue #676, opened 2023-04-19
DHCP Proxy Does Not Report Option 12 (Hostname)


AnastasiyaSoyka:
Currently, the DHCP proxy does not report option 12 (client hostname) to the DHCP server when requesting a DHCP lease. This means that routers which function also as a DNS server (most consumer routers), and use that information in order to create an A record pointing to the DHCP client, will not function.

This issue does not affect the core DHCP proxy functionality, but instead represents a quality-of-life issue.

I would be willing to submit a PR for this, but I'm not familiar with the codebase; could someone please point me in the direction of the necessary changes?

Luap99:
It should not be that complicated, however as of of today netavark does not know the host name of a container so the first step would be to send the Hostname from podman to netavark.
The type for podman is defined here: https://github.com/containers/common/blob/536bbe5e7e21d8e2087eee9615f3b3c995955840/libnetwork/types/network.go#L233
Then you need to make podman set this new struct field with the correct value, likely here: https://github.com/containers/podman/blob/911be1cbcb0ad654a8e10666262bdcee50ee54e0/libpod/networking_common.go#L47

Change the type here in netavark to read the new field. Then add it to the grpc type so you can send it to the dhcp-proxy and then you need to figure out how to set this field in for the dhcp call. I checked at it seems the our DHCP library seems to support the Hostname option so it should be possible.

That said I definitely wouldn't call this easy which of course depends on your expertise. The proxy code is definitely not very "clean", so figuring out how it works is not as trivial as it should be.

agorgl:
Waiting for this too!

SilverBut:
Hi @Luap99, I have an implementation on [my dev branch](https://github.com/SilverBut/netavark/tree/dev). Should I make a PR?

But I'm changing the proto, and maybe this feature should only be enabled if explicitly asked. Please suggest some idea.

Aaron-Rumpler:
@Luap99 and @SilverBut Any updates on this?

Luap99:
No, this is waiting for someone to propose a PR. I don't have time for it.

SilverBut:
Seems my dev branch have not updated for a while. Let me see what I can do.

briankoco:
Using the container name as the `dhcp_host_name` seems like a reasonable default absent an explicit hostname request. From a podman perspective, interpreting the container name as a DNS hostname also seems reasonable, at least per this comment:
[...]

PR #1130, opened 2024-11-25
Add container hostname to DHCP requests and use container id as client id

gtjoseph:
Currently, since a container's hostname isn't sent to netavark, we can't
include it in DHCP requests when using a macvlan network and DHCP IPAM
driver. Therefore, if DNS records need to be added for a container, it needs
to be done by external scripts.  Another issue is that since every time a
container starts it gets a new random mac address and since, in the absense
of a client id, the mac address is used by DHCP servers to keep track of
leases, it's likely that the container will get a new ip address on restart.
If the container is providing an exposed service, even just ssh, clients may
have the old address cached and may not be able to reach the container.

This commit accepts the container's hostname on incoming JSON setup commands
and adds it to the DHCP client configuration so an Option 12
(Hostname) parameter can be sent in DHCP messages. This should allow DDNS
updates of DNS resords. This is dependent on corresponding commits in Podman
and common to actually pass the host name of course.  This commit also sets
the Option 61 (Client identifier) to the container's ID.  Since this stays
constant across container restarts, it's likely the DHCP server will use the
existing lease and return the same IP address.

* Added "container_hostname" to network::types::NetworkOptions. The setup
  command will set this from the incoming JSON and pass it through ultimately
  to dhcp_service::DhcpV4Service via macvlan_dhcp.get_dhcp_lease().

* Added "container_id" to proxy.proto so it can also be passed to
  dhcp_service::DhcpV4Service and used to set the client id.

* Added tests to test-dhcp/002-setup.bats to check that hostname is passed
  correctly.

Fixes: #676 
Signed-off-by: George Joseph <g.devel@wxy78.net>

The new functionality requires the following PRs but netavark will build and run successfully with existing functionality without them.
containers/podman#24675
containers/common#2254

This PR DOES however require the following mozim PR to set the client id:
nispor/mozim#39

### DIFF
--- a/src/commands/teardown.rs
+++ b/src/commands/teardown.rs
@@ -96,6 +96,7 @@ impl Teardown {
                     firewall: firewall_driver.as_ref(),
                     container_id: &network_options.container_id,
                     container_name: &network_options.container_name,
+                    container_hostname: &network_options.container_hostname,
                     container_dns_servers: &network_options.dns_servers,
                     netns_host: hostns.file.as_fd(),
                     netns_container: netns.file.as_fd(),
--- a/src/dhcp_proxy/dhcp_service.rs
+++ b/src/dhcp_proxy/dhcp_service.rs
@@ -49,6 +49,30 @@ impl DhcpV4Service {
     pub fn new(nc: NetworkConfig, timeout: u32) -> Result<Self, DhcpServiceError> {
         let mut config = DhcpV4Config::new_proxy(&nc.host_iface, &nc.container_mac_addr);
         config.set_timeout(timeout);
+
+        // Sending the hostname to the DHCP server is optional but it can be useful
+        // in environments where DDNS is used to create or update DNS records.
+        if !nc.host_name.is_empty() {
+            config.set_host_name(&nc.host_name);
+        };
+
+        // DHCP servers use the "client id", which is usually the MAC address,
+        // to keep track of leases but each time the container starts, it gets
+        // a new, random, MAC address so there's a good chance that the container
+        // won't get the same IP address if it restarts. This can be an issue if
+        // a container provides a service and needs to be restarted because, even
+        // if DDNS is in use and the container has a DNS A record, a client may
+        // still have the old IP address cached until the DNS TTL expires.
+        //
+        // Since the container id remains constant for life of the container
+        // and it should be globally unique, we can use it as the client id to
+        // ensure the container gets the same IP address each time it starts.
+
+        // The client id is a byte array so we need to convert the container id
+        // to a byte array.  The client_id_type of "0" means the client id
+        // is not a hardware address.
+        config.set_client_id(0, nc.container_id.as_bytes());
+
         let client = match DhcpV4ClientAsync::init(config, None) {
             Ok(client) => Ok(client),
             Err(err) => Err(DhcpServiceError::new(InvalidArgument, err.to_string())),
--- a/src/network/types.rs
+++ b/src/network/types.rs
@@ -66,9 +66,13 @@ pub struct NetworkOptions {
     #[serde(rename = "container_id")]
     pub container_id: String,
 
-    /// The container name, used as dns name.
+    /// The container name.
     #[serde(rename = "container_name")]
     pub container_name: String,
+
+    /// The container hostname.
+    #[serde(rename = "container_hostname")]
+    pub container_hostname: Option<String>,
 
     /// The options used to create the interfaces with.
     /// The networks listed in "network_info" have to match this,
--- a/src/proto/proxy.proto
+++ b/src/proto/proxy.proto
@@ -16,6 +16,7 @@ message NetworkConfig {
   string host_name = 5;
   Version version = 6;
   string ns_path = 7;
+  string container_id = 8;
 
 }
 // Lease can either contain a IPv4 or IPv6 DHCP lease, and the common IP information
--- a/src/commands/setup.rs
+++ b/src/commands/setup.rs
@@ -79,6 +79,7 @@ impl Setup {
                     firewall: firewall_driver.as_ref(),
                     container_id: &network_options.container_id,
                     container_name: &network_options.container_name,
+                    container_hostname: &network_options.container_hostname,
                     container_dns_servers: &network_options.dns_servers,
                     netns_host: hostns.file.as_fd(),
                     netns_container: netns.file.as_fd(),
--- a/src/dhcp_proxy/types.rs
+++ b/src/dhcp_proxy/types.rs
@@ -23,6 +23,7 @@ impl FromStr for NetworkConfig {
             version: 0,
             ns_path: "".to_string(),
             container_iface: "".to_string(),
+            container_id: "".to_string(),
         })
     }
 }
--- a/src/network/driver.rs
+++ b/src/network/driver.rs
@@ -29,6 +29,7 @@ pub struct DriverInfo<'a> {
     pub dns_port: u16,
     pub config_dir: &'a Path,
     pub rootless: bool,
+    pub container_hostname: &'a Option<String>,
 }
 
 pub trait NetworkDriver {
--- a/src/network/macvlan_dhcp.rs
+++ b/src/network/macvlan_dhcp.rs
@@ -33,17 +33,19 @@ pub fn get_dhcp_lease(
     container_network_interface: &str,
     ns_path: &str,
     container_macvlan_mac: &str,
+    container_hostname: &str,
+    container_id: &str,
 ) -> NetavarkResult<DhcpLeaseInfo> {
     let nvp_config = NetworkConfig {
         host_iface: host_network_interface.to_string(),
         // TODO add in domain name support
         domain_name: "".to_string(),
-        //  TODO add in host name support
-        host_name: "".to_string(),
+        host_name: container_hostname.to_string(),
         version: 0,
         ns_path: ns_path.to_string(),
         container_iface: container_network_interface.to_string(),
         container_mac_addr: container_macvlan_mac.to_string(),
+        container_id: container_id.to_string(),
     };
     let lease = match tokio::task::LocalSet::new().block_on(
         match &tokio::runtime::Builder::new_current_thread()
@@ -127,12 +129,12 @@ pub fn release_dhcp_lease(
         host_iface: host_network_interface.to_string(),
         // TODO add in domain name support
         domain_name: "".to_string(),
-        //  TODO add in host name support
         host_name: "".to_string(),
         version: 0,
         ns_path: ns_path.to_string(),
         container_iface: container_network_interface.to_string(),
         container_mac_addr: container_macvlan_mac.to_string(),
+        container_id: "".to_string(),
     };
     match tokio::task::LocalSet::new().block_on(
         match &tokio::runtime::Builder::new_current_thread()
--- a/src/network/vlan.rs
+++ b/src/network/vlan.rs
@@ -188,6 +188,8 @@ impl driver::NetworkDriver for Vlan<'_> {
                 &data.container_interface_name,
                 self.info.netns_path,
                 &container_vlan_mac,
+                self.info.container_hostname.as_deref().unwrap_or(""),
+                self.info.container_id,
             )?;
             // do not overwrite dns servers set by dns podman flag
             if !self.info.container_dns_servers.is_some() {
--- a/test-dhcp/002-setup.bats
+++ b/test-dhcp/002-setup.bats
@@ -15,7 +15,8 @@ load helpers
   "domain_name": "example.com",
   "host_name": "foobar",
   "version": 0,
-  "ns_path": "$NS_PATH"
+  "ns_path": "$NS_PATH",
+  "container_id": "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
 }
   \0
 EOF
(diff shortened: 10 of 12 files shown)


mheon:
Sure, LGTM

src/dhcp_proxy/dhcp_service.rs
```diff
@@ -49,6 +49,11 @@ impl DhcpV4Service {
     pub fn new(nc: NetworkConfig, timeout: u32) -> Result<Self, DhcpServiceError> {
         let mut config = DhcpV4Config::new_proxy(&nc.host_iface, &nc.container_mac_addr);
         config.set_timeout(timeout);
+
+        if !nc.host_name.is_empty() {
+            config.set_host_name(&nc.host_name);
+            config.use_host_name_as_client_id();
```
Luap99, line 55: Is there a reason to set the host name as client id? I have not looked at other DHCP clients if they do that by default but I would assume just setting the host name alone should be good enough for server to consider the name?

src/network/types.rs
```diff
@@ -89,6 +89,13 @@ pub struct NetworkOptions {
     /// Custom DNS servers for aardvark-dns.
     #[serde(rename = "dns_servers")]
     pub dns_servers: Option<Vec<IpAddr>>,
+
+    /// The container hostname.
+    #[serde(
+        rename = "container_hostname",
+        default = "NetworkOptions::default_hostname"
+    )]
+    pub container_hostname: String,
```
Luap99, line 98: Can you move this field up below container_name as they are rather similar.

Second, instead of defining a default function you could use `#[serde(default)]`. However given this is optional I rather instead wrap the type in `Option<>` to make this more clear. Then on the caller side you can do
`container_hostname.unwrap_or("")`

src/dhcp_proxy/dhcp_service.rs
gtjoseph: DHCP servers look for existing leases by client id which is normally derived from the MAC address.  By default though a random MAC address gets assigned every time you (re)start a container so if you stop and start a container in quick succession, you could get a different IP address. Because of DNS TTL, hosts could have the old address cached and possibly not be able to reach the container.  By using the hostname as the client id, you _should_ be able to restart a container and get the same IP address.

src/network/types.rs
gtjoseph: > Can you move this field up below container_name as they are rather similar.

Done.  Decades of "C" development compel me to put new fields at the end of structures. :)




> Second, instead of defining a default function you could use `#[serde(default)]`. However given this is optional I rather instead wrap the type in `Option<>` to make this more clear. Then on the caller side you can do `container_hostname.unwrap_or("")`

This is my first practical exposure to Rust so I _think_ I did this correctly.  I did test it with current podman to make sure it didn't choke if container_hostname wasn't present in the JSON.

src/network/mod.rs
```diff
@@ -32,4 +32,8 @@ impl types::NetworkOptions {
         }?;
         Ok(opts)
     }
+
+    pub fn default_hostname() -> String {
+        "".to_string()
+    }
```
Luap99, line 38: This should be removed

src/dhcp_proxy/dhcp_service.rs
Luap99: Yes that makes sense thanks.

src/commands/teardown.rs
```diff
@@ -106,6 +106,7 @@ impl Teardown {
                     dns_port,
                     config_dir: Path::new(&config_dir),
                     rootless,
+                    container_hostname: &network_options.container_hostname,
```
Luap99, line 109: oh and also move this up here below container_name to keep it more readable.

Luap99 (commented):
Also we should have a test for it. Currently we only have some basic dhcp tests in test-dhcp/ but we need to cover both case at least (with and without hostname)

The `basic setup` test sets `"host_name": "foobar"` so we should verify the hostname was passed. We should be able to read the dnsmasq log assuming it logs the hostname.

But then we also should have a second case with an empty hostname so we know we keep no hostname working as well.

src/dhcp_proxy/dhcp_service.rs
gtjoseph: @SilverBut pointed out a potential issue here... https://github.com/containers/netavark/issues/676#issuecomment-2501696991

I'm going to re-think a bit.

gtjoseph:
I'm marking this as draft until I can think of a better way to handle client id.

src/dhcp_proxy/dhcp_service.rs
gtjoseph: I've made the client id the first 16 characters of the container ID which should resolve the concerns.  The way I did it however is a workaround because mozim doesn't provide a way to explicitly set the client id on DhcpV4Config.  I'm going to see if I can get that added so we don't need the workaround.

gtjoseph:
HA!  I just discovered a bug in mozim.  If you don't set `use_host_name_as_client_id()`, it doesn't send the host name at all.  I'll be submitting a PR to them to fix that and add a `set_client_id()` method.

src/proto/proxy.proto
```diff
@@ -16,6 +16,7 @@ message NetworkConfig {
   string host_name = 5;
   Version version = 6;
   string ns_path = 7;
+  string container_id = 8;
```
Luap99, line 19: I have not tested this as we never made any changes to the over the wire format but one thing that MUST be considered is that the dhcp proxy is a long running service. This means a new client sending a new format must still work with the old service running. So it would be important to know if the new field would cause the old service to fail.

src/dhcp_proxy/dhcp_service.rs
```diff
@@ -49,6 +49,25 @@ impl DhcpV4Service {
     pub fn new(nc: NetworkConfig, timeout: u32) -> Result<Self, DhcpServiceError> {
         let mut config = DhcpV4Config::new_proxy(&nc.host_iface, &nc.container_mac_addr);
         config.set_timeout(timeout);
+
+        if !nc.host_name.is_empty() {
+            let cid: String = nc.container_id.chars().take(16).collect();
```
Luap99, line 54: seems like there is no reason to limit to 16 here, more chars the less likely we have a conflict with other systems on the network. It seems like the client ID filed can be up to 63 chars?

src/proto/proxy.proto
gtjoseph: Ah, I hadn't thought of that. Easily tested though.

src/dhcp_proxy/dhcp_service.rs
gtjoseph: I limited it to 16 because that's the size of a UUID which is a common client id.  No reason not to expand it though.  Like every one of the hundreds of RFCs I've read over the years, [RFC2132 section 9.14](https://datatracker.ietf.org/doc/html/rfc2132#section-9.14) doesn't tell you what you _really_ want to know.  It says the minimum length is 2 but it doesn't specify a maximum length.  It does mention using an FQDN as an example and since a FQDN can be 253 bytes, a 64 byte container ID should be fine.

gtjoseph:
I'm setting this back to draft.  I've submitted PR nispor/mozim#39 to mozim to fix the hostname issue and add a set_client_id() method so I'd like to see that get merged first.  This way we don't need the workaround for setting the client id.

src/proto/proxy.proto
gtjoseph: Works fine...
* Stopped all containers and let dhcp-proxy stop after 30 seconds.
* Reinstalled podman and netavark from Fedora packages.
* Created and started a new container with macvlan/dhcp.
* No issues.  As expected, container did not get hostname as container name and neither client id nor hostname were sent to DHCP server.
* Confirmed dhcp-proxy was running.
* Installed netavark built from PR.
* Verified that the dhcp-proxy was still running with the original PID.
* Created and started a new container with macvlan/dhcp.
* No issues.  Again, as expected, container did not get hostname as container name and neither client id nor hostname were sent to DHCP server.
* Installed podman built from my PR with containers still running.
* Created and started another new container with macvlan/dhcp.
* No issues.  As expected, container now got hostname as container name and neither client id nor hostname were sent to DHCP server.
* Stopped all containers and let dhcp-proxy stop itself after 30 seconds.
* Restarted the previous container.
* As expected, container got hostname as container name and both client id and hostname were sent to DHCP server.

gtjoseph:
The last force push just updated the commit message.  No code changes.

gtjoseph:
@Luap99 When I run `make validate` locally I get tons of the following errors.  Any idea on how to get validate to work?

```
error: the `-Z unstable-options` flag must also be passed to enable the flag `check-cfg`
error: could not compile `unicode-ident` (lib)
error: the `-Z unstable-options` flag must also be passed to enable the flag `check-cfg`
error: could not compile `autocfg` (lib)
```

Fedora 41 with...
```
cargo.x86_64                        1.82.0-1.fc41 updates
cargo-rpm-macros.noarch             26.3-3.fc41   fedora
cargo2rpm.noarch                    0.1.18-1.fc41 fedora
rust.x86_64                         1.82.0-1.fc41 updates
rust-analyzer.x86_64                1.82.0-1.fc41 updates
rust-debugger-common.noarch         1.82.0-1.fc41 updates
rust-doc.x86_64                     1.82.0-1.fc41 updates
rust-gdb.noarch                     1.82.0-1.fc41 updates
rust-lldb.noarch                    1.82.0-1.fc41 updates
rust-rpm-sequoia-debugsource.x86_64 1.6.0-1.fc39  <unknown>
rust-src.noarch                     1.82.0-1.fc41 updates
rust-srpm-macros.noarch             26.3-3.fc41   <unknown>
rust-std-static.x86_64              1.82.0-1.fc41 updates
rust2rpm.noarch                     26.2.0-1.fc41 updates
rust2rpm-helper.x86_64              0.1.6-1.fc41  updates
rustfmt.x86_64                      1.82.0-1.fc41 updates
```

gtjoseph:
@Luap99 Just FYI...  I had to switch to the nightly rust toolchain to get `make validate` to work.  However, the 2024-11-26 nightly found issues in netavark stuff I didn't touch which may indicate that if the CI toolchain is updated it may catch issues it didn't before.  Just a heads up.

```
rustc 1.83.0 (90b35a623 2024-11-26)
cargo 1.83.0 (5ffbef321 2024-10-29)
```

```
error: elided lifetime has a name
   --> src/network/bridge.rs:337:40
    |
330 | impl<'a> Bridge<'a> {
    |      -- lifetime `'a` declared here
...
337 |     ) -> NetavarkResult<(SetupNetwork, PortForwardConfig)> {
    |                                        ^^^^^^^^^^^^^^^^^ this elided lifetime gets resolved as `'a`
    |
    = note: `-D elided-named-lifetimes` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(elided_named_lifetimes)]`

error: empty line after doc comment
  --> src/firewall/state.rs:20:1
   |
20 | / ///                 - ports/$netID_$conID -> port config
21 | |
   | |_
22 |   const FIREWALL_DIR: &str = "firewall";
   |   ------------------------ the comment documents this constant
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#empty_line_after_doc_comments
   = note: `-D clippy::empty-line-after-doc-comments` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::empty_line_after_doc_comments)]`
   = help: if the empty line is unintentional remove it

error: the following explicit lifetimes could be elided: 'a
  --> src/firewall/varktables/types.rs:85:6
   |
85 | impl<'a> VarkChain<'a> {
   |      ^^            ^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
   = note: `-D clippy::needless-lifetimes` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::needless_lifetimes)]`
help: elide the lifetimes
   |
85 - impl<'a> VarkChain<'a> {
85 + impl VarkChain<'_> {
   |

error: could not compile `netavark` (lib) due to 3 previous errors

```

Luap99:
New lint fixes for rust v1.83 are already in https://github.com/containers/netavark/pull/1134

Luap99:
> @Luap99 Do you guys have a policy on how to address upstream projects (nispor/mozim in this case) that aren't responsive? Think I should go back to my workaround for setting client id? It was...
> 
> ```
> config.set_host_name(container_id);
> config.use_host_name_as_client_id();
> config.set_host_name(host_name);
> ```

We have to see, for now we all will be on PTO over the holidays and we are most likely aiming for our next release end of January/start of February.  So until then there is plenty of time for him to react and merge it.

We don't have any rules about it, often the answer is wait and hope. However in this case the maintainer is also working for Red Hat and we have been working with him in the past. I will send him an email and ask him to have a look. Maybe he just missed the notification.

So for now just wait and we can see in January if we need to use the work around.

gtjoseph:
Sounds good, thanks.

Luap99:
I merged https://github.com/containers/netavark/pull/1145 which includes your fixes

gtjoseph:
Thanks for the assist @Luap99 !
The containers/common#2254 PR has already merged so I've unDRAFTed this PR as it should be good to go.  I guess I'll wait for this PR to merge before unDRAFTing the containers/podman#24675 PR.

Luap99 (approved):
/lgtm

Thank you

PR #1142, opened 2024-12-10
cargo: set rust-version

Luap99:
rust 1.83 starts to upgrade the Cargo.lock version field to version 4,
however version 4 support was only added in 1.78 which is newer than our
MSRV 1.76. To ensure cargo respects the MSRV we need to set in the in
Cargo.toml.

And also some dep updates from renovate so I don't have to rebase all the PRs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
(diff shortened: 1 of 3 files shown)


Luap99:
@baude @mheon PTAL

mheon:
LGTM

PR #1144, opened 2024-12-13
generate protobuf in OUT_DIR

### LINKED ISSUE #1137: cargo publish issue
```
Caused by:
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
  Added: /home/pholzing/CODE/netavark/targets/package/netavark-1.13.1/src/proto-build/netavark_proxy.rs
```

I pushed v1.13.1 with --no-verify now but for the next release we should fix that.

Luap99:
We should not create build time generated files under src/, with the latest rust version this seems to cause problems on cargo publish.

Instead just default to the OUT_DIR env that is always set by cargo.

Fixes #1137

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,5 @@ targets/
 netavark.1
 vendor/
 .idea/*
-src/proto-build/netavark_proxy.rs
 contrib/systemd/*/*.service
 .vscode*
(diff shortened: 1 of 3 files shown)


Luap99:
@baude @mheon PTAL

PR #1165, opened 2025-01-21
Add handling for firewalld's StrictForwardPorts setting

mheon:
This mostly amounts to throwing a sensible error in cases where the setting is enabled and the user has requested ports be forwarded.

There are no tests at present because firewalld 2.3.x is currently restricted to Rawhide, and is required for effective testing.

A manpage has also been added with details on how to work with the StrictForwardPorts setting (as well as some other tidbits on firewalld + Podman integration).

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -4,9 +4,12 @@ MANDIR ?= $(DATADIR)/man
 GO ?= go
 GOMD2MAN ?= go-md2man
 
-docs: $(patsubst %.md,%,$(wildcard *.1.md))
+docs: $(patsubst %.md,%,$(wildcard *.[0-9].md))
 
 %.1: %.1.md
+	$(GOMD2MAN) -in $^ -out $@
+
+%.7: %.7.md
 	$(GOMD2MAN) -in $^ -out $@
 
 .PHONY: .install.md2man
@@ -17,6 +20,8 @@ docs: $(patsubst %.md,%,$(wildcard *.1.md))
 install:
 	install -d ${DESTDIR}/${MANDIR}/man1
 	install -m 0644 *.1 ${DESTDIR}/${MANDIR}/man1
+	install -d ${DESTDIR}/${MANDIR}/man7
+	install -m 0644 *.7 ${DESTDIR}/${MANDIR}/man7
 
 .PHONY: clean
 clean:
--- a/docs/netavark-firewalld.7.md
+++ b/docs/netavark-firewalld.7.md
@@ -1,0 +1,90 @@
+% NETAVARK-FIREWALLD 7 Netavark Firewalld Interactions Man Page
+% Matthew Heon
+% January 2025
+
+## Name
+
+netavark-firewalld - description of the interaction of Netavark and firewalld
+
+## Description
+
+Netavark can be used on systems with firewalld enabled without issue.
+When using the default `nftables` or `iptables` firewall drivers, on systems where firewalld is running, firewalld will automatically be configured to allow connectivity to Podman containers.
+All subnets of Podman-managed networks will be automatically added to the `trusted` zone to allow this access.
+
+### Firewalld Driver
+
+There is also a dedicated firewalld driver in Netavark.
+This driver uses the firewalld DBus API to natively interact with firewalld.
+It can be enabled by setting `firewall_driver` to `firewalld` in `containers.conf`.
+The native firewalld driver offers better integration with firewalld, but presently suffers from several limitations.
+It does not support isolation (the `--opt isolate=` option to `podman network create`.
+Connections to ports forwarded by a container on the same host can only be made through the IPv4 localhost IP (`127.0.0.1`).
+Using other IPs on the host will not work, unless the connection comes from a separate host.
+
+### Strict Port Forwarding
+
+Since firewalld version 2.3.0, a new setting, `StrictForwardPorts`, has been added.
+The setting is located in `/etc/firewalld/firewalld.conf` and defaults to `no` (disabled).
+When disabled, port forwarding with Podman works as normal.
+When it is enabled (set to `yes`), port forwarding with root Podman will become nonfunctional.
+Attempting to start a container or pod with the `-p` or `-P` options will return errors.
+When StrictForwardPorts is enabled, all port forwarding must be done through firewalld using the firewall-cmd tool.
+This ensures that containers cannot allow traffic through the firewall without administrator intervention.
+Please note that rootless Podman is unaffected by this setting, and will function as it always has.
+
+Instead, containers should be started without forwarded ports specified, and preferably with static IPs.
+
+To forward a port externally, the following command should be run, substituting the desired host and container port numbers, protocol, and the container's IP.
+```
+# firewall-cmd --permanent --zone {ZONE} --add-forward-port=port={HOST_PORT_NUMBER}:proto={PROTOCOL}:toport={CONTAINER_PORT_NUMBER}:toaddr={CONTAINER_IP}
+```
+
+If you are not sure which zone to use, the `public` zone should always work.
+
+If the container does not have a static IP, it can be found with `podman inspect`:
+```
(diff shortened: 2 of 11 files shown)


mheon:
@Luap99 PTAL, particularly at the manpage

docs/Makefile
```diff
@@ -4,11 +4,14 @@ MANDIR ?= $(DATADIR)/man
...
 GOMD2MAN ?= go-md2man
 
-docs: $(patsubst %.md,%,$(wildcard *.1.md))
+docs: $(patsubst %.md,%,$(wildcard *.[0-9].md))
 
 %.1: %.1.md
 	$(GOMD2MAN) -in $^ -out $@
 
+%.5: %.5.md
+	$(GOMD2MAN) -in $^ -out $@
```
Luap99, line 13: properly should look into de-duplicating this with .1 one. Though not a blocker

docs/netavark-firewalld-forwarding.5.md
```diff
@@ -0,0 +1,85 @@
+% NETAVARK-FIREWALLD 5 Netavark Firewalld Interactions Man Page
```
Luap99, line 1: The name here is different from the actual file name `netavark-firewalld-forwarding` NETAVARK-FIREWALLD seems good enough so I suggest you rename the file.

Second, section 5 is wrong. That is for file formats/config files per [man-pages(7)](https://man7.org/linux/man-pages/man7/man-pages.7.html). IT should be in section 7 based on that.

docs/netavark-firewalld-forwarding.5.md
```diff
@@ -0,0 +1,85 @@
...
+All subnets of Podman-managed networks will be automatically added to the `trusted` zone to allow this access.
+
+### Firewalld Driver
+
+There is also a dedicated firewalld driver in Netavark.
+This driver uses the firewalld DBus API to natively interact with firewalld.
+It can be enabled by setting `firewall_driver` to `firewalld` in `containers.conf`.
+The native firewalld driver offers better integration with firewalld, but presently suffers from several limitations.
+It does not support isolation (the `--opt isolate=` option to `podman network create`.
+It also does not currently support connections to a forwarded port of a container on the same host via public IP, only via the IPv4 localhost IP (127.0.0.1).
```
Luap99, line 22: given most users are behind A NAT (for ipv4 at least) `public IP` may be misunderstood to refer to the public internet ip.

Maybe the wording should be reversed,i.e.: 

> When initiating a connection from the host system to the forwarded port it ONLY works when connecting to the IPv4 localhost (127.0.0.1). Using the host ips of other interface will not work. However when the connection comes from an external host it will work.

That raises the question about 127.0.0.0/8 ips. Users can do something like `-p 127.0.0.10:8080:8080` to bind another localhost ip? Does that still work or is only 127.0.0.1 special?

docs/netavark-firewalld-forwarding.5.md
```diff
@@ -0,0 +1,85 @@
...
+When it is enabled (set to `yes`), port forwarding with Podman will become nonfunctional.
+Attempting to start a container or pod with the `-p` or `-P` options will return errors.
+When StrictForwardPorts is enabled, all port forwarding must be done through firewalld using the firewall-cmd tool.
+This ensures that containers cannot allow traffic through the firewall without administrator intervention.
+
+Instead, containers should be started without forwarded ports specified, and preferably with static IPs.
+
+To forward a port externally, the following command should be run, substituting the desired host and container port numbers, protocol, and the container's IP.
+```
+# firewall-cmd --permanent --zone public --add-forward-port=port={HOST_PORT_NUMBER}:proto={PROTOCOL}:toport={CONTAINER_PORT_NUMBER}:toaddr={CONTAINER_IP}
```
Luap99, line 38: I guess the zone here should be a variable? There is no need to use the public zone if users don't need it.

docs/netavark-firewalld-forwarding.5.md
```diff
@@ -0,0 +1,85 @@
...
+### Firewalld Driver
+
+There is also a dedicated firewalld driver in Netavark.
+This driver uses the firewalld DBus API to natively interact with firewalld.
+It can be enabled by setting `firewall_driver` to `firewalld` in `containers.conf`.
+The native firewalld driver offers better integration with firewalld, but presently suffers from several limitations.
+It does not support isolation (the `--opt isolate=` option to `podman network create`.
+It also does not currently support connections to a forwarded port of a container on the same host via public IP, only via the IPv4 localhost IP (127.0.0.1).
+
+### Strict Port Forwarding
```
Luap99, line 24: We should note that this only applies to rootful podman. Rootless binds the port on the host and needs a regular accept port rule like it always did.

src/firewall/firewalld.rs
```diff
@@ -706,3 +706,51 @@ pub fn rm_firewalld_if_possible(net: &ipnet::IpNet) {
         ),
     };
 }
+
+/// Check whether firewalld's StrictForwardPorts setting is enabled.
+/// Returns false if firewalld is not installed, not running, or there is any
+/// error with the process.
+pub fn is_firewalld_strict_forward_enabled() -> bool {
+    let conn = match Connection::system() {
```
Luap99, line 714: looking at firewalld::add_firewalld_if_possible() it might be nice if to share the connecting and not open a new time for every single thing which seems super inefficient for no good reason.

We could call Connection::system() once and is_firewalld_running once as well and then just pass an `Option<Connection>` around.

docs/netavark-firewalld-forwarding.5.md
```diff
@@ -0,0 +1,85 @@
...
+```
+# firewall-cmd --permanent --policy localhostForward --remove-rich-rule='rule family=ipv4 destination address=127.0.0.0/8 forward-port port={HOST_PORT_NUMBER} protocol={PROTOCOL} to-port={CONTAINER_PORT_NUMBER} to-addr={CONTAINER_IP}'
+```
+
+The associated `localhostForward` policy does not need to be removed.
+
+Please also note that, at present, it is not possible to access forwarded ports of a container on the same host via the host's public IP; the rich rule above must be applied instead and access must be via 127.0.0.1, not the public IP of the host.
+
+Please note that the firewalld driver presently bypasses this protection, and will still allow traffic through the firewall when `StrictForwardPorts` is enabled without manual forwarding through `firewall-cmd`.
+This may be changed in a future release.
```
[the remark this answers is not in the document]
mheon: Reserving the right to do so; folks actually using the feature can tell us if they care.

mheon:
Comments should all be addressed

docs/netavark-firewalld.7.md
```diff
@@ -0,0 +1,90 @@
+% NETAVARK-FIREWALLD 5 Netavark Firewalld Interactions Man Page
```
Luap99, line 1: The number here needs to be updated too.

Luap99 (commented):
minor nit, rest LGTM

docs/netavark-firewalld.7.md
mheon: Damn it, fixed

Luap99 (approved):
LGTM

docs/netavark-firewalld.7.md
```diff
@@ -0,0 +1,90 @@
...
+### Strict Port Forwarding
+
+Since firewalld version 2.3.0, a new setting, `StrictForwardPorts`, has been added.
+The setting is located in `/etc/firewalld/firewalld.conf` and defaults to `no` (disabled).
+When disabled, port forwarding with Podman works as normal.
+When it is enabled (set to `yes`), port forwarding with root Podman will become nonfunctional.
+Attempting to start a container or pod with the `-p` or `-P` options will return errors.
+When StrictForwardPorts is enabled, all port forwarding must be done through firewalld using the firewall-cmd tool.
+This ensures that containers cannot allow traffic through the firewall without administrator intervention.
+Please note that rootless Podman is unaffected by this setting, and will function as it always has.
```
baude, line 34:
```suggestion
Please note that rootless Podman is unaffected by this setting and will function as it always has.
```

baude:
nits on man pages, code lgtm, and merging ... fix this in #885 if you like

/lgtm

PR #1180, opened 2025-02-15
Add outbound_addr to allow for SNAT instead of MASQ

lto-dev:
Added outbound_addr to the bridge driver. 

I created a bridge-snat plugin to use this feature and do SNAT instead of MASQUERADE on certain networks since i have multiple public ip's on same machine and i needed the traffic to go out on them and not on the gateway.
I can share the plugin as well if anyone wants it.


src/firewall/state.rs
```diff
@@ -272,8 +272,9 @@ mod tests {
             network_hash_name: "hash".to_string(),
             isolation: IsolateOption::Never,
             dns_port: 53,
+            outbound_addr: None,
         };
-        let net_conf_json = r#"{"subnets":["10.0.0.0/24"],"bridge_name":"bridge","network_id":"c2c8a073252874648259997d53b0a1bffa491e21f04bc1bf8609266359931395","network_hash_name":"hash","isolation":"Never","dns_port":53}"#;
+        let net_conf_json = r#"{"subnets":["10.0.0.0/24"],"bridge_name":"bridge","network_id":"c2c8a073252874648259997d53b0a1bffa491e21f04bc1bf8609266359931395","network_hash_name":"hash","isolation":"Never","dns_port":53,"outbound_addr":null}"#;
```
Luap99, line 277: This is a breaking change, you break the deserialize from older version as this json is stored on the fs and if netavark is updated in the meantime the firewall reload service will fail to parse that from older versions.

So this change must be reverted and you should instead and a new test case with a outbound_addr defined so we know both versions pass.

src/firewall/varktables/types.rs
```diff
@@ -192,6 +192,7 @@ pub fn create_network_chains(chains: Vec<VarkChain<'_>>) -> NetavarkResult<()> {
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
```
Luap99, line 195: it would likely be better to define a helper struct to pass the arguments

src/network/internal_types.rs
```diff
@@ -26,6 +26,8 @@ pub struct SetupNetwork {
     pub isolation: IsolateOption,
     /// port used for the dns server
     pub dns_port: u16,
+    /// outbound address for SNAT
```
Luap99, line 29: you should define
```
#[serde(skip_serializing_if = "Option::is_none")]
#[serde(default)]
```
for this field then the json parsing should work from the previous versions

src/network/types.rs
```diff
@@ -57,6 +57,10 @@ pub struct Network {
     /// Network DNS servers for aardvark-dns.
     #[serde(rename = "network_dns_servers")]
     pub network_dns_servers: Option<Vec<IpAddr>>,
+
+    /// outbound addr for this network.
+    #[serde(rename = "outbound_addr")]
+    pub outbound_addr: Option<IpAddr>,
```
Luap99, line 63: That is not the way to do it, no new fields on the network. This should be set as regual option in the options map.

Luap99 (changes requested):
please add proper integration test in `test/` for both iptables and nftables

I have not yet looked into the rules that are added and if they are correct


Also I think this would need to be added to the firewalld code as well cc @mheon

Luap99:
Note I do not have time to properly review the rules right now. However please make sure your commits make sense as individual unit. Currently they are pretty meaningless so I suggest you squash everything into one

src/firewall/iptables.rs
```diff
@@ -54,15 +55,15 @@ impl firewall::FirewallDriver for IptablesDriver {
...
-                    conn,
+                let config = NetworkChainConfig {
                     network,
-                    &network_setup.network_hash_name,
-                    is_ipv6,
-                    network_setup.bridge_name.clone(),
-                    network_setup.isolation,
-                    network_setup.dns_port,
-                );
+                    network_hash_name: network_setup.network_hash_name.clone(),
```
Luap99, line 60: this is cloning the network hash now which should not be necessary AFAICS. It is always good to avoid extra clones where possible. You should keep the hash &str defined (yes that means defining a lifetime one the struct)

src/network/bridge.rs
```diff
@@ -387,6 +388,10 @@ impl<'a> Bridge<'a> {
             network_hash_name: id_network_hash.clone(),
             isolation: isolate,
             dns_port: self.info.dns_port,
+            outbound_addr: parse_option::<IpAddr>(
+                &self.info.network.options,
+                OPTION_OUTBOUND_ADDR,
+            )?,
```
Luap99: options should be parsed as first thing in validate() to return errors on incorrect values early and not once we already created the interfaces as they would have to be cleaned up again.

mheon:
For firewalld, this would force all of our NAT to be through rich rules, as the standard port-forwarding rules don't allow for this.  Would be a pretty significant change if we wanted to make it.

Luap99:
> For firewalld, this would force all of our NAT to be through rich rules, as the standard port-forwarding rules don't allow for this. Would be a pretty significant change if we wanted to make it.

I guess we should throw an error if firewalld is used with this option. Given the likely low usage of firewalld users I don't think we need to implement this right now for this PR then.

lto-dev:
> For firewalld, this would force all of our NAT to be through rich rules, as the standard port-forwarding rules don't allow for this. Would be a pretty significant change if we wanted to make it.

i wouldn't know where to start with firewalld, never used it.

lto-dev:
> Should we not have two options then one for ipv4_outbound address and one for ipv6_outbound address instead then?

I was trying to mimic some of the docker behavior, i use ipv4. 
I'll have to take a look into ipv6 snat for this if you decide you want it. When i started the pr i was reading ipv6 nat is discouraged so i didn't approach it.

it looks like their bridge supports ipv6...or not, documentation is kinda conflicting
com.docker.network.bridge.host_binding_ipv4 | all IPv4 and IPv6 addresses | Default IP when binding container ports.

let me know if you want it and what names to use for it

Luap99:
I think for consistency it makes sense to have it assuming kernel wise is basically the same rule with ipv6? If it turns out to be more complicated then I am fine with leaving it.

Naming wise I think `outbound_addr4` and `outbound_addr6` maybe?

test/250-bridge-nftables.bats
```diff
@@ -1052,6 +1052,32 @@ function check_simple_bridge_nftables() {
...
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-outbound-addr6.json setup $(get_container_netns_path)
+
+    # Check that the nftables rules were created with SNAT for IPv6
+    run_in_host_netns nft list chain inet netavark nv_2f259bab_fd10-88-a--_nm64
+    assert "${lines[3]}" =~ "ip6 daddr != ff00::/8 snat ip6 to fd20:100:200::1"
+
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-outbound-addr6.json teardown $(get_container_netns_path)
+ 
+    # Check that the chain is removed
+    #expected_rc=1 run_in_host_netns nft list chain inet netavark nv_2f259bab_fd10-88-a--_nm64
```
Luap99: Is there a reason this is commented out?

Luap99 (commented):
mostly LGTM

test/250-bridge-nftables.bats
lto-dev: missed that one, i was debugging had a bunch of other nft list chains there. i'll correct it

Luap99 (approved):
LGTM @mheon PTAL

baude:
LGTN

PR #1181, opened 2025-02-15
Add outbound_addr to allow for SNAT instead of MASQ

lto-dev:


lto-dev:
duplicated by mistake

PR #1187, opened 2025-02-22
Zbus5

siretart:
Based on patch provided by NoisyCoil in
https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=1093690;filename=update-zbus.patch;msg=5

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2063,17 +2063,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "sha1"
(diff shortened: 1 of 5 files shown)


Luap99 (commented):
changes LGTM

This seems to require a MSRV bump to v1.77. I have no objections bumping the version so I can take care of that on Monday

Luap99:
https://github.com/containers/netavark/pull/1188

PR #1191, opened 2025-03-06
test/001-basic: Make commit test optional

ricardobranco777:
Some distros like openSUSE & Debian don't build netavark from git, so the commit information won't be available.
    
Make this test conditional when `NETAVARK_UPSTREAM` is set.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -22,6 +22,7 @@ env:
     AARDVARK_DNS_URL: "https://api.cirrus-ci.com/v1/artifact/github/containers/aardvark-dns/success/binary.zip?branch=${AARDVARK_DNS_BRANCH}"
     FEDORA_NETAVARK_AARCH64_AMI: "fedora-netavark-aws-arm64-${IMAGE_SUFFIX}"
     EC2_INST_TYPE: "t4g.xlarge"
+    NETAVARK_UPSTREAM: "1"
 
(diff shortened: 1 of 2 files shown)


danishprakash (commented):
This would help us quite a bit given that we run upstream tests for our podman* releases.

Could you please add more information in the commit message @ricardobranco777?

Luap99 (commented):
I think it makes sense but technically we loose test coverage upstream with this.

In this case it is only the commit sha so maybe it doesn't matter much but I would prefer if we could keep that. Maybe we should move this behind some optional, i.e. we could define a env var like NETAVARK_UPSTREAM and only when this is set we check the commit. Then we set this env in our upstream CI and you downstream would not have to worry about this env.

Luap99 (commented):
Thanks but that still doesn't let us run them in CI as the env is not set, please add it to the top env section in .cirrus.yml so we have the env in our CI set.

ricardobranco777:
> Thanks but that still doesn't let us run them in CI as the env is not set, please add it to the top env section in .cirrus.yml so we have the env in our CI set.

True. Fixed. Thanks for your patience.

Luap99 (approved):
LGTM

@mheon @baude PTAL

mheon:
Sure
/lgtm

PR #1193, opened 2025-03-07
.github: remove cirrus rerun action

Luap99:
As pointed out in buildah[1] the action is broken in bad ways where it can trigger 1000+ rerun wasting our cloud resources.

We have little flakes here and in general blindly re-rerunning is not a the best so just get rid of it for now.

[1] https://github.com/containers/buildah/issues/6035


baude:
LGTM

Luap99:
@mheon PTAL and merge this is causing errors now that the podman action was removed.

PR #1198, opened 2025-03-13
Add link to c/common contributor's guide

Honny1:
This PR adds a link to c/common contributor's guide.

Fixes: https://issues.redhat.com/browse/RUN-2321

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,7 @@ and [PRs](https://github.com/containers/netavark/pulls) tracking system.
 ## Plugins
 
 Netavark also supports executing external plugins, see [./plugin-API.md](./plugin-API.md).
+
+## [Contributing](https://github.com/containers/common/blob/main/CONTRIBUTING.md)
+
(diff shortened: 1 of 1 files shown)


Luap99 (commented):
Mhh, I mean yes sure a good part there is relevant PRs/ DCO sign off, etc... but I also wrote a ton of stuff about go there that are of no value here. And we have zero instructions on how to actual contribute here with rust so that seems a bit odd.

I understand that might be out of scope for your task but that is likely something that we should fix as well.
cc @baude @mheon

Honny1:
Good point, maybe creating a CONTRIBUTING.md without specific go things would work, but it would increase the maintenance cost. 

Same problem facing this PR:
- https://github.com/containers/aardvark-dns/pull/580

TomSweeneyRedHat:
We should have a CONTRIBUTING.md file, with at least contents similar to the https://github.com/containers/buildah/blob/main/SECURITY.md file.  If we decide that that is too generic for this project, which seems to be Paul's leaning, then make a more "real" CONTRIBUTING.md

Honny1:
@mheon will create a guide for rust projects. Then I'll link it to aardvark-dns.

PR #1202, opened 2025-03-14
Fix detection of Firewalld's StrictForwardPorts property

mheon:
The code was written to work with a boolean property. Unfortunately, it's not a boolean, it's a DBus Variant holding a String which can be either "yes" or "no". This, unfortunately, complicates our decoding code somewhat considerably.

I don't know how this actually worked when I tested it in the first place, unless the API type changed in firewalld before final release, which I don't *think* happened? Fortunately, the code here is written to be very forgiving of errors, but I added some warning logs to give us a trail if stuff starts blowing up in the future (because Variant means the type can, theoretically, change in the future, which would be unpleasant).

Fixes https://issues.redhat.com/browse/RHEL-80266

### DIFF
--- a/src/firewall/firewalld.rs
+++ b/src/firewall/firewalld.rs
@@ -1062,7 +1062,35 @@ pub fn is_firewalld_strict_forward_enabled(dbus_con: &Option<Connection>) -> boo
         "Get",
         &("org.fedoraproject.FirewallD1.config", "StrictForwardPorts"),
     ) {
-        Ok(b) => b.body().deserialize().unwrap_or(false),
+        Ok(b) => {
+            let variant_str: String = match b.body().deserialize::<Value>() {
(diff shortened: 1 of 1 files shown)


Luap99 (approved):
What firewalld version do we need? Do we need to wait to f42 to test or is this already on f41?
It would be great to write some tests for that, of course if our fedora base does not support it there is not much to do.

LGTM

mheon:
We need 2.3.0 or higher, so it has to be F42

Luap99:
Ok I created https://github.com/containers/netavark/issues/1203 so we don't forget it
Do we need to tag a new release with it?

/lgtm

mheon:
I have a card for RHEL 9.6/10.0, backport and release sounds appropriate. Can go out to Fedora as well, since 42 with the new firewalld will be available soon-ish

Issue #1133, opened 2024-11-27
when container is run with a network hosts search domains are ignored


mattp-:
on host
```
mphillips81@mmp-pw-dev ~/repos/ose (main)✗ % cat /etc/resolv.conf
options attempts:1 timeout:2

search foo.com bar.com 
```
on podman:
```
mphillips81@mmp-pw-dev ~/repos/ose (main)✗ % docker run -it  --network=ose_default obfuscated.com/rhel7-dpkg:latest /bin/sh
sh-4.2# cat /etc/resolv.conf
search dns.podman
nameserver 10.89.4.1
options attempts:1 timeout:2
sh-4.2# exit
mphillips81@mmp-pw-dev ~/repos/ose (main)✗ % docker run -it  obfuscated.com/rhel7-dpkg:latest /bin/sh
sh-4.2# cat /etc/resolv.conf
search foo.com  bar.com 
nameserver 10.0.2.3
nameserver 10.10.10.10
nameserver 10.10.10.11
options attempts:1 timeout:2
sh-4.2# exit
```
sorry for the obfuscated bits, it should be reproducible with any image.
when run via docker the search domains are maintained as expected, both with and without network. I'm not sure if this is the right repo to file an issue, please redirect me if not. thank you!

mattp-:
forgot to mention this was tested/verified with 
```
mphillips81-dev-mse.pw1.bcc.bloomberg.com /root # podman version
Client:       Podman Engine
Version:      4.9.4-rhel
API Version:  4.9.4-rhel
Go Version:   go1.21.11 (Red Hat 1.21.11-1.el9_4)
Built:        Thu Aug  8 08:51:41 2024
OS/Arch:      linux/amd64
```
this may have been fixed in a newer version, i dont have a means to easily verify that at the moment

Luap99:
Its not fixed and it is a podman issue. Netavark has no knowledge of the current search domains. We just return dns.podman as we use that for aardvark-dns but it is up to podman to append the other entries if this what the correct behavior should be.

PR #1214, opened 2025-03-27
remove search domain from response

Luap99:
Podman will use this as main search domain and ignore everything else then. We don't need to set the search domain at all, aardvark-dns resolves names without it fine.

Fixes: #1133
Fixes: https://issues.redhat.com/browse/RHEL-83787


mheon:
LGTM

PR #1215, opened 2025-03-27
GHA: Automate release

ashley-cui:
Automatically create release from tag, and build artifacts. Release notes are pulled from RELEASE_NOTES.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,143 @@
+name: Release
+on:
+  push:
+    tags:
+      - '*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release version to build and upload (e.g. "v9.8.7")'
+        required: true
+      buildonly:
+        description: 'Build only: Do not create release'
+        default: "true"  # 'choice' type requires string value
+        type: choice
+        options:
+          - "true"  # Must be quoted string, boolean value not supported.
+          - "false"
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    steps:
+    - name: Determine Version
+      id: getversion
+      run: |
+        if [[ -z "${{ inputs.version }}" ]]
+        then
+              VERSION=${{ github.ref_name }}
+        else
+              VERSION=${{ inputs.version }}
+        fi
+        if ! grep -Eq 'v[0-9]+(\.[0-9]+(\.[0-9]+(-.+)?)?)?$' <<<"$VERSION"
+          then
+            echo "Unable to parse release version '$VERSION' from github event JSON, or workflow 'version' input."
+            exit 1
+          fi
+
+        echo "version=$VERSION" >> $GITHUB_OUTPUT
+        echo "::notice::Building $VERSION"
+    - name: Determine release
+      id: buildonly
+      run: |
+        # The 'tag' trigger will not have a 'buildonly' input set. Handle
+        # this case in a readable/maintainable way.
+        if [[ -z "${{ inputs.buildonly }}" ]]
+        then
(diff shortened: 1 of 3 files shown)


ashley-cui:
Here's the test run: 
[Run](https://github.com/ashley-cui/netavark-actions-test/actions/runs/14115153961),[Release](https://github.com/ashley-cui/netavark-actions-test/releases/tag/v9.9.9)

.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+    name: Build Artifacts
+    runs-on: ubuntu-latest
+    needs: check
+    steps:
+    - name: Checkout Version
+      uses: actions/checkout@v4
+      with:
+        ref: ${{needs.check.outputs.version}}
+    - name: Install Protoc
+      uses: arduino/setup-protoc@v3
```
Luap99, line 68: I would prefer a normal apt-get install here, from contrib/container_images/Dockerfile.Rust
`apt-get update && apt-get -y install protobuf-compiler libprotobuf-dev`

Also do we know which rust version is used, if it is the normal ubuntu one that might be very out of date and fail to compile. Or do the github actions images always include newer versions?

(@baude @mheon Why do we even build a binary that we upload? It is not a static one so it is unlikely to run everywhere, IMO just the vendor tarball is all we need, but anyway that is besides the release action here so I am fine to defer that discussion to another day)

.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+    - name: Build Artifacts
+      run: |
+          make vendor-tarball
+          mkdir release-artifacts
+          mv netavark.gz release-artifacts
+          mv netavark-${{needs.check.outputs.version}}-vendor.tar.gz release-artifacts
+          pushd release-artifacts
+          sha256sum * > sha256sum
```
Luap99, line 76: the make target also generates the SHASUM file so I don't think it would need to be creates again.

Also if it helps we could just change the make target to put the files in a directory, I would prefer that because if we modify the target it is easy to forget that the paths are used in the github action because it will not run on PRs.

.github/workflows/release.yml
```diff
@@ -0,0 +1,142 @@
...
+      uses: actions/checkout@v4
+      with:
+        ref: ${{needs.check.outputs.version}}
+    - name: Get release notes
+      run: |
+        ver="${VERSION%-rc*}"
+        releasenotes="$VERSION-release-notes.md"
+        awk -v ver=$ver '/^## / { if (p) { exit }; if ($2 == ver) { p=1; next } } p' RELEASE_NOTES.md > $releasenotes
+        if [[ -z $(grep '[^[:space:]]' $releasenotes) ]]; then
+          if [[ $VERSION != *-rc* ]]; then
```
Luap99, line 104: FYI we generally don't do rc's here but it certainly doesn't hurt to handle the case anyway in case we start doing some.

.github/workflows/release.yml
ashley-cui: Rust is by default installed in the GHA, I'm not actually sure of the version, I can check it. I'm not too familiar with rust, but I can just run `rustup update` in the action to make sure we're always on latest, if that sounds like a good idea?

.github/workflows/release.yml
ashley-cui: Yep, only kept this bit in since netavark 1.0.0 had rc's so it's not out of the question for the future.

Luap99 (approved):
LGTM

.github/workflows/release.yml
ashley-cui: updated!

PR #1221, opened 2025-04-04
Release GHA: Automatically publish crate

ashley-cui:
Unfortunately this can't be tested until we release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -141,3 +141,28 @@ jobs:
             $RC \
             $LATEST \
             release-artifacts/*
+
+  publish-crate:
+    name: Publish Crate
+    if: needs.check.outputs.buildonly == 'false'
+    runs-on: ubuntu-latest
+    needs: check
+    steps:
+    - name: Update Rust
+      run: |
+        rustup update stable
+        rustc --version
+        cargo --version
+    - name: Install Protoc
+      run: |
+        sudo apt-get update
+        sudo apt-get -y install protobuf-compiler libprotobuf-dev
+    - name: Checkout Version
+      uses: actions/checkout@v4
+      with:
+        ref: ${{needs.check.outputs.version}}
+    - name: Publish crate
+      run: make crate-publish
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+


.github/workflows/release.yml
```diff
@@ -141,3 +141,28 @@ jobs:
             $RC \
             $LATEST \
             release-artifacts/*
+
+  publish-crate:
```
Luap99, line 145: Does this need to depend on release?

I guess technically there are separate and we pushed the tag at this point already so I guess having a crate release without a github release is fine.

.github/workflows/release.yml
```diff
@@ -141,3 +141,28 @@ jobs:
...
+        sudo apt-get update
+        sudo apt-get -y install protobuf-compiler libprotobuf-dev
+    - name: Checkout Version
+      uses: actions/checkout@v4
+      with:
+        ref: ${{needs.check.outputs.version}}
+    - name: Publish crate
+      run: make crate-publish
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
```
Luap99, line 167: is that secret created already? I don't see it in the github settings?

ashley-cui: Added now! Had to obtain write access first.

Luap99: Did you add your personal token? I don't see an option to use a bot account or something and in this case I prefer to use my account as the user who pushed shows up on https://crates.io/crates/netavark/versions.
Not against you but because you are not a main maintainer here it would look a bit odd if you would be listed as publisher without doing commits.

ashley-cui: Yeah absolutely, could you update the token in the settings?

Luap99: yes I will do that later, I like to look into the crates.io settings if there are some best practises regarding setting up this token for CI use.

.github/workflows/release.yml
Luap99: It is fine like that.

.github/workflows/release.yml
Luap99: ok for future reference, I set the crate scope to netavark only and only allow the token to publish new updates which seems to be as restrictive as one can be when generating a token.

Luap99 (approved):
LGTM

Issue #1226, opened 2025-04-10
fix CI problem with incomplete Cargo.lock chnages


Luap99:
Somehow renovate created a PR without updating Cargo.lock https://github.com/containers/netavark/pull/1220

In theory any user could commit such thing as well so we need CI to enforce that the Cargo.toml/lock are always in sync.
I think we might be a missing a tree is clean check like we do in podman if there are incomplete vendor changes. A normal build should not cause any file changes in the repo.

PR #1233, opened 2025-04-23
ci: ensure build does not modify comitted files

Luap99:
ci: ensure tree is clean after build

To ensure Cargo.lock or other files that are committed are not modified
by the build in CI.

Fixes: https://github.com/containers/netavark/issues/1226

---

Makefile: use cargo --locked in CI

CI should never modify the lockfile so add --locked when running under
CI, the CI env is always defined by cirrus.

https://doc.rust-lang.org/cargo/commands/cargo.html#manifest-options


mheon:
LGTM

Luap99:
@baude PTAL and merge, renovate seem to keep filing broken PRs so we need this to protect against further broken merges.

PR #1235, opened 2025-04-25
Makefile: correctly set SOURCES

Luap99:
Use find to list all source files recursively, the ** doesn't actually seem to work correctly in all cases. It requires the use the the globstar shell option which may or may not be set.

While at it also add Makefile so we rebuild if the Makefile is changed, e.g. when adding new flags to cargo.

Fixes: a08bb557fd ("Makefile: do not rebuild if nothing changed")


baude:
LGTM

Issue #1019, opened 2024-07-04
Container with network alias containing a space causes aardvark-dns crash


l-austenfeld:
When you create a container that has a network alias containing a space, a malformed aardvark-dns config will be created that will cause it to crash.

```
❯ podman --version
podman version 5.1.1
❯ /usr/lib/podman/netavark --version
netavark 1.11.0
❯ /usr/lib/podman/aardvark-dns --version
aardvark-dns 1.11.0
```

Steps to reproduce:  
- `podman network create test` to create a network with dns enabled
- `podman run --rm -it --network test debian` to create a working container. You can see a `aardvark-dns` process running in `htop` for example and dns requests in the container are working
- `podman run --rm -it --network test --network-alias "test alias" debian` in a second terminal. If you observe the `aardvark-dns` process using `htop` it will disappear when the container is started and dns in both containers is now broken

Contents of `/run/user/1000/containers/networks/aardvark-dns/test`:
```
10.89.1.1
8965e883bfc09b4ce60728c66bc85ed62fa589218174dbbecca1a00dcf0eafc7 10.89.1.4  eloquent_keller,8965e883bfc0
f8bd2db9d504b780bd9b5567e2823ce99db12d207cf468d8614d36a9ba242231 10.89.1.5  goofy_wilbur,test alias,f8bd2db9d504
```

Luap99:
Such a name is not a valid dns or hostname, thus it would make the most sense for podman to reject it right away.

Of course we can improve error handling for aardvark-dns too but this is not a name that is going to work regardless.

l-austenfeld:
I agree that such a name does not make sense, in my case it is the result of a forgejo runner generating a ci job name like `build-image (amd64)` as part of a job matrix (job names themselves are not allowed to have spaces) and then setting that as a container name alias.  
Both podman and docker accept aliases like this (I assume for uses other than dns).

I don't know the best way to handle this, but rejecting invalid names before generating the config file sounds reasonable.

l-austenfeld:
I temporarily fixed the issue on my system using the following patch:
```
diff --git a/src/dns/aardvark.rs b/src/dns/aardvark.rs
index 6c1314d..03906b9 100644
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -282,7 +282,12 @@ impl Aardvark {
     }
 
     fn commit_entry(entry: &AardvarkEntry, mut file: File) -> Result<()> {
[...]

PR #1236, opened 2025-04-25
aardvark: handle names more safely

Luap99:
Since we write space, command and newline separated file these
characters have a special meaning when parsing the file. Any such
character inside the strings can break the parser. We need checks to
disallow these characters.

While it would have been easy to add a check at the time we write the
file there we could not really make a good decision on what to do with
the value. As such I create a newtype which wraps the string but it can
only be constructed via the TryFrom trait. That is enough to gurantee
that any caller must have called the validation code.

Note that we only print a warning and ignore such invalid value, a hard
error might cause compatibility issues as per the issue some automated
tooling generates such names.

Fixes: https://github.com/containers/netavark/issues/1019
Fixes: https://github.com/containers/podman/issues/25927

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -20,15 +20,38 @@ const SYSTEMD_CHECK_PATH: &str = "/run/systemd/system";
 const SYSTEMD_RUN: &str = "systemd-run";
 const AARDVARK_COMMIT_LOCK: &str = "aardvark.lock";
 
-#[derive(Clone, Debug)]
+/// For better safety we wrap &str in our own custom type here
+/// where we can enforce that the caller can only construct it
+/// via the TryFrom trait. With that we can guarantee via the
+/// type system that we never get invalid chars here when we
+/// write the config file.
+#[derive(Debug)]
+pub struct SafeString<'a>(&'a str);
+
+impl<'a> TryFrom<&'a str> for SafeString<'a> {
+    type Error = &'static str;
+
+    fn try_from(value: &'a str) -> std::result::Result<Self, Self::Error> {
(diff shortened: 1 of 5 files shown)


Luap99:
Moving to draft as I still need to add a test, but code wise this should be ready. I really like this kind of code in rust to use the type system to create safe types that are impossible to misuse.

baude:
LGTM

Luap99:
@mheon PTAL

src/dns/aardvark.rs
```diff
@@ -19,15 +20,38 @@ const SYSTEMD_CHECK_PATH: &str = "/run/systemd/system";
...
+pub struct SafeString<'a>(&'a str);
+
+impl<'a> TryFrom<&'a str> for SafeString<'a> {
+    type Error = &'static str;
+
+    fn try_from(value: &'a str) -> std::result::Result<Self, Self::Error> {
+        if value.is_empty() {
+            return Err("name is empty");
+        }
+        if value.contains([',', '\n', ' ']) {
```
mheon, line 38: Feels like this is nowhere near aggressive enough. A valid DNS hostname contains letters, numbers, and hyphens only. An alias containing, for example, `#` should still break us. I know we dislike regexes but with something as restrictive as this I think it makes sense to use one.

Luap99: Well yes, but so far my only focus was to ensure we do not generate an invalid file. Everything else should get handled by aardvark-dns IMO. I didn't find a clear way where exactly is described  what works and what not. I know the actual spec is rather strict but I worry that docker might allow more?

PR #1247, opened 2025-05-12
run cargo update and update nftables to 0.6.2

Luap99:
This is a rather large change as it switched to using Cow<> which
requires a fair amount of changes. But in the end it should help to
avoid some copies for at least for the 'static str chain names we have.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,9 +93,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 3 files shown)


mheon:
Little gross but LGTM on the whole

Luap99:
@baude PTAL

flouthoc (approved):
LGTM
/lgtm
/approve

PR #1250, opened 2025-05-13
rpm: update description

Luap99:
We do support nftables for a while now. Just noticed while checking the fedora builds.


Luap99:
cc @lsm5

flouthoc (approved):
LGTM

Issue #1262, opened 2025-05-28
dhcp-proxy error message in logs


Luap99:
When a lease is cleared up I consistently see this error in the log:
```
[ERROR netavark::commands::dhcp_proxy] no available capacity
```

We should figure out where it is coming from, I don't see this text in our codebase directly

Issue #1024, opened 2024-07-12
DHCP proxy - no available capacity / crash when using external DHCP service


jjzazuet:
Hi, I have a few standalone containers running under podman, using a `macvlan` network to make them available to an internal LAN network. I'm observing the following:

1: Every time the external router assigns a DHCP host configuration to a container, `netavark` logs this message:
```
dhcp-proxy: [ERROR netavark::commands::dhcp_proxy] no available capacity
```

2: Every week or so I will get a hard crash on `netavark` with the following logs, and the container will no longer be reachable at the static IP lease addresses:
```
Jul 11 21:29:36 build-00 user.notice dhcp-proxy: thread '<unnamed>' panicked at library/std/src/sys/pal/unix/stack_overflow.rs:158:13:
Jul 11 21:29:36 build-00 user.notice dhcp-proxy: failed to set up alternative stack guard page: Out of memory (os error 12)
Jul 11 21:29:36 build-00 user.notice dhcp-proxy: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Jul 11 21:29:36 build-00 user.notice dhcp-proxy: thread '<unnamed>' panicked at library/std/src/sys/pal/unix/stack_overflow.rs:154:13:
Jul 11 21:29:36 build-00 user.notice dhcp-proxy: failed to allocate an alternative stack: Out of memory (os error 12)
Jul 11 21:29:36 build-00 daemon.err /etc/init.d/netavark-dhcp-proxy[38874]: start-stop-daemon: failed to start `/usr/libexec/podman/netavark'
Jul 11 21:29:36 build-00 user.notice dhcp-proxy:  * start-stop-daemon: failed to start `/usr/libexec/podman/netavark'
```

I just tried building and upgrading to the latest upstream version of `netavark`, but I'm seeing the same log messages. I'll wait to see if I get a crash with this version.

```
# /opt/netavark version
{
  "version": "1.12.0-dev",
  "commit": "e182147b6aea964f572a4ca981bc000698d59539",
  "build_time": "2024-07-12T03:39:23.741229481+00:00",
  "target": "x86_64-alpine-linux-musl",
  "default_fw_driver": "iptables"
}
```

```
# podman network inspect podman30
[
     {
          "name": "podman30",
          "id": "71d03f55fc0de8a074d5e5de88759269e32da004c568f99bb94a420f2e7f31a2",
          "driver": "macvlan",
          "network_interface": "br30",
[...]

PR #1263, opened 2025-05-28
dhcp_proxy,teardown: don't attempt to reset timer which is not active

flouthoc:
When `--activity-timeout 0` there should be no reason to push messages on `activity_timeout_tx` as `activity_timeout_rx` is not receving any messages as a result all the messages pushed are just stuck in queue, resulting in `no available capacity` when queue becomes full.

Closes: https://github.com/containers/netavark/issues/1262


src/commands/dhcp_proxy.rs
```diff
@@ -55,6 +55,8 @@ struct NetavarkProxyService<W: Write + Clear> {
     cache: Arc<Mutex<LeaseCache<W>>>,
     // the timeout for the dora operation
     dora_timeout: u32,
+    // inactivity_timeout duration in seconds
+    inactivity_timeout: Duration,
```
Luap99, line 59: Would it be cleaner to wrap timeout_sender in an Option<> and then the option is set to None when the timeout is zero?
Looking at the code there should not be a reason to create the channel at all when the timeout is zero I think.

flouthoc:
This should also close https://github.com/containers/netavark/issues/1024

PR #1268, opened 2025-06-03
fix new clippy lint warnings

Luap99:
drop DhcpProxy error type from NetavarkError

The tonic status is quite large and thus a new clippy lint now seem to
fail[1].

Only the test binary used the variant anyway so just drop it and also
print only the error text as the metadata is not useful in the error
message. Then update the tests to check the message instead of exit
code.

[1] https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err

make more use of NetavarkError

A clippy lint[1] wants us to use the other() function on the io error.
However we already have our own error type for custom messages so just
use that and no co-opt the std::io::Error type.

[1] https://rust-lang.github.io/rust-clippy/master/index.html#io_other_error

## Summary by Sourcery

Standardize error handling across the project by using NetavarkError for error construction, remove unused DHCPProxy status variant, simplify client exit logic, and update tests to validate error messages and a generic failure code

Enhancements:
- Convert internal functions to return NetavarkResult and replace std::io::Error usage with NetavarkError::msg
- Remove the DHCPProxy variant and tonic::Status conversion from NetavarkError and simplify client error handling to a single exit code
- Adjust error creation in dns, core_utils, firewall, and DHCP proxy libraries to use custom error messages

Tests:
- Update BATS tests to expect a generic exit code of 1 and assert on specific error message patterns instead of exit codes

### DIFF
--- a/src/dns/aardvark.rs
+++ b/src/dns/aardvark.rs
@@ -108,7 +108,7 @@ impl Aardvark {
         false
     }
 
-    pub fn start_aardvark_server(&self) -> Result<()> {
+    pub fn start_aardvark_server(&self) -> NetavarkResult<()> {
         log::debug!("Spawning aardvark server");
 
         let mut aardvark_args = vec![];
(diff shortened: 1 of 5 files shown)


mheon:
LGTM

PR #1269, opened 2025-06-04
release v1.15.2

Luap99:
So we can get the thread leak fix out to users.

## Summary by Sourcery

Release v1.15.2: fix thread leak in the DHCP proxy, suppress bogus errors when activity timeout is zero, and make inactivity timeout optional.

Bug Fixes:
- Fix a thread leak in the DHCP proxy for each started container
- Suppress bogus errors when the DHCP proxy is run with an activity timeout of zero

Enhancements:
- Make the DHCP proxy inactivity timeout channel optional and disable related tasks when timeout is set to zero

Documentation:
- Add v1.15.2 release notes

Chores:
- Bump package version to v1.15.2
- Update mozim dependency to 0.2.6

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -487,9 +487,9 @@ checksum = "2a2330da5de22e8a3cb63252ce2abb30116bf5265e89c0e01bc17015ce30a476"
 
 [[package]]
 name = "dhcproto"
-version = "0.9.0"
+version = "0.12.0"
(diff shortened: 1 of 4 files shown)


Luap99:
@baude @flouthoc @mheon PTAL

mheon:
Sure, LGTM

PR #1281, opened 2025-07-14
Set bridge MTU to match default route.

jankaluza:
When creating a custom network, netavark previously used the default Linux bridge MTU (1500), which undermined performance optimization strategy of using a large MTU (65520) for improved TCP throughput.

This patch ensures that when the user does not explicitly set an MTU, podman attempts to derive a more suitable MTU from the system's default route interface, preserving intended performance benefits.

Also moved get_default_route_interface into core_utils for reuse across modules.

Fixes: https://github.com/containers/podman/issues/23883
Fixes: https://github.com/containers/podman/issues/20009

## Summary by Sourcery

Set bridge MTU to match the system’s default route interface MTU when the user does not specify one, preserving high-throughput network performance and moving shared utilities into core_utils

Bug Fixes:
- Fixes #23883 by avoiding the 1500-byte default bridge MTU

Enhancements:
- Derive and apply the default route interface’s MTU for bridge and veth setups when no explicit MTU is provided
- Expand validate_bridge_link to return both the interface index and its MTU
- Propagate the computed MTU into veth pair creation

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -8,6 +8,7 @@ use netlink_packet_route::link::{
 };
 
 use crate::dns::aardvark::SafeString;
+use crate::network::core_utils::get_default_route_interface;
 use crate::network::dhcp::{dhcp_teardown, get_dhcp_lease};
 use crate::{
     dns::aardvark::AardvarkEntry,
@@ -557,18 +558,22 @@ fn create_interfaces(
     Option<sysctl::SysctlDWriter<'static, String, String>>,
 )> {
     let mut sysctl_writer = None;
-    let (bridge_index, mac) = match host.get_link(netlink::LinkID::Name(
+    let (bridge_index, mtu, mac) = match host.get_link(netlink::LinkID::Name(
         data.bridge_interface_name.to_string(),
     )) {
-        Ok(bridge) => (
-            validate_bridge_link(
+        Ok(bridge) => {
+            let (bridge_index, mtu) = validate_bridge_link(
                 bridge,
                 data.vlan.is_some(),
                 host,
                 &data.bridge_interface_name,
-            )?,
-            None,
-        ),
+            )?;
+            (
+                bridge_index,
+                if data.mtu == 0 { mtu } else { data.mtu },
+                None,
+            )
+        }
         Err(err) => match err.unwrap() {
             NetavarkError::Netlink(e) => {
                 if -e.raw_code() != libc::ENODEV {
@@ -652,6 +657,21 @@ fn create_interfaces(
                     data.bridge_interface_name.to_string(),
                     InfoKind::Bridge,
                 );
+
+                let mut mtu = data.mtu;
+                if mtu == 0 {
+                    // if we have a default route, use its mtu as default
+                    if let Ok(iface_name) = get_default_route_interface(host) {
+                        match core_utils::get_mtu_from_iface(host, &iface_name) {
+                            Ok(iface_mtu) => {
+                                mtu = iface_mtu;
+                            },
+                            Err(e) => debug!(
+                                "failed to get mtu for default interface {iface_name}: {e}, using kernel default",
+                            ),
+                        }
+                    }
+                }
                 create_link_opts.mtu = data.mtu;
 
                 if data.vlan.is_some() {
@@ -699,7 +719,7 @@ fn create_interfaces(
                 host.set_up(netlink::LinkID::ID(link.header.index))
                     .wrap("set bridge up")?;
 
-                (link.header.index, mac)
+                (link.header.index, mtu, mac)
             }
             _ => return Err(err),
         },
@@ -714,6 +734,7 @@ fn create_interfaces(
         internal,
         hostns_fd,
         netns_fd,
+        mtu,
     )?;
     Ok((mac, sysctl_writer))
 }
@@ -729,11 +750,12 @@ fn create_veth_pair<'fd>(
     internal: bool,
     hostns_fd: BorrowedFd<'fd>,
     netns_fd: BorrowedFd<'fd>,
+    mtu: u32,
 ) -> NetavarkResult<String> {
     let mut peer_opts =
         netlink::CreateLinkOptions::new(data.container_interface_name.to_string(), InfoKind::Veth);
     peer_opts.mac = data.mac_address.clone().unwrap_or_default();
-    peer_opts.mtu = data.mtu;
+    peer_opts.mtu = mtu;
     peer_opts.netns = Some(netns_fd);
 
     let mut peer = LinkMessage::default();
@@ -741,7 +763,7 @@ fn create_veth_pair<'fd>(
 
     let mut host_veth =
         netlink::CreateLinkOptions::new(data.host_interface_name.clone(), InfoKind::Veth);
-    host_veth.mtu = data.mtu;
+    host_veth.mtu = mtu;
     host_veth.primary_index = primary_index;
     host_veth.info_data = Some(InfoData::Veth(InfoVeth::Peer(peer)));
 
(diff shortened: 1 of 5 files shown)


jankaluza:
@Luap99 , I would need some help with tests. What I do not understand in the tests is how the veth interfaces are generated in them. I thought I would find some test code which generates the interface so I can change its MTU to non-default one and later check this change is propagated to bridge, but it seems it only creates "lo" and suddenly the tests contain the veth interfaces. I will keep digging into it tomorrow.

Luap99:
> @Luap99 , I would need some help with tests. What I do not understand in the tests is how the veth interfaces are generated in them. I thought I would find some test code which generates the interface so I can change its MTU to non-default one and later check this change is propagated to bridge, but it seems it only creates "lo" and suddenly the tests contain the veth interfaces. I will keep digging into it tomorrow.

You should not have to interact directly with the veth interfaces (other than checking that the mtu was set on them)

The way the tests work is you have to run `run_in_host_netns COMMAND` to configure the "host" netns. So basically in your test you need to add a interface with a default route so your test can read it from there.
These commands should create a dummy interface with a default route which should be enough:
```
ip link add dummy0 type dummy
ip link set dummy0 mtu 9000
ip addr add 192.168.0.0/24 dev dummy0
ip link set dummy0 up
ip route add default via 192.168.0.0
```

src/network/bridge.rs
```diff
@@ -699,7 +716,7 @@ fn create_interfaces(
                 host.set_up(netlink::LinkID::ID(link.header.index))
                     .wrap("set bridge up")?;
 
-                (link.header.index, mac)
+                ((link.header.index, mtu), mac)
```
Luap99: there is no need to use a tuple inside a tuple, three values in a tuple should be valid

src/network/bridge.rs
```diff
@@ -652,7 +653,23 @@ fn create_interfaces(
...
+                                create_link_opts.mtu = iface_mtu;
+                            },
+                            Err(e) => debug!(
+                                "failed to get mtu for default interface {}: {}, using kernel default",
+                                iface_name, e
+                            ),
+                        }
+                    }
+                }
+                let mtu = create_link_opts.mtu;
```
Luap99: I think it would look a bit cleaner to define `let mut mtu = data.mtu` before the branch and then assign to mtu so we only need on `create_link_opts.mtu = mtu` assignment at the end which I find clearer. 

create_link_opts.mtu = data.mtu;

Luap99 (commented):
I think this looks about right, just some style nits and we will need a test for it.

Luap99:
Also you need to use `cargo fmt` to format and there might be lint issues, you can run `make validate` locally to check

jankaluza:
This should be ready for another review :-).

test/630-bridge-vlan.bats
```diff
@@ -140,3 +140,37 @@ echo "$config"
...
+@test "bridge - mtu from default route" {
+    run_in_host_netns ip link add default_route type dummy
+    run_in_host_netns ip link set default_route mtu 9000
+    run_in_host_netns ip addr add 192.168.0.0/24 dev default_route
+    run_in_host_netns ip link set default_route up
+    run_in_host_netns ip route add default via 192.168.0.0
+
+    local vlan=20
+    local config=$(createVlanConfig managed $vlan)
+
```
Luap99: This has nothing to do with the vlan option, I suggest you move the tests into a new bridge mtu file.
Also looking at our tests I actually cannot see any existing bridge mtu config setting test either (only for macvlan/ipvlan). So we should add one as well to make sure the config file does overwrite the default.

test/630-bridge-vlan.bats
```diff
@@ -140,3 +140,37 @@ echo "$config"
...
+    # check that interface exists
+    run_in_container_netns ip -j --details link show eth0
+    link_info="$output"
+    assert_json "$link_info" ".[].mtu" "=="  "9000" "MTU matches default route MTU"
+}
+
+@test "bridge - mtu from existing bridge" {
+    run_in_host_netns ip link add podman0 type bridge
+    run_in_host_netns ip link set podman0 mtu 9000
```
Luap99: I recommend to use a different mtu value here just to have it different from the other test. Consider the case where the code hard codes to 9000 in which case both would pass but not do what we want.

test/630-bridge-vlan.bats
jankaluza: I've moved the tests to 640-bridge-mtu.bats and added two more tests which adds the "mtu" option. I've found and issue thanks to that which I fixed. The code now does following when existing bridge is found to not override the mtu option with bridge mtu:

`            (bridge_index, if data.mtu == 0 { mtu } else { data.mtu }, None)`

test/640-bridge-mtu.bats
```diff
@@ -0,0 +1,65 @@
...
+    run_in_host_netns ip addr add 192.168.0.0/24 dev default_route
+    run_in_host_netns ip link set default_route up
+    run_in_host_netns ip route add default via 192.168.0.0
+
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-managed.json setup $(get_container_netns_path)
+
+    # check that interface exists
+    run_in_container_netns ip -j --details link show eth0
+    link_info="$output"
+    assert_json "$link_info" ".[].mtu" "=="  "9000" "MTU matches default route MTU"
```
Luap99: This is duplicated several times, please split it into a function that accepts the mtu as first arg to deduplicate.

test/640-bridge-mtu.bats
```diff
@@ -0,0 +1,65 @@
...
+@test "bridge - mtu from default route" {
+    run_in_host_netns ip link add default_route type dummy
+    run_in_host_netns ip link set default_route mtu 9000
+    run_in_host_netns ip addr add 192.168.0.0/24 dev default_route
+    run_in_host_netns ip link set default_route up
+    run_in_host_netns ip route add default via 192.168.0.0
+
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-managed.json setup $(get_container_netns_path)
+
+    # check that interface exists
```
Luap99: we should also check the the host side veth (veth0) and the bridge we created also have the right mtu

test/640-bridge-mtu.bats
```diff
@@ -0,0 +1,65 @@
...
+
+load helpers
+
+
+@test "bridge - mtu from default route" {
+    run_in_host_netns ip link add default_route type dummy
+    run_in_host_netns ip link set default_route mtu 9000
+    run_in_host_netns ip addr add 192.168.0.0/24 dev default_route
+    run_in_host_netns ip link set default_route up
+    run_in_host_netns ip route add default via 192.168.0.0
```
Luap99, line 34: this should also be put in a function to deduplicate with the other test that sets up the default route

test/640-bridge-mtu.bats
```diff
@@ -0,0 +1,65 @@
...
+}
+
+@test "bridge - mtu from config with existing bridge" {
+    run_in_host_netns ip link add podman0 type bridge
+    run_in_host_netns ip link set podman0 mtu 9001
+    run_in_host_netns ip link set up podman0
+
+    export RUST_LOG=debug
+    run_netavark --file ${TESTSDIR}/testfiles/bridge-mtu.json setup $(get_container_netns_path)
+    echo "$output"
```
Luap99: debug leftover?

test/testfiles/bridge-mtu.json
```diff
@@ -0,0 +1,35 @@
...
+      "dns_enabled": false,
+      "ipam_options": {
+        "driver": "host-local"
+      },
+      "options": {
+        "mtu": "9002"
+      }
+    }
+  }
+}
```
Luap99: missing final newline

Luap99:
Oh and btw `#23883` won't work as bug reference sine the bug is in podman. You would have to put in
`containers/podman#23883`

It also fixes https://github.com/containers/podman/issues/20009

test/640-bridge-mtu.bats
```diff
@@ -0,0 +1,68 @@
...
+function check_iface_mtu() {
+    local host_or_container=$1
+    local iface=$2
+    local mtu=$3
+
+    if [ "$host_or_container" = "host" ]; then
+        run_in_host_netns ip -j --details link show "$iface"
+    else
+        run_in_container_netns ip -j --details link show "$iface"
+    fi
```
Luap99, line 17: nit it should be possible to do `run_in_${host_or_container}_netns ip -j --details link show "$iface"`
so no need for the branch, but not blocking this is fine as well

Luap99 (approved):
LGTM
@mheon PTAL

Issue #1076, opened 2024-09-07
flaky test in debci


siretart:
I'm afraid that this is very similar to https://github.com/containers/netavark/issues/433 but I can see it on amd64, more specifically, on Debian's CI system called debci. Here, the tests are running in a qemu VM with 4G of ram and 2 cores.

It seems to be quite reliably to break, but all failures are fairly inconsistent. Example runs I've observed so far:

* https://ci.debian.net/packages/n/netavark/unstable/amd64/51297650/
* https://ci.debian.net/packages/n/netavark/unstable/amd64/51298757/
* https://ci.debian.net/packages/n/netavark/unstable/amd64/51298772/
* https://ci.debian.net/packages/n/netavark/testing/amd64/51297649/
* https://ci.debian.net/packages/n/netavark/testing/amd64/51310986/

I've managed to trigger it once on my laptop, by running the tests on four qemu machines in parallel. I've kept going running 5 machines in parallel, but could not reproduce anymore.

Any ideas/suggestions? Maybe I should increase the `sleep` statements in `helpers.bash`?

siretart:
I've been informed that the debci logs are being recycled.
To preserve the logs, I've capture them here https://gist.github.com/siretart/19cfe578df59aa7a0158d9488ca0ce25

siretart:
So, I've been staring at `helpers.bash`, and would like to run a thought by you:

All tests fail while running `run_nc_test()`. That function, in simple terms, opens two `nc` processes, one to listen on a port and the other to write to another port, potentially in different network namespaces, to test whether  netavark is setting up the connection properly.

In all failing cases we observe that the assertion fails with the random string (genererated by https://github.com/containers/netavark/blob/4d896f6f94b496ec279b68ed3aadd4c9595eddef/test/helpers.bash#L660) not matching the empty string). Here is an example output:

```
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #|     FAIL: ncat received data
# #| expected: 'MtMvuBHY3k'
# #|   actual: ''
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

[...]

PR #1282, opened 2025-07-14
test: Introduce NCATOPTS variable to specify ncat options

ricardobranco777:
test: Introduce NCATOPTS variable to specify ncat options

ncat 7.96 changed default behaviour and we may need to specify some options.

https://nmap.org/changelog.html#7.96

## Summary by Sourcery

Introduce NCATOPTS environment variable in test helpers to override ncat options and apply it to run_nc_test, addressing changed default behavior in ncat 7.96.

New Features:
- Add NCATOPTS environment variable for specifying ncat options in tests

Enhancements:
- Use NCATOPTS in run_nc_test to pass custom options to ncat

### DIFF
--- a/test/helpers.bash
+++ b/test/helpers.bash
@@ -5,6 +5,10 @@ NETAVARK=${NETAVARK:-./bin/netavark}
 
 TESTSDIR=${TESTSDIR:-$(dirname ${BASH_SOURCE})}
 
+# ncat 7.96 changed default behaviour
+# https://nmap.org/changelog.html#7.96
+NCATOPTS="$NCATOPTS"
+
 # export RUST_BACKTRACE so that we get a helpful stack trace
 export RUST_BACKTRACE=full
 
@@ -623,7 +627,7 @@ function run_nc_test() {
     local connect_ip=$4
     local host_port=$5
 
-    local nc_common_args=""
+    local nc_common_args="$NCATOPTS"
     exec {stdin}<>/dev/null
 
     case $proto in


ricardobranco777:
Bear with me while I fix the openSUSE Tumbleweed tests to work with the latest Nmap.

They work fine with ncat 7.95 but fail with 7.97.  These are the changes in ncat 7.96:
- > [Ncat]Ncat in connect mode no longer defaults to half-closed TCP connections. This makes it more compatible with other netcats. The -k option will enable the old behavior. See https://seclists.org/nmap-dev/2013/q1/188 [Daniel Miller]
- > [Nsock][GH#2788]Fix an issue affecting Ncat where unread bytes in the SSL layer's buffer could not be read until more data arrived on the socket, which could lead to deadlock. [Daniel Miller]
- > [Ncat][GH#2422]New Ncat option -q to delay quit after EOF on stdin, the same as traditional netcat's -q option. [Daniel Miller]
- > [Ncat][GH#2843]Ncat in listen mode with -e or -c correctly handles error and EOF conditions that had not been being delivered to the child process.

ricardobranco777:
I tried `-k` & `-q` without success.   Why did you choose ncat over OpenBSD's netcat?

WIP.

Luap99:
>  Why did you choose ncat over OpenBSD's netcat?

I don't remeber the exact details but the openbsd behavior caused hangs because it did not close the connection on EOF so I went with the ncat one.

The more I look at ncat the less we should use it. It breaks us all the time. I have send way to many fix ncat something PRs  across our projects. So maybe the better alternative is to use another tool/write out own (ref https://github.com/containers/netavark/issues/1076)

ricardobranco777:
Closing this as I couldn't find a set of flags to enable the previous behaviour and the long-term solution would be a custom tool.  This also affects aardvark-dns & podman.

For now I suggest a documentation fix stating that only the 7.92 & 7.95 versions work.

Luap99:
@ricardobranco777 Thanks for bringing this change to my attention.

Fedora is kept on ncat 7.92 due licensing issues AFAIK https://bugzilla.redhat.com/show_bug.cgi?id=2296006 so this is not something our CI testing here would notice.

So yeah I think we should prioritize removing ncat altogether since it doesn't seem to fit our needs here well and it breaks distros that ship that update it is certainly not great to require them (you) to work around it.

PR #1283, opened 2025-07-16
Some minor changes, and small follow up to the mtu PR

Luap99:
see the commits, some small minor fixes that don't really change the behavior.
With this we safe one netlink roundtrip when looking up the default route interface and also log a warning when we cannot lookup the mtu, the kernel should always return one to us.


Luap99:
@mheon @jankaluza PTAL

mheon:
LGTM

jankaluza:
Good improvements of my first Rust patch and also the surrounding code.

LGTM

PR #1299, opened 2025-08-06
type safety: make Socket aware of namespace

Luap99:
Using a PhantomData and zero sized struct we can attach an additional generic data for either the host or container namespace to the struct. Because it is zero sized and not used it is optimized away and only the type checker sees it to enforce the right types are used.

With that we basically create two different netlink socket types Socket<HostNS> and Socket<ContainerNS> so they must be used in all type signatures from now on. To keep the changes smaller I have set HostNS as default generic for the struct so we don't need to change most function signatures.

For al call sides where we pass sockets around the compiler now enforces that we use the right ones avoid any possible mix ups.


Luap99:
yeah this is just so functions can clearly define if they use the host or netns socket. That way we cannot even compile code that might accidentally is given the wrong socket. While testing would likely catch it having it enforced at compile time is even better and it also means your IDE, etc... should already tell you this doesn't work...

The most obvious case is add_default_routes() function, we only ever want add routes inside the container. So we prevent passing the host socket.

mheon:
LGTM. Had no idea this was a thing.

TomSweeneyRedHat:
Any plans to move this forward @Luap99 ?

Issue #1258, opened 2025-05-27
Handle nftables reload/restart


glennzilla:
On systems using nftables but no firewalld, network connections to running containers stop working if the nftables service is ever restarted or reloaded.  (The netavark table disappears from nftables and isn't automatically reloaded.)  While issue #230 addressed handling firewalld reload, it doesn't address non-firewalld nftables systems.  There needs to be a similar systemd "oneshot" command to reload the firewall rules in nftables (netavark table) whenever the nftables service is restarted or reloaded.

The existing `/usr/libexec/podman/netavark firewalld-reload` command does reload the netavark table in nftables, though it also leaves that process running indefinitely, listening for firewalld dbus events, which is pointless on non-firewalld systems.  Alternatively, running `podman network reload --all` also works.  Until this is addressed officially, I added the following nftables systemd override (/etc/systemd/system/nftables.service.d/override.conf), as a workaround:

```
[Service]
ExecStart=podman network reload --all
ExecReload=podman network reload --all
```

Rishikpulhani:
Hi @Luap99,

I'm picking up this issue and have a proposed implementation plan I'd like to confirm.

My plan is to create a new "oneshot" command, firewall-reload, using the core logic from the existing firewalld-reload command to restore the firewall rules from the state files in /run/containers/netavark/.

To fully automate this for nftables-only systems, I'm assuming the correct approach is to modify the netavark.spec file to create a systemd override for nftables.service upon installation. This seems better than requiring users to manually apply a workaround. The override would look like this:

```
[Service]
ExecStart=netavark firewall-reload
ExecReload=netavark firewall-reload
```

Luap99:
> To fully automate this for nftables-only systems, I'm assuming the correct approach is to modify the netavark.spec file to create a systemd override for nftables.service upon installation. This seems better than requiring users to manually apply a workaround. The override would look like this:

I don't think we should mess with nftables.service file directly packing wise. Likely best to leave that up to end users for now.
If we want a systemd service it would be likely best as separate service file similar to netavark-firewalld-reload.service that can be enabled/disabled separately

Rishikpulhani:
Hi @Luap99,
I have raised the PR #1301 , and all automated checks are now passing. Could you please review it when you have a moment? Thank you.

PR #1300, opened 2025-08-07
add nftables reload command

aniskhalfallah:
I added a one-shot command that restores the nftables rules and exits. I initially created a separate systemd service that runs when nftables is restarted, which works on restart but not on reload. To make it respond to reloads as well, we would need to listen to netlink events, but that requires extra capabilities and complexity. however this "one shot" command can be manually hooked into `nftables.service` by adding
```
ExecStart=netavark nftables-reload
ExecReload=netavark nftables-reload
```
Fixes [#1258](https://github.com/containers/netavark/issues/1258)

## Summary by Sourcery

New Features:
- Introduce nftables-reload subcommand for one-shot restoration of nftables rules

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -4,6 +4,7 @@ use crate::error::{NetavarkError, NetavarkResult};
 
 pub mod dhcp_proxy;
 pub mod firewalld_reload;
+pub mod nftables_reload;
 pub mod setup;
 pub mod teardown;
 pub mod update;
--- a/src/commands/nftables_reload.rs
+++ b/src/commands/nftables_reload.rs
@@ -1,0 +1,36 @@
+use crate::{
+    error::NetavarkResult,
+    firewall::{get_supported_firewall_driver, state::read_fw_config},
+    network::constants,
+};
+use std::{
+    ffi::{OsStr, OsString},
+    path::Path,
+};
+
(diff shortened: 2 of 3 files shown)


Rishikpulhani:
@aniskhalfallah 
Apologies, but I was also working on this issue and had shared my thoughts in the discussion thread earlier.
Since we’re both contributing to the same project, it would be great if we could coordinate a bit—perhaps by commenting on the issue before picking it up—so we can avoid duplicating efforts.

Appreciate your understanding, and looking forward to collaborating smoothly!

aniskhalfallah:
Hi @Rishikpulhani,
I actually have been looking into this issue for a while and had forked the repo earlier, but didn’t leave a comment at the time. I got a bit stuck figuring out why the command wasn’t working on a service reload, and ended up thinking that letting users override the nftables service might be the most practical approach for now. Happy to sync up going forward so we can align our efforts.

Luap99:
Hey sorry for the delay here, I reviewed https://github.com/containers/netavark/pull/1301 just now. I went with that as I like the generic firewall-reload command name better because there isn't really anything nftables in this feature (besides the nft unit to auto reload which is optional). The reload can pick all our supported driver and any user regardless if they use nftables or not can call the command.

Luap99:
https://github.com/containers/netavark/pull/1301 was merged

Issue #1177, opened 2025-02-13
bridge: unmanaged mode should properly support aardvark-dns


Luap99:
Currently if one does try to create network with a gateway that is not on a ip on the host it aardvark-dns will fail because we always assume the gateway is the correct aardvark-dns bind ip. That however is often not the case in umanaged mode where one might expose the existing bridge on their LAN and use the real gateway which can be on another system.

It should be quite simple for us to check the existing bridge for ip addresses and then use them instead of the gateway from the config. If there is no ip address on the bridge then we just fail with a clear error that the user must disable dns (`--disable-dns`) when creating the network.

shivkr6:
> check the existing bridge for ip addresses

We'll have to do that with netlink, right?
I understand the issue and I'm working on it.

Luap99:
> We'll have to do that with netlink, right?

yeah there is dump_addresses() function here already. It shows all address which we don't want so it likely makes sense to modify that function to allow for it to be filter to a specific interface. 

You can find some more info on how to use unmanaged bridge networks in this post https://blog.podman.io/2025/02/netavark-1-14-new-features/

PR #1304, opened 2025-08-11
bridge: unmanaged aardvark-dns fix

shivkr6:
Bind aardvark-dns to the ip addresses of the bridge instead of the gateway when gateway IP addresses are not on the host.

Fixes: https://github.com/containers/netavark/issues/1177

1) Find the IP addresses of the bridge with the modified dump_addresses() function.
I have written a unit test for this function verifying that the function filters the IP addresses of an interface correctly. This function works for all interfaces, not just bridges.

I guess we don't have to check if the bridge's link is valid with validate_bridge_link() because netavark does that when it creates a veth pair? Please correct me if I am wrong.

2) If the dns is enabled and the bridge mode is unmanaged, then the bind IP of aardvark-dns is changed to the IP addresses of the bridge instead of the gateway. If there are no IP address on the bridge then we just fail with a clear error that the user must disable dns (--disable-dns) when creating the network.

I think we should also change the field name from `network_gateways` to `bind_ips` in the `AardvarkEntry struct`?


shivkr6:
I have tested the modified dump_addresses() function but have not tested the fix in the second commit. Can someone please test if it works?

src/network/netlink.rs
```diff
@@ -409,11 +411,26 @@ impl Socket {
...
-        let msg = AddressMessage::default();
+    // If interface's LinkID is supplied, then only the ip addresses of that specific interface is returned. Otherwise all ip addresses of all interfaces are returned
+    pub fn dump_addresses(
+        &mut self,
+        interface: Option<LinkID>,
+    ) -> NetavarkResult<Vec<AddressMessage>> {
+        let mut msg = AddressMessage::default();
 
-        let results = self
-            .make_netlink_request(RouteNetlinkMessage::GetAddress(msg), NLM_F_DUMP | NLM_F_ACK)?;
```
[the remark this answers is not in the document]
shivkr6: I think RTM_GETADDR is a read only operation and does not require the NLM_F_ACK flag. Please correct me if I am wrong.

shivkr6:
Let me know if I should add more tests or extract the IP address collection logic into a helper function as suggested by sourcery-ai.

src/network/netlink.rs
```diff
@@ -409,11 +411,26 @@ impl Socket {
...
+        match interface {
+            Some(LinkID::ID(id)) => {
+                msg.header.index = id;
+            }
+            Some(LinkID::Name(name)) => {
+                let link_message = self.get_link(LinkID::Name(name))?;
+                msg.header.index = link_message.header.index;
+            }
```
Luap99: While this makes sense for consistency I think it make most sense to just accept the Option<id> without the string name support. In my mind each call here should map to one netlink request.

src/network/bridge.rs
```diff
@@ -252,18 +253,36 @@ impl driver::NetworkDriver for Bridge<'_> {
...
-                .ipam
-                .gateway_addresses
-                .iter()
-                .map(|ipnet| ipnet.addr())
-                .collect();
+            // Fixes #1177: In unmanaged mode, the gateway IP may not be on the host.
+            // We need to find an IP on the bridge itself for aardvark-dns to bind to.
+            let bind_addr: Vec<IpAddr> = if data.mode == BridgeMode::Unmanaged {
+                let addresses = host_sock
+                    .dump_addresses(Some(LinkID::Name(data.bridge_interface_name.clone())))?;
```
Luap99: we already created/got the bridge before. We should not have to lookup by name again. Just store/return the bridge id so it can be used here so we safe another netlink roundtrip.

src/network/bridge.rs
```diff
@@ -252,18 +253,36 @@ impl driver::NetworkDriver for Bridge<'_> {
...
+                let mut bind_addr = Vec::with_capacity(addresses.len());
+                for addr_msg in addresses {
+                    for attr in addr_msg.attributes {
+                        if let netlink_packet_route::address::AddressAttribute::Address(ip) = attr {
+                            bind_addr.push(ip);
+                        }
```
Luap99: There is only one Address per addr_msg I think? So this could break and avoid walking further attributes.

Second, do we need to exclude ipv6, link local addresses here? I am not sure we support binding them in aardvark-dns and I don't think we can configure it either in resolv.conf via podman today as link local with zone id cannot work in v6.

src/network/bridge.rs
```diff
@@ -252,18 +253,36 @@ impl driver::NetworkDriver for Bridge<'_> {
...
+                    }
+                }
+                if bind_addr.is_empty() {
+                    return Err(NetavarkError::msg(format!("bridge '{}' in unmanaged mode has no IP addresses, but aardvark-dns requires at least one address to bind to. Please add an IP address or disable DNS for this network (--disable-dns).", data.bridge_interface_name)));
+                }
```
Luap99, line 297: I think we must set response.dns_server_ips as well because that is what is being send to podman for it to put that into resolv.conf of the container.

src/network/bridge.rs
shivkr6: > There is only one Address per addr_msg I think? So this could break and avoid walking further attributes.

I don't think it matters because we are only walking ```netlink_packet_route::address::AddressAttribute::Address(ip)```?
image

> Second, do we need to exclude ipv6, link local addresses here? I am not sure we support binding them in aardvark-dns and I don't think we can configure it either in resolv.conf via podman today as link local with zone id cannot work in v6.

Does aardvark-dns only support binding to ```AddressScope::Universe``` addresses?

src/network/bridge.rs
```diff
@@ -556,6 +579,7 @@ fn create_interfaces(
 ) -> NetavarkResult<(
     String,
     Option<sysctl::SysctlDWriter<'static, String, String>>,
+    u32,
```
shivkr6: I'm getting this warning:
```very complex type used. Consider factoring parts into `type` definitions
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity
`#[warn(clippy::type_complexity)]` on by default```
Should I make it a struct or just making it a type definition is good enough?

File: src/network/bridge.rs (hunk 4 above)
Comment (Luap99) [reply src/network/bridge.rs]: > I don't think it matters because we are only walking netlink_packet_route::address::AddressAttribute::Address(ip)?

What I mean is you can add `break;` below the push to the vector to avoid walking the rest of the netlink attributes as it can only be once there.




> > Second, do we need to exclude ipv6, link local addresses here? I am not sure we support binding them in aardvark-dns and I don't think we can configure it either in resolv.conf via podman today as link local with zone id cannot work in v6.
> 
> Does aardvark-dns only support binding to `AddressScope::Universe` addresses?

I would think so, that should be the only addresses we have to deal with here.

File: src/network/bridge.rs (hunk 6 above)
Comment (Luap99) [reply src/network/bridge.rs]: I think you can define a local struct for that, makes documenting the field values easier so it is easier to read

**Comment by shivkr6** [comment]
I had to write 2 separate tests in test/620-bridge-mode.bats because after the following lines, 
```
expected_rc=1 run_netavark --file ${TESTSDIR}/testfiles/bridge-unmanaged-dns.json setup $(get_container_netns_path)
    assert_json ".error" "bridge 'brtest0' in unmanaged mode has no universe scope IP addresses, but aardvark-dns requires at least one universe scope address to bind to. Please add an universe scope IP address or disable DNS for this network (--disable-dns)."
```
I have to perform a teardown because of the following error:
```
    nsenter -n -m -w -t 87100 ./bin/netavark --rootless false --config /tmp/netavark_bats.uixcfk/config --file /home/Shivang/Desktop/Projects/fun_personal/netavark/test/testfiles/bridge-unmanaged-dns.json setup /proc/87104/ns/net
   {"error":"create veth pair: interface eth0 already exists on container namespace: Netlink error: File exists (os error 17)"}
   [ rc=1 (** EXPECTED 0 **) ]
   #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
   #| FAIL: exit code is 1; expected 0
   #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
and if I perform a teardown with the configuration file (test/testfiles/bridge-unmanaged-dns.json) I get the following error.
```
    nsenter -n -m -w -t 86298 ./bin/netavark --rootless false --config /tmp/netavark_bats.Ro80GM/config --file /home/Shivang/Desktop/Projects/fun_personal/netavark/test/testfiles/bridge-unmanaged-dns.json teardown /proc/86302/ns/net
   {"error":"remove aardvark entries: IO error: No such file or directory (os error 2)"}
   [ rc=1 (** EXPECTED 0 **) ]
   #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
   #| FAIL: exit code is 1; expected 0
   #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
This error does not happen if I specify the config file with dns disabled. I guess this is because when we ran netavark before, it already created the veth pair and the teardown doesn't work with dns enabled because we got an error in netavark before it could create entries for aardvark-dns.

File: src/network/bridge.rs [hunk 7]
```diff
@@ -252,18 +267,41 @@ impl driver::NetworkDriver for Bridge<'_> {
...
-                .gateway_addresses
-                .iter()
-                .map(|ipnet| ipnet.addr())
-                .collect();
+            // Fixes #1177: In unmanaged mode, the gateway IP may not be on the host.
+            // We need to find an IP on the bridge itself for aardvark-dns to bind to.
+            let bind_addr: Vec<IpAddr> = if data.mode == BridgeMode::Unmanaged {
+                let addresses = host_sock.dump_addresses(
+                    Some(bridge_index),
+                    Some(AddressFamily::Inet),
```
Luap99: I don't think that is right as we still like v6 addresses which are AddressFamily::Inet6. IF you add a test case with a v6 address it will not work.

I think we can drop the AddressFamily filter from the function and the just filter in the loop to match
if AddressFamily::Inet || AddressFamily::Inet6

test/620-bridge-mode.bats
```diff
@@ -44,3 +44,28 @@ load helpers
...
+}
+
+@test bridge - unmanaged mode with aardvark-dns bridge ip {
+    run_in_host_netns ip link add brtest0 type bridge
+    run_in_host_netns ip link set brtest0 up
+    run_in_host_netns ip -j --details link show brtest0
+    link_info="$output"
+    assert_json "$link_info" '.[].flags[] | select(.=="UP")' == "UP" "Host bridge interface is up"
+
+    run_in_host_netns ip addr add 10.88.0.1/16 dev brtest0
```
Luap99: Let's do not add .1 address here since that is commonly used as gateway address.

Second, also add v6 address.

test/testfiles/bridge-unmanaged-dns.json
```diff
@@ -0,0 +1,24 @@
...
+			"dns_enabled": true,
+			"driver": "bridge",
+			"id": "53ce4390f2adb1681eb1a90ec8b48c49c015e0a8d336c197637e7f65e365fa9e",
+			"internal": false,
+			"ipv6_enabled": false,
+			"name": "podman",
+			"network_interface": "brtest0",
+			"options": {
+				"mode": "unmanaged"
+			}
```
Luap99, line 29: I am surprised this is even valid like this, I would expect we need to define a subnet here, otherwise we cannot assign a ip address to the container.

src/network/bridge.rs
shivkr6: > I don't think that is right as we still like v6 addresses which are AddressFamily::Inet6. IF you add a test case with a v6 address it will not work.

I thought that you said aardvark-dns does not support binding to IPv6 addresses?
> Second, do we need to exclude ipv6, link local addresses here? I am not sure we support binding them in aardvark-dns and I don't think we can configure it either in resolv.conf via podman today as link local with zone id cannot work in v6.
https://github.com/containers/netavark/pull/1304#discussion_r2267621097

> I think we can drop the AddressFamily filter from the function and the just filter in the loop to match
if AddressFamily::Inet || AddressFamily::Inet6

Means filtering in the userspace (our rust program) right?

shivkr6: @Luap99 I don't think aardvark-dns supports binding to IPv6 addresses because when I tried to bind it to an IPv6 address, I got the following error:
```
    nsenter -n -m -w -t 69564 ip -6 addr add fd00:88:0::100/64 dev brtest0
    nsenter -n -m -w -t 69564 ./bin/netavark --rootless false --config /tmp/netavark_bats.EALFNK/config --file /home/Shivang/Desktop/Projects/fun_personal/temp_netavark/netavark/test/testfiles/bridge-unmanaged-dns.json setup /proc/69568/ns/net
   {"error":"error while applying dns entries: aardvark-dns failed to start: Error from child process\nError starting server failed to bind udp listener on [fd00:88::100]:53: IO error: Cannot assign requested address (os error 99)\nfailed to bind udp listener on [fe80::50a6:6bff:fe84:255a]:53: IO error: Invalid argument (os error 22)"}
   [ rc=1 (** EXPECTED 0 **) ]
```

I've defined the subnet as well in test/testfiles/bridge-unmanaged-dns.json
```
"subnet": "fd00:88:0::/64"

```

Luap99: > Means filtering in the userspace (our rust program) right?


> fe80::50a6:6bff:fe84:255a

That is a ipv6 link local address which must be skipped as it cannot work without zone.

The other ipv6 address should work though. If you add an `ip addr` before you call netavark you most likely see the the addresses is not ready yet as the kernel performs DAD, i.e. it should how TENTATIVE.
If that is the problem I suggest you just disable DAD via sysctl for the bridge like netavark does intenrally for the managed mode. (Note this must happen before you set the bridge link up)

shivkr6: >  If that is the problem I suggest you just disable DAD via sysctl for the bridge like netavark does intenrally for the managed mode. (Note this must happen before you set the bridge link up)

Thanks! I didn't know this was a thing.

I analyzed the output of ```strace ip addr show scope global``` and found out that the kernel ignores the AddressScope::Universe filter and instead returns all scope IP addresses.
I think this is because AddressScope::Universe (RT_SCOPE_UNIVERSE) has the value of 0 (https://docs.rs/netlink-packet-route/latest/src/netlink_packet_route/address/addr_scope.rs.html#3) which is the default value for this field, due to this the kernel treats it as a wildcard I think.
We'll have to perform the filtering in the userspace I guess. This was the reason why aardvark-dns tried to bind to a link local IPv6 address in my previous error.

Sorry, I also want to clarify something.
Which of these points are correct?
1) Aardvark-dns can bind to *all IPv4 addresses* and *all IPv6 addresses except link local IPv6 addresses*.
2) Aardvark-dns can only bind to *global scope IPv4* and *global scope IPv6 addresses*.

shivkr6: I assumed that it's point 1 and pushed all the changes.

Luap99: > We'll have to perform the filtering in the userspace I guess. This was the reason why aardvark-dns tried to bind to a link local IPv6 address in my previous error.

Yes that makes sense and number 1 is correct

src/network/netlink.rs
```diff
@@ -409,11 +411,28 @@ impl Socket {
...
-            .make_netlink_request(RouteNetlinkMessage::GetAddress(msg), NLM_F_DUMP | NLM_F_ACK)?;
+        if let Some(id) = interface_id_filter {
+            msg.header.index = id;
+        }
+        if let Some(addr_family) = address_filter {
+            msg.header.family = addr_family;
+        }
+        if let Some(scope) = scope_filter {
+            msg.header.scope = scope;
+        }
```
Luap99: If we don't actually can use these I would just drop them and only keep the interface filter, these functions only exists for netavark so no need to complicate the function signature when we cannot make use of the extra filters.

src/network/bridge.rs
```diff
@@ -252,18 +267,53 @@ impl driver::NetworkDriver for Bridge<'_> {
...
+                    None,
+                    Some(AddressScope::Universe),
+                )?;
+
+                let addresses: Vec<IpAddr> = addr_msgs
+                    .into_iter()
+                    .filter_map(|addr_msg| {
+                        // address is either a IPv4 address, or it's an IPv6 address that is not a link-local address.
+                        if (addr_msg.header.family == AddressFamily::Inet6
+                            && !(addr_msg.header.scope == AddressScope::Link))
```
Luap99: I think the cleaner way to write this is
```
(addr_msg.header.family == AddressFamily::Inet6
&& addr_msg.header.scope != AddressScope::Link)
|| addr_msg.header.family == AddressFamily::Inet
```
to reduce the brackets which makes it harder to read

shivkr6:
PTAL @Luap99

Luap99:
@mheon PTAL

shivkr6:
Thanks! :)

PR #1305, opened 2025-08-11
Feat(dhcp): Send DHCPRELEASE on container teardown

### LINKED ISSUE #1271: DHCP proxy service does not relinquish lease on container teardown
Hi.

Sending a DHCP RELEASE message on container teardown could help with DDNS setups where containers use fixed hostnames or in cases where the pool of available addresses is small and lease time is long. Most dhcp clients I know of relinquish leases when stopped or can be configured to do so.

But sending DHCP RELEASE is not mandated per dhcp spec so this is not a bug report. More like a feature request.

Rishikpulhani:
When a container using DHCP is torn down, its lease is left active on the DHCP server until it expires. This can be problematic in environments with small IP pools or long lease times. In setups using Dynamic DNS (DDNS), it can also lead to stale DNS records.

This change introduces the capability for netavark to send a DHCPRELEASE message to the DHCP server when a container's network is torn down.

### Implementation 
- Adding a `release_lease` method to the `DhcpV4Service` in `dhcp_service.rs`, which wraps the `release` function from the underlying mozim client.
- Modifying the `teardown` gRPC handler in `dhcp_proxy.rs` to create a temporary `DhcpV4Service` instance, retrieve the lease from the cache, and call the new `release_lease` method in a "best-effort" manner.

### Testing
This feature was verified using the BATS suite in the test-dhcp/ directory.
-  The test 003-teardown.bats was updated with a new assertion.
- The test confirms that after run_teardown is called, the netavark-dhcp-proxy service log contains the message "Successfully sent DHCPRELEASE for <MAC_ADDRESS>".
- This approach of checking the proxy's own log was chosen because the virtual bridge in the test environment does not reliably forward the broadcast DHCPRELEASE packet to the dnsmasq listener. The proxy log provides definitive proof that the send logic was triggered correctly.

fixes #1271 

### Help Needed: Blocked on Testing

I am opening this as a **Draft PR** because I am currently blocked on testing due to persistent environmental issues in my WSL Fedora setup. I've attempted several testing architectures (macvlan, unmanaged bridge with a containerized DHCP server, etc.) but have been consistently blocked by WSL-specific networking and `systemd` user session problems.

The code is ready for a logical review, and I would greatly appreciate it if a maintainer with a stable test environment could help verify the functionality or suggest a reliable testing setup for WSL.

## Summary by Sourcery

Enable sending DHCPRELEASE on container network teardown to free up IP leases immediately

New Features:
- Add release_lease to DhcpV4Service to send DHCPRELEASE using the mozim client
- Update teardown handler to invoke release_lease for cached leases in a best-effort manner

Enhancements:
- Log lease release failures as warnings without aborting the teardown process

## Summary by Sourcery

Enable netavark to send DHCPRELEASE when tearing down container networks to prevent stale leases and DNS records

New Features:
- Send DHCPRELEASE on container network teardown to proactively free IP leases
- Add release_lease method to DhcpV4Service to wrap mozim client’s release functionality
- Invoke release_lease in the teardown handler in a best-effort manner

Enhancements:
- Log DHCPRELEASE attempts with debug messages and warn on failures without aborting teardown

Tests:
- Update 003-teardown.bats to assert the proxy log contains the success message for DHCPRELEASE

### DIFF
--- a/src/commands/dhcp_proxy.rs
+++ b/src/commands/dhcp_proxy.rs
@@ -38,6 +38,8 @@ use tonic::{
     transport::Server, Code, Code::Internal, Code::InvalidArgument, Request, Response, Status,
 };
 
+type TaskData = (Arc<tokio::sync::Mutex<DhcpV4Service>>, AbortHandle);
+
 #[derive(Debug)]
 /// This is the tonic netavark proxy service that is required to impl the Netavark Proxy trait which
 /// includes the gRPC methods defined in proto/proxy.proto. We can store a atomically referenced counted
@@ -59,7 +61,7 @@ struct NetavarkProxyService<W: Write + Clear> {
     timeout_sender: Option<Arc<Mutex<Sender<i32>>>>,
     // All dhcp poll will be spawned on a new task, keep track of it so
     // we can remove it on teardown. The key is the container mac.
-    task_map: Arc<Mutex<HashMap<String, AbortHandle>>>,
+    task_map: Arc<Mutex<HashMap<String, TaskData>>>,
 }
 
 impl<W: Write + Clear> NetavarkProxyService<W> {
@@ -136,8 +138,8 @@ impl<W: Write + Clear + Send + 'static> NetavarkProxy for NetavarkProxyService<W
         };
     }
 
-    /// When a container is shut down this method should be called. It will clear the lease information
-    /// from the caching system.
+    /// When a container is shut down this method should be called. It will release the
+    /// DHCP lease and clear the lease information from the caching system.
     async fn teardown(
         &self,
         request: Request<NetworkConfig>,
@@ -149,12 +151,25 @@ impl<W: Write + Clear + Send + 'static> NetavarkProxy for NetavarkProxyService<W
         let cache = self.cache.clone();
         let tasks = self.task_map.clone();
 
-        let task = tasks
-            .lock()
-            .expect("lock tasks")
-            .remove(&nc.container_mac_addr);
-        if let Some(handle) = task {
-            handle.abort();
+        let maybe_service_arc = {
+            // Scope for the std::sync::MutexGuard
+            let mut tasks_guard = tasks.lock().expect("lock tasks");
+
+            if let Some((service_arc, handle)) = tasks_guard.remove(&nc.container_mac_addr) {
+                handle.abort();
+                Some(service_arc)
+            } else {
+                None
+            }
+        };
+        if let Some(service_arc) = maybe_service_arc {
+            let mut service = service_arc.lock().await;
+            if let Err(e) = service.release_lease() {
+                warn!(
+                    "Failed to send DHCPRELEASE for {}: {}",
+                    &nc.container_mac_addr, e
+                );
+            }
         }
 
         // Remove the client from the cache dir
@@ -406,7 +421,7 @@ async fn process_setup<W: Write + Clear>(
     network_config: NetworkConfig,
     timeout: u32,
     cache: Arc<Mutex<LeaseCache<W>>>,
-    tasks: Arc<Mutex<HashMap<String, AbortHandle>>>,
+    tasks: Arc<Mutex<HashMap<String, TaskData>>>,
 ) -> Result<NetavarkLease, Status> {
     let container_network_interface = network_config.container_iface.clone();
     let ns_path = network_config.ns_path.clone();
@@ -422,11 +437,13 @@ async fn process_setup<W: Write + Clear>(
             let mut service = DhcpV4Service::new(network_config, timeout)?;
 
             let lease = service.get_lease().await?;
-            let task = tokio::spawn(process_client_stream(service));
+            let service_arc = Arc::new(tokio::sync::Mutex::new(service));
+            let service_arc_clone = service_arc.clone();
+            let task_handle = tokio::spawn(process_client_stream(service_arc_clone));
             tasks
                 .lock()
                 .expect("lock tasks")
-                .insert(mac.to_string(), task.abort_handle());
+                .insert(mac.to_string(), (service_arc, task_handle.abort_handle()));
             lease
         }
         //V6 TODO implement DHCPv6
--- a/src/dhcp_proxy/dhcp_service.rs
+++ b/src/dhcp_proxy/dhcp_service.rs
@@ -11,8 +11,9 @@ use crate::network::netlink::Route;
 use crate::wrap;
 use log::debug;
 use mozim::{DhcpV4ClientAsync, DhcpV4Config, DhcpV4Lease as MozimV4Lease};
+use std::sync::Arc;
+use tokio::sync::Mutex;
 use tokio_stream::StreamExt;
-
 use tonic::{Code, Status};
 
 /// The kind of DhcpServiceError that can be caused when finding a dhcp lease
@@ -39,6 +40,7 @@ impl DhcpServiceError {
 }
 
 /// DHCP service is responsible for creating, handling, and managing the dhcp lease process.
+#[derive(Debug)]
 pub struct DhcpV4Service {
     client: DhcpV4ClientAsync,
     network_config: NetworkConfig,
@@ -129,6 +131,28 @@ impl DhcpV4Service {
             "Could not find a lease within the timeout limit".to_string(),
         ))
     }
+
+    /// Sends a DHCPRELEASE message for the given lease.
+    /// This is a "best effort" operation and should not block teardown.
+    pub fn release_lease(&mut self) -> Result<(), DhcpServiceError> {
+        if let Some(lease) = &self.previous_lease {
+            debug!(
+                "Attempting to release lease for MAC: {}",
+                &self.network_config.container_mac_addr
+            );
+            // Directly call the release function on the underlying mozim client.
+            self.client
+                .release(lease)
+                .map_err(|e| DhcpServiceError::new(Bug, e.to_string()))
+        } else {
+            // No previous lease recorded; nothing to release. Best-effort -> succeed silently.
+            debug!(
+                "No previous lease to release for MAC: {}",
+                &self.network_config.container_mac_addr
+            );
+            Ok(())
+        }
+    }
 }
 
 impl std::fmt::Display for DhcpServiceError {
@@ -149,50 +173,47 @@ impl From<DhcpServiceError> for Status {
     }
 }
 
-pub async fn process_client_stream(mut client: DhcpV4Service) {
-    while let Some(lease) = client.client.next().await {
-        match lease {
+pub async fn process_client_stream(service_arc: Arc<Mutex<DhcpV4Service>>) {
+    let mut client = service_arc.lock().await;
+    while let Some(lease_result) = client.client.next().await {
+        match lease_result {
             Ok(lease) => {
                 log::info!(
                     "got new lease for mac {}: {:?}",
                     &client.network_config.container_mac_addr,
                     &lease
                 );
-                // get previous lease and check if ip addr changed, if not we do not have to do anything
+
                 if let Some(old_lease) = &client.previous_lease {
                     if old_lease.yiaddr != lease.yiaddr
                         || old_lease.subnet_mask != lease.subnet_mask
                         || old_lease.gateways != lease.gateways
                     {
-                        // ips do not match, remove old ones and assign new ones.
                         log::info!(
                             "ip or gateway for mac {} changed, update address",
                             &client.network_config.container_mac_addr
                         );
-                        match update_lease_ip(
+                        if let Err(err) = update_lease_ip(
                             &client.network_config.ns_path,
                             &client.network_config.container_iface,
                             old_lease,
                             &lease,
                         ) {
-                            Ok(_) => {}
-                            Err(err) => {
-                                log::error!("{err}");
-                                continue;
-                            }
+                            log::error!("{err}");
                         }
                     }
                 }
-                client.previous_lease = Some(lease)
+                client.previous_lease = Some(lease);
             }
-            Err(err) => log::error!(
-                "Failed to renew lease for {}: {err}",
-                &client.network_config.container_mac_addr
-            ),
+            Err(err) => {
+                log::error!(
+                    "Failed to renew lease for {}: {err}",
+                    &client.network_config.container_mac_addr
+                );
+            }
         }
     }
 }
-
 fn update_lease_ip(
     netns: &str,
     interface: &str,


src/commands/dhcp_proxy.rs
```diff
@@ -164,6 +164,36 @@ impl<W: Write + Clear + Send + 'static> NetavarkProxy for NetavarkProxyService<W
             .remove_lease(&nc.container_mac_addr)
             .map_err(|e| Status::internal(e.to_string()))?;
 
+        // We have the lease, now try to release it. This is a "best effort"
+        // operation. If it fails, we log the error but do not fail the
+        // teardown process.
+        match mozim::DhcpV4Lease::try_from(lease.clone()) {
+            Ok(mozim_lease) => {
+                // We use a clone of 'nc' for the new service, and the original 'nc' for logging.
+                match DhcpV4Service::new(nc.clone(), self.dora_timeout) {
```
Luap99: This is creating an entirely new socket for dhcp lib. I don't think that is very desirable.

I think it would be best to send the release from within process_client_stream(). We could use a nodify type which allows teardown to send a signal to the function that it should end, i.e. using https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notify_one and then have a tokio::select!{} in the function that either waits for the next lease or the end signal.
When we spawn process_client_stream() you should store the sender inside the task map instead of the abort ahdnle. That way process_client_stream() can exit on its own gracefully.

Or alternately we put the client/DhcpV4Service inside Arc<> share that to the function and then also store it in the task map, then kill the process_client_stream task and then use the lease from the task map and make the release call here.

test-dhcp/003-teardown.bats
```diff
@@ -5,6 +5,13 @@
 
 load helpers
 
+# Redefine start_proxy locally to enable debug logging for this test only.
+# This version will be used instead of the one from helpers.bash.
+function start_proxy() {
+  RUST_LOG=debug ip netns exec "$NS_NAME" $NETAVARK dhcp-proxy --dir "$TMP_TESTDIR" --uds "$TMP_TESTDIR" &>"$TMP_TESTDIR/proxy.log" &
+  PROXY_PID=$!
+}
```
Luap99: I see no reason why we should not have debug log tunred on for all tests. I rather not duplicate this here.

test-dhcp/003-teardown.bats
```diff
@@ -30,6 +37,12 @@ EOF
        assert "$output" == "true"
        # Run teardown
        run_teardown "$input_config"
+       # Check the NETAVARK PROXY log to confirm it successfully SENT the release.
+       # In this test environment, the virtual bridge may not reliably deliver the
+       # broadcast DHCPRELEASE packet to the dnsmasq listener. We check the
+       # proxy's own log for the success message, as this directly confirms
+       # the code path was executed successfully.
+       run_helper grep "Successfully sent DHCPRELEASE for ${CONTAINER_MAC}" "$TMP_TESTDIR/proxy.log"
```
Luap99: > This approach of checking the proxy's own log was chosen because the virtual bridge in the test environment does not reliably forward the broadcast DHCPRELEASE packet to the dnsmasq listener. 

Can you expand on this? I think we also have to check dnsmasq logs, because our logs could just lie and say we send the event without ever doing so.
If we are able to request a lease we should be able to send the release event as well or am I missing something?

Luap99 (commented):
Thanks for the PR

src/commands/dhcp_proxy.rs
Rishikpulhani: @Luap99 I have implemented the alternative approach you suggested by placing the client/DhcpV4Service inside an Arc<> and sharing it across function and storing it in the task map. When you have the time, please review the changes and let me know your thoughts.

test-dhcp/003-teardown.bats
Rishikpulhani: Hi @Luap99, you're right to question this. My initial implementation did exactly what you suggested – I was checking the dnsmasq log for the DHCPRELEASE message.

However, the test was consistently failing. The dnsmasq.log showed a perfect DHCPDISCOVER/OFFER/REQUEST/ACK sequence, but the DHCPRELEASE message never appeared

Initially i thought that maybe the test's cleanup logic is terminating the dnsmasq process before the kernel's virtual network has time to deliver the in-flight DHCPRELEASE packet from the proxy.

To debug this, I instrumented the netavark-dhcp-proxy with detailed logging. The proxy's debug log definitively shows that the release_lease function is executed correctly and that the mozim client successfully sends the release packet to the network socket.

I thought that this is just the case with the test environment and so I just added a check for checking that whether the packet was emitted or not instead of checking the dnsmasq logs 

Is this an acceptable testing strategy given the circumstances, or is there a better way to resolve this race condition in the BATS environment that I might be missing? 

PS: I am actually facing the same issue with my new implementation

Rishikpulhani: Hey @Luap99,   

I went through the codebase of both **Netavark** and **Mozim**, as well as the logs emitted during tests.  

In the Mozim library, I used the following `release` method on the `DhcpV4ClientAsync`:  
<https://docs.rs/mozim/latest/mozim/struct.DhcpV4ClientAsync.html#method.release>  

This internally calls the `release` method on `DhcpV4Client`:  
<https://docs.rs/mozim/latest/src/mozim/dhcpv4/client.rs.html#554>  

When reviewing the proxy logs, I found this **Release** message being sent:  

```
[DEBUG mozim::dhcpv4::msg] DHCP message Message { opcode: BootRequest, htype: Eth, hlen: 6, hops: 0, xid: 1505379826, secs: 0, flags: Flags { broadcast: false }, ciaddr: 10.220.46.58, yiaddr: 0.0.0.0, siaddr: 0.0.0.0, giaddr: 0.0.0.0, chaddr: [106, 108, 253, 24, 156, 237, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], sname: Some([102, 111, 111, 98, 97, 114]), fname: None, magic: [99, 130, 83, 99], opts: DhcpOptions({ServerIdentifier: ServerIdentifier(10.220.46.1), MessageType: MessageType(Release), Hostname: Hostname("foobar"), ClientIdentifier: ClientIdentifier([0, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102])}) }
   [DEBUG mozim::socket] Sending raw ethernet package: [102, 98, 123, 131, 209, 27, 106, 108, 253, 24, 156, 237, 8, 0, 69, 0, 1, 97, 0, 0, 64, 0, 128, 17, 135, 153, 10, 220, 46, 58, 10, 220, 46, 1, 0, 68, 0, 67, 1, 77, 198, 56, 1, 1, 6, 0, 89, 186, 69, 242, 0, 0, 0, 0, 10, 220, 46, 58, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 106, 108, 253, 24, 156, 237, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 102, 111, 111, 98, 97, 114, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 99, 130, 83, 99, 54, 4, 10, 220, 46, 1, 53, 1, 7, 12, 6, 102, 111, 111, 98, 97, 114, 61, 65, 0, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 97, 98, 99, 100, 101, 102, 255]
   [DEBUG mozim::socket] Raw socket sent: 367 bytes
 ```
However, for some reason, this message never reaches the DHCP server.  
It’s not due to the server being stopped — I even added delays in `003-teardown.bats` before the log check:  

```bash
sleep 10
assert `grep -c "Successfully sent DHCPRELEASE for ${CONTAINER_MAC}" "$TMP_TESTDIR/dnsmasq.log"` == 1
```
This had no effect. It seems that although the Mozim client is sending the packet, the DHCP server does not receive it.
My suspicion is that the DHCP server may be dropping these packets.

Any help in identifying what might be going wrong here would be greatly appreciated. Thanks!

Rishikpulhani:
@Luap99 
I'm writing to provide an update on the testing for the DHCPRELEASE feature. I've successfully verified that the implementation is working correctly and wanted to document my findings regarding the test environment.

### Investigation of the Log Behavior
My initial observation was that while the DHCPDISCOVER/OFFER/REQUEST/ACK (DORA) messages were visible in the dnsmasq log, the DHCPRELEASE message was not.

My working hypothesis is that this is due to the nature of the packets and the virtual test network's architecture. The initial DORA messages are initiated via broadcast, which are flooded across the br0 bridge and correctly picked up by the dnsmasq listener. However, the DHCPRELEASE message is a unicast packet, sent directly from the client's IP to the server's IP. It appears that on a Linux virtual bridge, this directed traffic between two endpoints may not be visible to a listener on the bridge device (br0) itself.

Furthermore, my review of the mozim client indicates that it sends the DHCPRELEASE packet in a "fire-and-forget" manner. It considers the operation successful upon sending the packet and does not wait for an acknowledgment, which is compliant with the DHCP standard for this message type.

This behavior supports the conclusion that our primary testing goal should be to verify that the client successfully sends the DHCPRELEASE packet. Verifying receipt by the dnsmasq server is not a reliable metric in this specific test environment. Since both the client and server operate on the same virtual bridge, it is plausible that the unicast DHCPRELEASE message is not delivered to the server's listener. This is a quirk of the test setup and would not occur in a real-world scenario where the client and server reside on different machines

### Verification with tcpdump
To definitively prove the packet was being sent by my code, I captured the traffic directly from the br0 bridge inside the test's temporary network namespace.
The steps taken were:
- Paused the 003-teardown.bats test just before the run_teardown command.
```
# Run teardown
       echo "Pausing for 60 seconds to allow for inspection. Find the netns and run tcpdump now."
       sleep 100
       run_teardown "$input_config"
```
- Identified the active network namespace
```bash
$ sudo ip netns
```
- Executed tcpdump inside the namespace, listening on the br0 interface.
```
$ sudo ip netns exec akWOUvMNBG tcpdump -i br0 -n 'port 67 or port 68' -vv
dropped privs to tcpdump
tcpdump: listening on br0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
17:04:05.775866 IP (tos 0x0, ttl 128, id 0, offset 0, flags [DF], proto UDP (17), length 353)
    10.102.85.56.bootpc > 10.102.85.1.bootps: [udp sum ok] BOOTP/DHCP, Request from 9a:bc:e6:d8:61:25, length 325, xid 0xf1260074, Flags [none] (0x0000)
          Client-IP 10.102.85.56
          Client-Ethernet-Address 9a:bc:e6:d8:61:25
          sname "foobar"
          Vendor-rfc1048 Extensions
            Magic Cookie 0x63825363
            Server-ID (54), length 4: 10.102.85.1
            Client-ID (61), length 65: "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"
            Hostname (12), length 6: "foobar"
            DHCP-Message (53), length 1: Release
tcpdump: pcap_loop: The interface disappeared
1 packet captured
1 packet received by filter
0 packets dropped by kernel
```
This tcpdump output provides definitive proof that the code is correctly generating and sending the unicast DHCPRELEASE packet onto the virtual network. 
Based on this, I have updated the BATS test to verify success by checking for the "Successfully sent DHCPRELEASE..." message in the netavark-proxy.log, as this provides direct proof of the code's execution.

I've detailed my testing methodology and findings above and would be grateful for your review of this approach. Please let me know if you have any feedback or suggestions. Thanks!

Luap99:
@Rishikpulhani Ok I think the package should be able to get received generally, but maybe there is another bug somewhere. I don't really have much time to debug it deeply but I think if we send it and your tcpdump confirms that it is good enough for me now.

src/dhcp_proxy/dhcp_service.rs
```diff
@@ -129,6 +134,19 @@ impl DhcpV4Service {
             "Could not find a lease within the timeout limit".to_string(),
         ))
     }
+
+    /// Sends a DHCPRELEASE message for the given lease.
+    /// This is a "best effort" operation and should not block teardown.
+    pub fn release_lease(&mut self, lease: &MozimV4Lease) -> Result<(), DhcpServiceError> {
```
Luap99: There is no reason to accept the least here, we could simply access self.previous_lease here, that then avoid the need for `previous_lease()` and cloning the lease struct.

Just do this here basically
```
if let Some(lease) = &self.previous_lease {
 // current function body here
}
```

That way the caller doesn't have to do the if let Some() check

src/dhcp_proxy/dhcp_service.rs
```diff
@@ -149,46 +167,61 @@ impl From<DhcpServiceError> for Status {
...
-                            &client.network_config.container_iface,
-                            old_lease,
-                            &lease,
-                        ) {
-                            Ok(_) => {}
-                            Err(err) => {
-                                log::error!("{err}");
-                                continue;
+pub async fn process_client_stream(service_arc: Arc<Mutex<DhcpV4Service>>) {
+    loop {
```
Luap99: could this not just stay as loop like this
```
while let Some(lease) = service_arc.lock().await.client.next().await {
...
}
```

The extra scopes to reduce lock holding is interesting but not actually doing anything since this function is the only one holding the lock anyway.

test-dhcp/003-teardown.bats
```diff
@@ -5,6 +5,11 @@
 
 load helpers
 
+function start_proxy() {
+  RUST_LOG=debug ip netns exec "$NS_NAME" $NETAVARK dhcp-proxy --dir "$TMP_TESTDIR" --uds "$TMP_TESTDIR" &>"$TMP_TESTDIR/proxy.log" &
+  PROXY_PID=$!
+}
```
Luap99: This can be removed now

Rishikpulhani: I have removed it; however, this means there will no longer be any tests related to DHCPRELEASE in the testing suite.

src/dhcp_proxy/dhcp_service.rs
Rishikpulhani: done

Rishikpulhani:
I have incorporated the changes.
PTAL @Luap99

src/dhcp_proxy/dhcp_service.rs
Rishikpulhani: Sorry @Luap99, but this was done to prevent a potential deadlock scenario.

If we use:
```
while let Some(lease) = service_arc.lock().await.client.next().await {
...
}
```
we end up holding the lock on service_arc for the entire loop. Later, inside the Ok arm of the match lease when we need to access the client inside DhcpV4Service, we again need to acquire the lock on service_arc to access the client:
```
match lease {
            Ok(lease) => {
                let mut client = service_arc.lock().await;
                log::info!(
                    "got new lease for mac {}: {:?}",
                    &client.network_config.container_mac_addr,
                    &lease
                );
                // rest of the code
}
```
This would result in a deadlock.

To avoid this, I restructured it into a loop with equivalent stopping conditions using an if/else and break. This way, the lock is released after acquiring the lease and then reacquired only when needed to access the client.

Rishikpulhani:
I’ve incorporated the changes you suggested.
PTAL @Luap99

src/dhcp_proxy/dhcp_service.rs
```diff
@@ -149,50 +173,53 @@ impl From<DhcpServiceError> for Status {
...
+                    );
                 }
-                client.previous_lease = Some(lease)
             }
-            Err(err) => log::error!(
-                "Failed to renew lease for {}: {err}",
-                &client.network_config.container_mac_addr
-            ),
+        } else {
+            break;
```
baude: this orphaned `break` always looks funny to me.  I'm not sure if you can envision a way to make this more readable.  nit only.

baude (commented):
LGTM on a quick review.

src/dhcp_proxy/dhcp_service.rs
Luap99: can't you just put the `let mut client = service_arc.lock().await;` call outside the loop. yes it stays locked for the entire function but that should not matter since there is no other user of the lock here.

```diff
$ git diff
diff --git a/src/dhcp_proxy/dhcp_service.rs b/src/dhcp_proxy/dhcp_service.rs
index 862a47a..1c76f35 100644
--- a/src/dhcp_proxy/dhcp_service.rs
+++ b/src/dhcp_proxy/dhcp_service.rs
@@ -174,49 +174,44 @@ impl From<DhcpServiceError> for Status {
 }
 
 pub async fn process_client_stream(service_arc: Arc<Mutex<DhcpV4Service>>) {
-    loop {
-        let lease_result = service_arc.lock().await.client.next().await;
+    let mut client = service_arc.lock().await;
 
-        if let Some(lease_result) = lease_result {
-            let mut client = service_arc.lock().await;
-            match lease_result {
-                Ok(lease) => {
-                    log::info!(
-                        "got new lease for mac {}: {:?}",
-                        &client.network_config.container_mac_addr,
-                        &lease
-                    );
+    while let Some(lease_result) = client.client.next().await {
+        match lease_result {
+            Ok(lease) => {
+                log::info!(
+                    "got new lease for mac {}: {:?}",
+                    &client.network_config.container_mac_addr,
+                    &lease
+                );
 
-                    if let Some(old_lease) = &client.previous_lease {
-                        if old_lease.yiaddr != lease.yiaddr
-                            || old_lease.subnet_mask != lease.subnet_mask
-                            || old_lease.gateways != lease.gateways
-                        {
-                            log::info!(
-                                "ip or gateway for mac {} changed, update address",
-                                &client.network_config.container_mac_addr
-                            );
-                            if let Err(err) = update_lease_ip(
-                                &client.network_config.ns_path,
-                                &client.network_config.container_iface,
-                                old_lease,
-                                &lease,
-                            ) {
-                                log::error!("{err}");
-                            }
+                if let Some(old_lease) = &client.previous_lease {
+                    if old_lease.yiaddr != lease.yiaddr
+                        || old_lease.subnet_mask != lease.subnet_mask
+                        || old_lease.gateways != lease.gateways
+                    {
+                        log::info!(
+                            "ip or gateway for mac {} changed, update address",
+                            &client.network_config.container_mac_addr
+                        );
+                        if let Err(err) = update_lease_ip(
+                            &client.network_config.ns_path,
+                            &client.network_config.container_iface,
+                            old_lease,
+                            &lease,
+                        ) {
+                            log::error!("{err}");
                         }
                     }
-                    client.previous_lease = Some(lease);
-                }
-                Err(err) => {
-                    log::error!(
-                        "Failed to renew lease for {}: {err}",
-                        &client.network_config.container_mac_addr
-                    );
                 }
+                client.previous_lease = Some(lease);
+            }
+            Err(err) => {
+                log::error!(
+                    "Failed to renew lease for {}: {err}",
+                    &client.network_config.container_mac_addr
+                );
             }
-        } else {
-            break;
         }
     }
 }
```

Luap99:
Oh and one other thing, please fix your commit message so the sign-off is the last line.

The first line it the commit title so commands like `git log --oneline` would only show the sign-off instead of the proper title which makes it harder to read the git logs.

src/dhcp_proxy/dhcp_service.rs
Rishikpulhani: Got it, I made the change

Luap99 (approved):
LGTM

PR #1307, opened 2025-08-11
bridge: use VLAN interface for unmanaged DHCP

### LINKED ISSUE #1294: Unmanaged bridge with VLAN filtering and DHCP not working
Hello all,

I have created a Linux bridge with vlan filtering activated, so that it allows untagged vlan 20 and tagged vlan 40. My goal is to use untagged 20 for the host and tagged 40 for containers. To achieve this I have created a podman network as follows:
```
sudo podman network create -o mode=unmanaged -o com.docker.network.bridge.name=br0 -o vlan=40 --disable-dns --ipv6 --ipam-driver=dhcp vlan40
```
When I create a container with this network, it gets an IP on the untagged vlan:
```
sudo podman network inspect vlan40
[
     {
          "name": "vlan40",
          "id": "5c321f8411b921a2ee147d960ca14ddc8f99f07aa758952cba68f693b38c3bf4",
          "driver": "bridge",
          "network_interface": "br0",
          "created": "2025-07-31T07:46:53.478723003+02:00",
          "ipv6_enabled": true,
          "internal": false,
          "dns_enabled": false,
          "options": {
               "mode": "unmanaged",
               "vlan": "40"
          },
          "ipam_options": {
               "driver": "dhcp"
          },
          "containers": {
               "89e5c2b50b7bd2e71ddfa2079d582ef57d35ce4530d2efebe390415ef83846ff": {
                    "name": "debian",
                    "interfaces": {
                         "eth0": {
                              "subnets": [
                                   {
                                        "ipnet": "10.10.20.41/24",
                                        "gateway": "10.10.20.1"
                                   }
                              ],
                              "mac_address": "ce:43:8d:4b:3a:f9"
                         }
                    }
               }
          }
     }
]
```

I tried to dump some traffic and I came accross this:

```
14:10:07.527530 ce:43:8d:4b:3a:f9 > 33:33:00:00:00:16, ethertype 802.1Q (0x8100), length 94: vlan 40, p 0, ethertype IPv6 (0x86dd), :: > ff02::16: HBH ICMP6, multicast listener report v2, 1 group record(s), length 28
14:10:07.568956 ce:43:8d:4b:3a:f9 > ff:ff:ff:ff:ff:ff, ethertype IPv4 (0x0800), length 376: 0.0.0.0.68 > 255.255.255.255.67: BOOTP/DHCP, Request from ce:43:8d:4b:3a:f9, length 334
14:10:07.744183 ce:43:8d:4b:3a:f9 > 33:33:00:00:00:16, ethertype 802.1Q (0x8100), length 94: vlan 40, p 0, ethertype IPv6 (0x86dd), :: > ff02::16: HBH ICMP6, multicast listener report v2, 1 group record(s), length 28
14:10:07.820961 1c:0b:8b:16:b2:ed > ce:43:8d:4b:3a:f9, ethertype IPv4 (0x0800), length 62: 10.10.20.1 > 10.10.20.41: ICMP echo request, id 58840, seq 0, length 28
14:10:07.820962 1c:0b:8b:16:b2:ed > ce:43:8d:4b:3a:f9, ethertype IPv4 (0x0800), length 342: 10.10.20.1.67 > 10.10.20.41.68: BOOTP/DHCP, Reply, length 300
14:10:07.821233 ce:43:8d:4b:3a:f9 > ff:ff:ff:ff:ff:ff, ethertype IPv4 (0x0800), length 388: 0.0.0.0.68 > 255.255.255.255.67: BOOTP/DHCP, Request from ce:43:8d:4b:3a:f9, length 346
14:10:07.823259 1c:0b:8b:16:b2:ed > ce:43:8d:4b:3a:f9, ethertype IPv4 (0x0800), length 342: 10.10.20.1.67 > 10.10.20.41.68: BOOTP/DHCP, Reply, length 300
14:10:08.196057 ce:43:8d:4b:3a:f9 > 33:33:ff:4b:3a:f9, ethertype 802.1Q (0x8100), length 90: vlan 40, p 0, ethertype IPv6 (0x86dd), :: > ff02::1:ff4b:3af9: ICMP6, neighbor solicitation, who has fe80::cc43:8dff:fe4b:3af9, length 32
14:10:09.220067 ce:43:8d:4b:3a:f9 > 33:33:00:00:00:16, ethertype 802.1Q (0x8100), length 94: vlan 40, p 0, ethertype IPv6 (0x86dd), fe80::cc43:8dff:fe4b:3af9 > ff02::16: HBH ICMP6, multicast listener report v2, 1 group record(s), length 28
14:10:09.220126 ce:43:8d:4b:3a:f9 > 33:33:00:00:00:02, ethertype 802.1Q (0x8100), length 74: vlan 40, p 0, ethertype IPv6 (0x86dd), fe80::cc43:8dff:fe4b:3af9 > ff02::2: ICMP6, router solicitation, length 16
14:10:09.860033 ce:43:8d:4b:3a:f9 > 33:33:00:00:00:16, ethertype 802.1Q (0x8100), length 94: vlan 40, p 0, ethertype IPv6 (0x86dd), fe80::cc43:8dff:fe4b:3af9 > ff02::16: HBH ICMP6, multicast listener report v2, 1 group record(s), length 28
```

It seems as if ipv6 link-local negociation is done on vlan 40, but dhcp request is fired on untagged vlan.

Could you tell me if this is a misunderstanding on my part to properly configure the linux bridge / podman network, or if there's an issue with the dhcp driver.

Thanks in advance !

blackdragoon26:
Fixes #1294.

## What changed

- For unmanaged bridge networks using both DHCP and a VLAN, create or reuse a
  bridge VLAN interface such as `br0.40`.
- Pass that VLAN interface to the existing DHCP proxy instead of passing the
  base bridge.
- Validate a reused interface's kind, parent bridge, VLAN ID, and VLAN
  protocol.
- Match the bridge's 802.1Q or 802.1ad protocol when creating the interface.
- Handle concurrent interface creation and Linux interface-name limits.
- Add an integration topology with untagged VLAN 20 and tagged VLAN 40 DHCP
  servers, including interface reuse and protocol-mismatch coverage.

The existing DHCP proxy/mozim task model is unchanged. This does not introduce
per-container runtimes, network-namespace switching, or a new dependency.

## Root cause

The container veth is configured correctly for the requested VLAN, but the
DHCP proxy opens its raw socket on the host interface supplied by Netavark.
When that interface is the base unmanaged bridge, bridge-originated DHCP
traffic follows the bridge self/default VLAN. In the reported trunk topology,
that allowed a network configured for VLAN 40 to receive a lease from the
untagged VLAN 20 network.

Binding the existing proxy flow to `br0.40` makes the kernel transmit DHCP on
the intended VLAN while retaining the current lease and renewal architecture.

## Interface lifetime

The bridge VLAN interface is shared by all containers using the same unmanaged
bridge and VLAN. It remains after an individual container teardown so the DHCP
proxy can continue lease renewal and later containers can reuse it. Because it
is an upper device of the unmanaged bridge, the kernel removes it when the
lower bridge is removed.

## Verification

- `cargo fmt --all -- --check`
- `cargo check --locked --all-targets`
- `cargo test --locked --lib`
- `cargo +1.97.0 clippy --locked -p netavark -- -D warnings`
- release cross-build for `x86_64-unknown-linux-gnu`
- existing unmanaged bridge VLAN integration test
- focused VLAN/DHCP integration tests:
  - VLAN 40 lease and gateway selection
  - DHCPACK received through the tagged server interface
  - container connectivity to the VLAN 40 gateway
  - interface persistence and reuse
  - rejection of an existing interface with the wrong VLAN protocol


blackdragoon26:
Hihi maintainers, there has been revamp in DHCP lease handling for unmanaged bridge VLANs

This is a major rework of the initial approach to resolve issue #1294 .
Previously, DHCP requests were sourced from the host-side interface, causing
unmanaged bridge VLAN configurations to route DHCP traffic on the wrong VLAN.
This update moves all DHCP traffic generation directly into the target network
namespace.

Key changes:
- Replace async renewal tasks with a dedicated per-lease DHCP worker.
- The worker uses setns() to enter the target namespace, starts a Tokio
runtime, and binds the DHCP client to container_iface instead of host_iface.
- Ensure renewals and rebinds utilize this same namespace-bound worker.
- Implement clean worker shutdown during teardown (replacing task aborts).
- Remove previous VLAN helper experiments, restoring the standard bridge flow.
- Update netavark dhcp-proxy test helper to validate the namespace transition.

blackdragoon26:
@sourcery-ai review

blackdragoon26:
Hi @Luap99  and @flouthoc , can you review the PR.

Luap99 (commented):
sorry I do not have time to properly review this. Overall this approach here feels wrong, creating a new tokio runtime per container does not scale at all and I am not confident in the way the netns switching is done.

I would much prefer it if we add this via support for the mozim lib to only bind the raw socket in the netns there when we actually open the socket and not have to all this here

blackdragoon26:
> sorry I do not have time to properly review this. Overall this approach here feels wrong, creating a new tokio runtime per container does not scale at all and I am not confident in the way the netns switching is done.
> 
> I would much prefer it if we add this via support for the mozim lib to only bind the raw socket in the netns there when we actually open the socket and not have to all this here

Yeah, that makes sense, thanks for that. 
I agree the current per-container runtime/thread plus Netavark-side `setns()` is too heavy and puts namespace switching in the wrong layer.

I’ll rework this by moving the namespace-specific raw socket creation into `mozim`: add an option on `DhcpV4Config ` to create/bind the DHCPv4 raw socket inside the target netns, with the namespace switch scoped only to socket creation and restored immediately afterward.

Then I will update this Netavark PR to remove the custom DHCP worker/runtime changes and just pass the target netns/container iface information into mozim. 
I will link the mozim PR first, then refresh this PR once that support is available.

blackdragoon26:
Per the feedback about keeping the namespace-sensitive socket work in mozim, I split that out first here:

https://github.com/nispor/mozim/pull/79

That PR adds scoped netns handling only around DHCP socket open/bind, and restores the original netns before returning to async code. I have the Netavark side reworked locally to use the normal task model again, and I’ll update this PR once the mozim direction is reviewed.

cathay4t:
I have the same concern as @Luap99 , I need more detail on what is not working instead of this is how I fix the my problem.

blackdragoon26:
Thanks for pushing back on the previous direction. I reworked this to avoid the netns/runtime approach entirely. @cathay4t 

The issue I reproduced is that for an unmanaged bridge with `vlan=40`, Netavark was asking the DHCP proxy to bind to the bridge itself. DHCP traffic sourced from the bridge can use the bridge self/default VLAN, so in a trunk setup it can obtain a lease from the wrong VLAN. In my repro, the container veth was correctly configured for VLAN 40, but DHCP got a lease from the untagged VLAN 20 network.

This update keeps the DHCP `proxy/mozim` flow unchanged. For unmanaged bridge networks with both `vlan=<id>` and DHCP, Netavark now uses a bridge VLAN interface for DHCP, e.g. `br0.40`, creating it if needed and validating it if it already exists. That makes the DHCP socket bind to an interface representing the intended VLAN while keeping renewals in the existing proxy task model.

I also added an integration test that sets up:
- untagged VLAN 20 DHCP range
- tagged VLAN 40 DHCP range
- unmanaged bridge network configured with `vlan=40`
- assertion that Netavark receives the VLAN 40 lease/gateway and dnsmasq sees the DHCPACK on `srv0.40`

Local verification on x86 Linux:
- `cargo check --bins`
- `cargo build --bins`
- `cargo test --lib`
- `cargo clippy --bins -- -D warnings`
- `make validate`
- manual repro for missing `br0.40`, pre-existing `br0.40`, and teardown

blackdragoon26:
@Luap99 , Can you please have a look in this PR now.

blackdragoon26:
> @Luap99 , Can you please have a look in this PR now.

@Luap99 , pinging you for the above, thanks.

blackdragoon26:
Hihi @Luap99 and @flouthoc , gentle reminder, for this PR review.

blackdragoon26:
> Hihi @Luap99 and @flouthoc , gentle reminder, for this PR review.

pinging for the above, 
Thank You

Luap99:
I do not have any time for this right now sorry

blackdragoon26:
> I do not have any time for this right now sorry

Oh okay, I will wait then.

blackdragoon26:
Hi @Luap99 , Re-Pinging you after around month.

blackdragoon26:
Hihii @Luap99, @flouthoc 
Yes, its been 350+ days, since this PR has opened :), and I have reworked the implementation several times in response to the feedback received. Thank you for the earlier pushback on my approach of thinking,  eventually it helped narrow the changes for good.

The  current  version PR is completely revamped with the newer version of repo changes. I believe the current implementation is the appropriate direction because it keeps the existing DHCP proxy/mozim architecture unchanged.

For unmanaged bridge networks using DHCP with a VLAN, Netavark now creates or reuses a matching bridge VLAN interface such as `br0.40` and supplies it as `host_iface` to the existing DHCP proxy. The kernel therefore transmits DHCP on the configured VLAN, while the existing lease, renewal, and container-interface configuration paths remain unchanged.

Also , the implementation also validates the parent bridge, VLAN ID, and VLAN protocol; handles concurrent interface creation; and documents the shared interface lifetime.

And, I really want to close this one nicely, so I made this diagram via excalidraw to share my proposed changed of this PR

image

There are focused integration test reproduces untagged VLAN 20 and tagged VLAN 40 DHCP networks, which verifies that the container receives the VLAN 40 lease, checks connectivity and the DHCPACK path, verifies interface reuse, and covers protocol mismatch rejection.

I would deeply appreciate another review, particularly regarding the shared VLAN-interface lifecycle for an unmanaged bridge. I now feel this is the most contained and maintainable way to address #1294

blackdragoon26:
@Luap99 @flouthoc , Kindly review my new PR.
Regards

PR #1313, opened 2025-08-15
Packit: fetch copr rpm version from Cargo.toml

lsm5:
Current installation of nv from podman-next is failing because it's expecting aardvark-dns >= 102:1.16 while current av on podman-next is only 102:1.15 . From the latest build log, that seems to be because Packit is determining 1.15.0 as the version for av and 1.16.0 as the version for nv.

This commit includes a script to fetch the version directly from Cargo.toml. A similar script will be added to av, thus ensuring both nv and av major.minor versions are in sync for copr rpms as well.

Related: https://github.com/containers/aardvark-dns/pull/625/

## Summary by Sourcery

Ensure Copr RPM packaging uses the Cargo.toml version by adding a Packit fix-spec-file action and script to generate source archives and update the spec file accordingly

Enhancements:
- Add git-archive-all to srpm build dependencies
- Invoke a fix-spec-file action in .packit.yaml to run a custom script during Copr builds
- Introduce .packit-copr-rpm.sh to extract the version from Cargo.toml, generate a source tarball, and patch the RPM spec (Version, Release, Source0, %autosetup) for Copr

### DIFF
--- a/.packit-copr-rpm.sh
+++ b/.packit-copr-rpm.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+# This script handles any custom processing of the spec file using the `fix-spec-file`
+# action in .packit.yaml. These steps only work on copr builds, not on official
+# Fedora builds.
+
+set -uexo pipefail
+
+PACKAGE=netavark
+SPEC_FILE=rpm/"$PACKAGE".spec
+
+# Get Version from HEAD
+VERSION=$(grep '^version' Cargo.toml | cut -d\" -f2)
+
+# RPM Version can't take "-"
+RPM_VERSION="${VERSION//-/\~}"
+
+# Generate source tarball from HEAD
+git-archive-all -C "$(git rev-parse --show-toplevel)" --prefix="$PACKAGE"-"$VERSION"/ rpm/"$PACKAGE"-"$VERSION".tar.gz
+
+# RPM Spec modifications
+
+# Use the Version from HEAD in rpm spec
+sed -i "s/^Version:.*/Version: $RPM_VERSION/" "$SPEC_FILE"
+
+# Use Packit's supplied variable in the Release field in rpm spec.
+sed -i "s/^Release:.*/Release: $PACKIT_RPMSPEC_RELEASE%{?dist}/" "$SPEC_FILE"
+
+# Use above generated tarball as Source in rpm spec
+sed -i "s/^Source0:.*.tar.gz/Source0: $PACKAGE-$VERSION.tar.gz/" $SPEC_FILE
+
+# Don't need Source1 for copr builds
+sed -i "/^Source1/d" "$SPEC_FILE"
+
+# Update setup macro to use the correct build dir
+sed -i "s/^%autosetup.*/%autosetup -Sgit -n %{name}-$VERSION/" "$SPEC_FILE"
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -15,8 +15,14 @@ packages:
 
 srpm_build_deps:
   - cargo
+  - git-archive-all
   - make
   - openssl-devel
+
+# https://packit.dev/docs/configuration/actions
+# fix-spec-file only operates on copr builds and doesn't affect official distro builds
+actions:
+  fix-spec-file: "bash .packit-copr-rpm.sh"
 
 jobs:
   - job: copr_build


Luap99:
> From the latest build log, that seems to be because Packit is determining 1.15.0 as the version for av and 1.16.0 as the version for nv.

Do you know why/how it looks up the version? Because I did the releases in the exact same way on both repos so it seems off to see different behaviors on podman-next.

lsm5:
> > From the latest build log, that seems to be because Packit is determining 1.15.0 as the version for av and 1.16.0 as the version for nv.
> 
> Do you know why/how it looks up the version? Because I did the releases in the exact same way on both repos so it seems off to see different behaviors on podman-next.

It used to get whatever's found in `git describe` , but I see it to be the same on both nv and av on running it locally. Plus I also saw some tag fetching from release-monitoring.org in the recent copr builds which I don't think happened earlier. I can trace it further with the packit team if we need to, but hopefully with the additional action it shouldn't be a problem.

Luap99:
> > > From the latest build log, that seems to be because Packit is determining 1.15.0 as the version for av and 1.16.0 as the version for nv.
> > 
> > 
> > Do you know why/how it looks up the version? Because I did the releases in the exact same way on both repos so it seems off to see different behaviors on podman-next.
> 
> It used to get whatever's found in `git describe` , but I see it to be the same on both nv and av on running it locally. Plus I also saw some tag fetching from release-monitoring.org in the recent copr builds which I don't think happened earlier. I can trace it further with the packit team if we need to, but hopefully with the additional action it shouldn't be a problem.

Yeah I don't mind fixing the spec, I leave that to you as package maintainer to decide what the best thing to do is.

I only want to know *why* we see different behavior given I did both releases in the exact same way branching wise so if copr uses different versions that seem highly unexpected to me.

lsm5:
@Luap99 PTAL

.packit-copr-rpm.sh
```diff
@@ -0,0 +1,38 @@
...
+# Use the Version from HEAD in rpm spec
+sed -i "s/^Version:.*/Version: $RPM_VERSION/" "$SPEC_FILE"
+
+# Use Packit's supplied variable in the Release field in rpm spec.
+# If PACKIT_RPMSPEC_RELEASE is unset or empty, default to 1
+: "${PACKIT_RPMSPEC_RELEASE:=1}"
+sed -i "s/^Release:.*/Release: $PACKIT_RPMSPEC_RELEASE%{?dist}/" "$SPEC_FILE"
+
+# Use above generated tarball as Source in rpm spec
+sed -i -E "s/^Source0:[[:space:]]*[^#[:space:]]+\.(tar\.gz|tar\.bz2|tar\.xz|zip|tgz|tbz2|txz)[[:space:]]*(#.*)?$/Source0: $PACKAGE-$VERSION.tar.gz \2/" "$SPEC_FILE"
```
Luap99: Do we need to overcomplicate the regex like that, i.e why do we match all the possible suffixes? Just matching
`^Source0:.*` seems good enough no?

.packit-copr-rpm.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# RPM Spec modifications
+
+# Use the Version from HEAD in rpm spec
+sed -i "s/^Version:.*/Version: $RPM_VERSION/" "$SPEC_FILE"
+
+# Use Packit's supplied variable in the Release field in rpm spec.
+# If PACKIT_RPMSPEC_RELEASE is unset or empty, default to 1
+: "${PACKIT_RPMSPEC_RELEASE:=1}"
+sed -i "s/^Release:.*/Release: $PACKIT_RPMSPEC_RELEASE%{?dist}/" "$SPEC_FILE"
```
Luap99: I am not very familiar with this syntax `: "${PACKIT_RPMSPEC_RELEASE:=1}"`, I mean it works so it is fine but isn't the common way to only set `${PACKIT_RPMSPEC_RELEASE:-1}` in the sed line below?

.packit-copr-rpm.sh
lsm5: Reverted. It was a sourcery suggestion. Ya, given that packit copr builds will be the only thing to use that script, no reason to complicate it.

.packit-copr-rpm.sh
lsm5: Reverted. Sourcery suggestion again and it worked. Gemini told me the `:-1` is a 1-time temporary replacement only while :=1 actually ends up replacing the variable value, so it seemed like a better choice. But anyway, in packit copr jobs, that variable is *always* set and this script is only used there, so yes, no practical point doing this.

Luap99: It is one time yes, but given we refer the var exactly once that seems ok. Anyhow I guess I learned `:=` is a thing so at least I learned something new

PR #1351, opened 2025-11-03
refactor(network): Rename netlink module to netlink_route

shivkr6:
The current `netlink.rs` module only handles `NETLINK_ROUTE` operations. We're about to add support for `NETLINK_NETFILTER` for conntrack, and putting both in the same file would be a bad idea.

This commit renames `netlink.rs` to `netlink_route.rs` to make its purpose clear before the new code is added.

The following changes were made:
- Renamed `src/network/netlink.rs` to `src/network/netlink_route.rs`.
- Renamed the primary `struct Socket` to `struct RouteSocket` to avoid naming conflicts with future netlink sockets.
- Updated all module paths and `use` statements to match.

This is a pure refactoring and introduces no functional changes.

### DIFF
--- a/src/network/netlink_route.rs
+++ b/src/network/netlink_route.rs
@@ -24,7 +24,7 @@ use netlink_packet_route::{
 };
 use netlink_sys::{protocols::NETLINK_ROUTE, SocketAddr};
 
-pub struct Socket {
+pub struct RouteSocket {
     socket: netlink_sys::Socket,
     sequence_number: u32,
     ///  buffer size for reading netlink messages, see NLMSG_GOODSIZE in the kernel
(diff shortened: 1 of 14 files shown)


src/network/netlink_route.rs
```diff
@@ -24,7 +24,7 @@ use netlink_packet_route::{
 };
 use netlink_sys::{protocols::NETLINK_ROUTE, SocketAddr};
 
-pub struct Socket {
+pub struct RouteSocket {
```
Luap99, line 27: would you not need the same/a similar struct for netfilter?

AFAICT the only difference here would be the protocol we open the socket with. So I think it would be more logical to keep that named Socket and accept a generic marker field as socket family which is then used in new() to open the right socket and implement the right functions. That would allow the same struct to be reused.

So overall I think keep netlink.rs with Socket, add a generic protocol param, implement new(),send(),recv() for all generics and then add a specific netlink_route.rs which holds the implementations for impl Socket<NETLINK_ROUTE> which has the current functions basically.

Does that make sense?

shivkr6: I thought that would be complex, but yeah, that is the most DRY approach. Makes sense, thanks!

PR #1353, opened 2025-11-05
Iptables support removal

l0rd:
Removal of the source code and documentation that supported iptables as a firewall driver.

### DIFF
--- a/docs/netavark-firewalld.7.md
+++ b/docs/netavark-firewalld.7.md
@@ -18,7 +18,7 @@ There is also a dedicated firewalld driver in Netavark.
 This driver uses the firewalld DBus API to natively interact with firewalld.
 It can be enabled by setting `firewall_driver` to `firewalld` in `containers.conf`.
 The native firewalld driver offers better integration with firewalld, but presently suffers from several limitations.
-It does not support isolation (the `--opt isolate=` option to `podman network create`.
+It does not support isolation (the `--opt isolate=` option to `podman network create`).
 Connections to ports forwarded by a container on the same host can only be made through the IPv4 localhost IP (`127.0.0.1`).
 Using other IPs on the host will not work, unless the connection comes from a separate host.
 
--- a/test-dhcp/README.md
+++ b/test-dhcp/README.md
@@ -3,9 +3,10 @@
 ## Running tests
 
 To run the tests locally in your sandbox, you can use one of these methods:
-* bats ./test/001-basic.bats  # runs just the specified test
-* bats ./test/                # runs all
+* bats ./test-dhcp/001-basic.bats  # runs just the specified test
+* bats ./test-dhcp/                # runs all
 
 ## Requirements
 - bats
 - jq
+- dnsmasq
--- a/test/200-bridge-firewalld.bats
+++ b/test/200-bridge-firewalld.bats
@@ -386,7 +386,7 @@ function setup() {
 function strict_port_forwarding_enabled_should_deny_port_forwarding() {
     local fw_driver="$1"
     if [ -z "$fw_driver" ]; then
-        echo "Error: No fw_driver provided." >&2 
+        echo "Error: No fw_driver provided." >&2
         return 1
     fi
 
@@ -402,7 +402,7 @@ function strict_port_forwarding_enabled_should_deny_port_forwarding() {
 function strict_port_forwarding_disabled_should_allow_port_forwarding() {
     local fw_driver="$1"
     if [ -z "$fw_driver" ]; then
-        echo "Error: No fw_driver provided." >&2 
+        echo "Error: No fw_driver provided." >&2
         return 1
     fi
 
@@ -421,7 +421,7 @@ function strict_port_forwarding_disabled_should_allow_port_forwarding() {
 function strict_port_forwarding_invalid_value_should_warn_and_allow_port_forwarding() {
     local fw_driver="$1"
     if [ -z "$fw_driver" ]; then
-        echo "Error: No fw_driver provided." >&2 
+        echo "Error: No fw_driver provided." >&2
         return 1
     fi
 


test/100-bridge-iptables.bats
Luap99, line 1: I don't think it is that simple, there is a fair amount of general tests that we only added to this file but are not fw specific so we must careful review them and cross check them against test/250-bridge-nftables.bats and test/200-bridge-firewalld.bats. I guess best case we add a new test file for the bridge test that are independent of the firewall driver.

rpm/netavark.spec
```diff
@@ -14,14 +14,9 @@
 # Minimum X.Y dep for aardvark-dns
 %define major_minor %((v=%{version}; echo ${v%.*}))
 
-# Set default firewall to nftables on CentOS Stream 10+, RHEL 10+, Fedora 41+
-# and default to iptables on all other environments
+# Set default firewall to nftables on all platforms
 # The `rhel` macro is defined on CentOS Stream, RHEL as well as Fedora ELN.
-%if (%{defined rhel} && 0%{?rhel} >= 10) || (%{defined fedora} && 0%{?fedora} >= 41)
 %define default_fw nftables
```
Luap99: I think we can make this even simpler, we default to nftables and I don't see that changing so we can remove all this logic from the specfile since the actual upstream default is nftables now

Luap99 (changes requested):
Thanks, overall this looks right but I didn't look super carefully. But note the testing comment, we need to make sure we do not remove important test cases that do not directly relate to iptables.

test/100-bridge-iptables.bats
l0rd: ok makes sense 👍

l0rd: @Luap99 there are only 3 tests in `100-bridge-iptables.bats` that don't have a corresponding test in `250-bridge-nftables.bats`:
- `dns with default drop policy`
- `dns with default drop policy with non-default dns port`
- `port forwarding with two networks and RPF - tcp`

I will adapt them so that we still cover the use cases, but I would rather add those to `250-bridge-nftables.bats` rather than creating a new test file. Are you ok with this?

Also, in `100-bridge-iptables.bats`, the variable `NETAVARK_FW` was not set (`iptables` was the default), but now I should unset it in `250-bridge-nftables.bats`.

Luap99: > I will adapt them so that we still cover the use cases, but I would rather add those to 250-bridge-nftables.bats rather than creating a new test file. Are you ok with this?

sure

> Also, in 100-bridge-iptables.bats, the variable NETAVARK_FW was not set (iptables was the default), but now I should unset it in 250-bridge-nftables.bats.

I think it is actually best to keep that. I think not setting it was wrong because we do offer the option do have a different default at compile time, so anyone using that would be unable to run the tests successfully.

I don't see a strong reason to test the default, after all it is a config setting and we just care about the behaviours of each firewall driver.

rpm/netavark.spec
l0rd: Fixed

test/100-bridge-iptables.bats
l0rd: Ok I have added a new commit to add the missing tests cases to `250-bridge-nftables.bats`. [The link](https://editor.mergely.com/wfSUAHyU) to the diff between the deleted  `100-bridge-iptables.bats` and the new `250-bridge-nftables.bats`.

test/250-bridge-nftables.bats
```diff
@@ -351,6 +358,94 @@ export NETAVARK_FW=nftables
...
+    # check that random DNS port was added to nftables rules
+    run_in_host_netns nft list chain inet netavark NETAVARK-HOSTPORT-DNAT
+    assert "${lines[2]}" =~ "ip6 daddr fd10:88:a::1 meta l4proto \{ tcp, udp \} th dport 53 dnat ip6 to \[fd10:88:a::1\]:$dns_port" "DNS forward rule ip6"
+    assert "${lines[3]}" =~ "ip daddr 10.89.3.1 meta l4proto \{ tcp, udp \} th dport 53 dnat ip to 10.89.3.1:$dns_port" "DNS forward rule ip4"
+
+    run_in_host_netns nft add chain inet netavark INPUT \{ type filter hook input priority 0 \; policy drop \; \}
+    run_in_host_netns nft add rule inet netavark INPUT ip saddr 10.89.3.0/24 meta l4proto \{ tcp, udp \} th dport $dns_port accept
+    run_in_host_netns nft add rule inet netavark INPUT ip6 saddr fd10:88:a::/64 meta l4proto \{ tcp, udp \} th dport $dns_port accept
+    run_in_host_netns nft add rule inet netavark INPUT ct state related,established accept
+    run_in_host_netns nft add rule inet netavark INPUT meta l4proto ipv6-icmp accept # allow ICMPv6, required for DNS resolution
```
Luap99, line 416: I am not sure what value this adds practically, you are basically adding the accept rule for dns manually which is not what we want to test but also...

In general there is a big difference between how iptables and nftables work. In iptables all shared the same rule namespace so the default drop there caused us to not be able to get dns packages, so we just add to inject one accept rule before the the final drop happens.
With nftables it is different however because generally each tool uses its own unique table like we do with netavark. So no user should ever touch our INPUT rule and set to to drop as we "own" it.

What is more likely to happen is that a user has another table with an INPUT chain that drops our packages and in the that case it doesn't matter that we accept it in our table. As long as one chain in another table drops the package it gets dropped here so really I don't think us adding these rules even provides much purpose atm.

So now that I thought about all that I guess the answer is our current ruleset doesn't really makes sense and it needs to be rework for 2.0 so I guess we can keep this in for now. It doesn't do any harm either.
cc @mheon

Luap99:
@mheon PTAL

Luap99:
(needs a rebase)

Luap99:
@mheon @baude PTAL

mheon:
LGTM

test/250-bridge-nftables.bats
mheon: I'm generally going to say that, if it's not doing harm, is it worth changing? It doesn't cost us anything, and we're running out of time before 2.0

Luap99: well overall I would still love set based nftables ruleset but yeah given the timelines maybe something we can make happen at a later point

PR #1362, opened 2025-11-18
firewall: flush stale UDP conntrack entries on port_forward setup/teardown

### LINKED ISSUE #1045: Container with podman network not receiving UDP traffic
### Issue Description

Upon running a simple python server container listening on a UDP socket with an attached podman network, UDP traffic that is being sent to the port does not arrive.

Versions 5.2.0-dev-5d10f77da and 4.9.4-rhel both were tried with the same results.

This is a MRE of the issue we are having in production. Docker is fine, podman+cni is fine, podman+netavark exhibits this issue. Note restarting our UDP devices or changing the source port is very cumbersome and we wish to avoid this.

Steps to reproduce the issue
1. Create Dockerfile
```
FROM python:latest
WORKDIR /usr/local/bin
COPY server.py . 
CMD ["chmod", "+x", "server.py"]
CMD ["server.py"]
```
2. The corresponding server script
```
#!/bin/python3
import socket

server_socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
server_socket.bind(('', 17000))

while True:
    message, address = server_socket.recvfrom(1024)
    print(f"resceived from: {address}: {message}", flush = True)
```

3. `podman build . -t podman_udp_test`
4. `podman network create podman_udp`
5. Start sending UDP traffic to port 17000 etc with nping: `nping -g 17580 -p 17000 -c 1000000 --udp 127.0.0.1`
6. `podman run -p 17000:17000/udp --net podman_udp_network podman_udp_testing`

No output

Output from the server after receiving packets

### podman info output

```yaml
host:
  arch: amd64
  buildahVersion: 1.33.8
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - rdma
  - misc
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.10-1.el9.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.10, commit: fb8c4bf50dbc044a338137871b096eea8041a1fa'
  cpuUtilization:
    idlePercent: 99.38
    systemPercent: 0.28
    userPercent: 0.35
  cpus: 4
  databaseBackend: sqlite
  distribution:
    distribution: rhel
    version: "9.4"
  eventLogger: journald
  freeLocks: 2032
  hostname: ccms-pod
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.14.0-427.18.1.el9_4.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 640634880
  memTotal: 8058433536
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    dns:
      package: aardvark-dns-1.10.0-3.el9_4.x86_64
      path: /usr/libexec/podman/aardvark-dns
      version: aardvark-dns 1.10.0
    package: netavark-1.10.3-1.el9.x86_64
    path: /usr/libexec/podman/netavark
    version: netavark 1.10.3
  ociRuntime:
    name: crun
    package: crun-1.14.3-1.el9.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.14.3
      commit: 1961d211ba98f532ea52d2e80f4c20359f241a98
      rundir: /run/user/0/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +YAJL
  os: linux
  pasta:
    executable: ""
    package: ""
    version: ""
  remoteSocket:
    exists: false
    path: /run/podman/podman.sock
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: false
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: true
  serviceIsRemote: false
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.2.3-1.el9.x86_64
    version: |-
      slirp4netns version 1.2.3
      commit: c22fde291bb35b354e6ca44d13be181c76a0a432
      libslirp: 4.4.0
      SLIRP_CONFIG_VERSION_MAX: 3
      libseccomp: 2.5.2
  swapFree: 5367644160
  swapTotal: 5368705024
  uptime: 583h 32m 27.00s (Approximately 24.29 days)
  variant: ""
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  - ipvlan
  volume:
  - local
registries:
  search:
  - registry.access.redhat.com
  - registry.redhat.io
  - docker.io
store:
  configFile: /etc/containers/storage.conf
  containerStore:
    number: 5
    paused: 0
    running: 5
    stopped: 0
  graphDriverName: overlay
  graphOptions:
    overlay.mountopt: nodev,metacopy=on
  graphRoot: /var/lib/containers/storage
  graphRootAllocated: 47173337088
  graphRootUsed: 20552769536
  graphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Supports shifting: "false"
    Supports volatile: "true"
    Using metacopy: "true"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 34
  runRoot: /run/containers/storage
  transientStore: false
  volumePath: /var/lib/containers/storage/volumes
version:
  APIVersion: 4.9.4-rhel
  Built: 1719829634
  BuiltTime: Mon Jul  1 18:27:14 2024
  GitCommit: ""
  GoVersion: go1.21.11 (Red Hat 1.21.11-1.el9_4)
  Os: linux
  OsArch: linux/amd64
  Version: 4.9.4-rhel
```

### Podman in a container

No

### Privileged Or Rootless

Privileged

### Upstream Latest Release

Yes

There was no difference from running nping on localhost versus running it on a different machine that can access the podman container

### Additional information

Starting the python server first and then starting the UDP sender works as expected but this doesn't help our use case.

Stopping and restarting the UDP sender program while the container is running doesn't help. Only by changing the source port of the UDP sender program does traffic start being received, but we cannot easily change the source port of the UDP traffic.

shivkr6:
Add a new netlink_netfilter module to interact with the kernel's conntrack table using the netlink_packet_netfilter crate. This module allows dumping and deleting conntrack entries. All firewall drivers now call the new flush_udp_conntrack() function during port forwarding setup and teardown.

When a container with a UDP port mapping is started, stale conntrack entries can prevent traffic from reaching the new container instance. This change proactively deletes these stale entries for the mapped UDP ports, ensuring that new connections are not dropped by the kernel.

Fixes: https://github.com/containers/netavark/issues/1045

NOTE: This PR cannot be merged right now because:
~~1) I'm waiting for my PRs to be merged in the netlink-packet-netfilter crate.~~ [DONE]
~~2) I have to write integration tests to test this functionality.~~ [DONE]
~~3) netlink-packet-netfilter new release~~ 

CC: @Luap99 @mheon


shivkr6:
The unit tests seem to be failing because conntrack-tools is not installed in CI/CD

shivkr6:
Finished writing the integration tests for flushing conntrack entries.
image

Question: Is it possible to install conntrack-tools in the CI/CD test environment? conntrack-tools is needed for some of the unit tests.
CC: @Luap99 @mheon

shivkr6:
I moved the conntrack entries flushing code to the end of all setup_port_forward and teardown_port_forward functions, instead of keeping it at the start, to avoid a potential race condition where a new conntrack entry is created after the flush but before the port forwarding rules are completely set up.

Luap99:
> Question: Is it possible to install conntrack-tools in the CI/CD test environment? conntrack-tools is needed for some of the unit tests.

Yes we can do that. I make that happen.

Luap99:
https://github.com/containers/netavark/pull/1363 this should include conntrack now

(once that merges you can rebase)

src/firewall/firewalld.rs
```diff
@@ -373,6 +374,16 @@ impl firewall::FirewallDriver for FirewallD {
             setup_portfw.container_id
         );
 
+        if let Some(port_mappings) = setup_portfw.port_mappings {
```
Luap99: This is duplicated for every driver doing the exact same, would it not be better to de duplicate this by moving it into the  setup_firewall() and teardown_firewall() functions in bridge.rs?

That way we don't forget if ever a new driver is added.

src/network/netlink_netfilter.rs
```diff
@@ -0,0 +1,279 @@
...
+/// Fixes: https://github.com/containers/netavark/issues/1045
+pub fn flush_udp_conntrack(
+    port_mappings: &[PortMapping],
+    container_ipv4: Option<IpAddr>,
+    container_ipv6: Option<IpAddr>,
+) -> NetavarkResult<()> {
+    let mut host_ports_to_flush = HashSet::<u16>::new();
+    for pm in port_mappings
+        .iter()
+        .filter(|pm| pm.protocol.to_lowercase().contains("udp"))
```
Luap99: I think we should be very explicit in what we accept, if someone used `audp` that is wrong and certainly would cause an error with the other firewall driver code already.

As such I would match `pm.protocol == "udp"`

src/network/netlink_netfilter.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    for msg in &conntrack_dump {
+        if let Some(flow) = extract_ct_get_flow(msg) {
+            let delete_by_port = if let Some(ref origin) = flow.origin {
+                origin.protocol == Protocol::Udp && host_ports_to_flush.contains(&origin.dst_port)
+            } else {
+                false
+            };
+
+            let delete_by_ip = !delete_by_port && {
```
Luap99: Is there a point in checking `!delete_by_port &&` here? seems like that doesn't do anything because below you check delete_by_port || delete_by_ip

src/network/netlink_netfilter.rs
```diff
@@ -0,0 +1,279 @@
...
+                            ProtoTuple::Protocol(p) => protocol_from_tuple = Some(*p),
+                            ProtoTuple::SourcePort(p) => src_port = Some(*p),
+                            ProtoTuple::DestinationPort(p) => dst_port = Some(*p),
+                            _ => (),
+                        }
+                    }
+                }
+                _ => (),
+            }
+        }
```
Luap99: I am not a fan to nest a function inside another unless there is an advantage in doing so. I don't see one here so I would just move this outside in the normal scope. The function is private anyway,

test/700-udp-traffic-flush.bats
```diff
@@ -0,0 +1,117 @@
...
+        Ok((amt, _src)) => {
+            io::stdout().write_all(&buf[..amt]).unwrap();
+        }
+        Err(_) => {
+            std::process::exit(1);
+        }
+    }
+}
+EOF
+    rustc "$NETAVARK_TMPDIR/udp_listen.rs" -o "$NETAVARK_TMPDIR/udp_listen"
```
Luap99: Don't compile code in tests, this will compile the same thing for every tests and some test env might not want to bring in a full toolchain, i.e. distro level gating where compiling and running tests is often separated.

Either extend netavark-connection-tester in src/test/main.rs without arguments you want or looking at your program use `socat`, not sure on the exact args to make it exit after the first read but something like `socat UDP4-LISTEN:8080 STDIO`  should do it and we have been using socat in other repos already so it should not be an issue.

test/700-udp-traffic-flush.bats
```diff
@@ -0,0 +1,117 @@
...
+    local config_file="${TESTSDIR}/testfiles/bridge-udp-stale-conntrack.json"
+    local msg="payload_$$"
+
+    prepare_udp_listener
+
+    start_sticky_udp_spammer "$msg"
+
+    # avoid race condition and allow conntrack entry creation
+    sleep 0.5
+
```
Luap99, line 40: maybe check the conntrack entries before and after to ensure we have the wrong one removed

test/700-udp-traffic-flush.bats
```diff
@@ -0,0 +1,117 @@
...
+@test "iptables: receive udp traffic with pre-existing stale conntrack entry" {
+    export NETAVARK_FW="iptables"
+    # Explicitly add a rule to trigger connection tracking.
+    # This ensures the traffic generated by the spammer creates a conntrack entry that 
+    # Netavark must flush. Without this, the kernel might not track the flow, preventing 
+    # the reproduction of the stale entry issue (causing a false positive test pass).
+    run_in_host_netns iptables -A OUTPUT -p udp -m state --state NEW -j ACCEPT
+    run_udp_flush_test
+}
```
Luap99, line 58: we are going to remove iptables code in the next release so feel free to remove this iptables code we won't need it

test/700-udp-traffic-flush.bats
```diff
@@ -0,0 +1,117 @@
...
+    wait $SPAMMER_PID || true
+    run_netavark --file "$config_file" teardown $(get_container_netns_path)
+
+    if [ "$container_status" -ne 0 ]; then
+        fail "Container did not receive UDP traffic (listener timeout). The stale conntrack entry likely persisted."
+    fi
+    
+    if [[ "$container_output" != *"$msg"* ]]; then
+        fail "Container received traffic but payload did not match. Expected '$msg', got '$container_output'"
+    fi
```
Luap99: we are using our `assert` helper function to match things which produce more readable tests and errors

`assert "$container_output" =~ "$msg"` "received proper payload"

src/network/netlink_netfilter.rs
shivkr6: From the docs of the `PortMapping struct` in `src/network/types.rs`:
```
    /// Protocol is the protocol forward.
    /// Must be either "tcp", "udp", and "sctp", or some combination of these
    /// separated by commas.
    /// If unset, assumed to be TCP.
    #[serde(rename = "protocol")]
    pub protocol: String,
```
The documentation explicitly states that the string must be either "tcp", "udp", and "sctp", or some combination of these separated by commas.

The problem with matching `pm.protocol == "udp"` is: What if the protocol string contains multiple protocols like `"tcp, udp, sctp"`? UDP traffic would still be dropped because the direct string comparison will not match.

src/network/netlink_netfilter.rs
shivkr6: If we are already deleting that particular conntrack entry because of the port, why bother matching it against the IP? We should delete it regardless. I added this logic hoping the compiler would optimize the check and avoid matching the IP. I think `delete_by_port ||` is more readable, so I will change it to that.

src/network/netlink_netfilter.rs
Luap99: I think the comment must likely be updated, it was copied from podman where it accepts on the user input side comma separated but in our internal db representation we always normalize to avoid this.
I think the firewall drivers assume it can only be single field so it would fail if you actually would use it with a comma.

src/network/netlink_netfilter.rs
Luap99: I think it might be cleaner written then by splitting this filter out in a function then, I thinking something in the form of this:
```
fn filter_matches(flow, protocol,ip,ipv4,ipv6) -> bool {
if protocol == flow.proto && flow.port == port {
    return true
}
if flow.ipv4 == flow.proto {
  return return
}
... more matches you have to do
// no match don't delete
return false
}
```

Luap99:
https://github.com/containers/netavark/pull/1363 was merged so CI should have access to conntrack I hope.

shivkr6:
~~I'll do the remaining changes tomorrow~~ Done.

src/network/netlink_netfilter.rs
```diff
@@ -0,0 +1,278 @@
...
+        attributes
+    }
+    fn is_ipv6(&self) -> NetavarkResult<bool> {
+        match (&self.origin, &self.reply) {
+            (Some(origin), Some(reply)) => {
+                let origin_is_v6 = origin.is_ipv6();
+
+                if origin_is_v6 != reply.is_ipv6() {
+                    Err(NetavarkError::Message(
+                        "The families are different".to_string(),
```
mheon: Nit: if you need to repush, can you change this to "Mismatched IP families in conntrack flow" to make it more clear?

mheon:
I have an error message nit, but on the whole LGTM

test/testfiles/bridge-udp-stale-conntrack.json
```diff
@@ -0,0 +1,32 @@
+{
+   "container_id": "udp_traffic_container",
+   "container_name": "udp_traffic_test",
+   "port_mappings": [
+       {
+           "host_ip": "127.0.0.1",
+           "container_port": 8080,
+           "host_port": 10080,
+           "range": 1,
```
Luap99, line 9: it would likely be nice to have another test with range > 1 to test the iteration is working properly, though not blocking over it.

Cargo.toml
```diff
@@ -47,6 +47,7 @@ nix = { version = "0.30.1", features = ["net", "sched", "signal", "socket", "use
 rand = "0.9.2"
 sha2 = "0.10.9"
 netlink-packet-route = "0.25.1"
+netlink-packet-netfilter = { git = "https://github.com/shivkr6/netlink-packet-netfilter.git", branch = "conntrack-new" }
```
Luap99: just so it is not overlooked, we cannot merge this while the upstream lib isn't merged so I mark as changes requested

test/700-udp-traffic-flush.bats
```diff
@@ -0,0 +1,82 @@
...
+# Regression test for flushing stale conntrack udp entries https://github.com/containers/netavark/issues/1045
+#
+
+load helpers
+
+function start_sticky_udp_spammer() {
+    target_port=10080
+    target_ip=127.0.0.1
+    local message=$1
+
```
Luap99: I didn't notice early as it seems like we don't have a CI test for these files here but there a bunch of trailing white spaces in this file, you can use a command like `git diff --ws-error-highlight=new,old HEAD~1 -- test/700-udp-traffic-flush.bats` to see them

I generally have my editor setup up to delete them on save automatically so maybe you can do the same in yours.

Luap99 (changes requested):

Luap99:
Forgot to say overall great work!

test/700-udp-traffic-flush.bats
shivkr6: Thanks! Installed `shfmt` to do it.

shivkr6:
I have no idea why the `firewalld: receive udp traffic with pre-existing stale conntrack entry (range)` test is failing even after successfully the flushing the conntrack entries. IDK why socat inside the container still cannot receive the udp traffic :(

Please help @Luap99 @mheon

shivkr6:
> Forgot to say overall great work!

Thanks! The appreciation from you and Matt really keeps my motivation and confidence high

shivkr6:
It's extremely weird that the `firewalld: receive udp traffic with pre-existing stale conntrack entry (range)` integration test passes when I change the container_port from 8081 to 8080.
This passes successfully:
```
@test "firewalld: receive udp traffic with pre-existing stale conntrack entry (range)" {
    setup_firewalld
    export NETAVARK_FW="firewalld"
    # Explicitly add a rule to trigger connection tracking.
    run_in_host_netns firewall-cmd --direct --add-rule ipv4 filter OUTPUT 0 -p udp -m state --state NEW -j ACCEPT
    run_ct_udp_flush_test "${TESTSDIR}/testfiles/bridge-udp-stale-conntrack-range.json" "127.0.0.1" "10081" "8080"
}
```

shivkr6:
I skipped the `firewalld: receive udp traffic with pre-existing stale conntrack entry (range)` test because port forwarding range seems to be broken with firewalld. All the traffic gets redirected to the first container port no matter which host range port you pick. It's not an issue related to conntrack.

mheon:
Tests green, just waiting on that upstream library now. Nice work @shivkr6

shivkr6:
Thanks @mheon. The bigger PR, https://github.com/rust-netlink/netlink-packet-netfilter/pull/12, got merged. The remaining smaller PRs yet to be merged are https://github.com/rust-netlink/netlink-packet-netfilter/pull/14 and https://github.com/rust-netlink/netlink-packet-netfilter/pull/15

CC @Luap99

polarathene:
> The remaining smaller PRs yet to be merged are [rust-netlink/netlink-packet-netfilter#14](https://github.com/rust-netlink/netlink-packet-netfilter/pull/14) and [rust-netlink/netlink-packet-netfilter#15](https://github.com/rust-netlink/netlink-packet-netfilter/pull/15)

Those two appear to be merged now. Any other blockers?

**EDIT:** Crate needs to publish a new release, since there doesn't seem to be any activity there pushing that along I've [opened an issue requesting that](https://github.com/rust-netlink/netlink-packet-netfilter/issues/16).

shivkr6:
@polarathene , the maintainer of the rust-netlink crates, said:
“I don't want to release the current codebase as a new version because I dislike the current Subsystem layout.”

He wants the subsystem layout of netlink-packet-netfilter to be similar to netlink-packet-route.
We had this discussion in the rust-netlink Matrix channel.

I’m open to doing the subsystem layout rewrite around February or March 2026; I have some exams right now.

polarathene:
> @polarathene , the maintainer of the rust-netlink crates, said:
> “I don't want to release the current codebase as a new version because I dislike the current Subsystem layout.”

I don't see how that's a release blocker 🤔
- `0.3.0` / `0.2.1` could be released in the interim (_it's been 2 years since that crate really saw any activity until your contributions_)
- New version bump for the subsystem layout refactor.

No worries though, thanks for chiming in with an update! 😁

Luap99:
@polarathene That is not up for us to decide but the maintainer of that crate though.

duck-rh:
@shivkr6 just merged upstream \o/ 🎆

shivkr6:
> @shivkr6 just merged upstream \o/ 🎆

Hi, I'll rebase the PR tomorrow and it should be ready to merge

frispete:
Ohh, looks, like it still has serious troubles to compile...

I haven't looked at any rust code in my life before, but it seems, that the netavark rust interface does not fit the netlink-netfilter git code.

shivkr6:
Should be ready to be merged @Luap99 @mheon

mheon:
Looks like some of the testing-farm machines are missing the conntrack binary

mheon:
I remain LGTM, though we should sort out the testing farm issues before we merge. TYVM @shivkr6

PR #1369, opened 2025-12-03
make network setup order deterministic

Luap99:
Change the file format of the networks json to use an array for the PerNetworkOptions. With that we can get a defined order in which we setup the networks and do that deterministically.

In order to provide backwards compatibility we use a custom json deserializer function that accepts both the old and new format. That allows us to still use the old podman against the new netavark with this patch.

I also added PartialEq trait to the types so the unit test can compare two loaded Networks.

I only converted a few of our test configs to the new format to keep the diff smaller. We can convert most of them another time.


src/network/types.rs
```diff
@@ -78,7 +81,9 @@ pub struct NetworkOptions {
     /// The networks listed in "network_info" have to match this,
     /// both use the network name as key for the map.
     #[serde(rename = "networks")]
-    pub networks: HashMap<String, PerNetworkOptions>,
+    // pub networks: HashMap<String, PerNetworkOptions>,
+    #[serde(deserialize_with = "deserialize_network_options_map_or_vec")]
```
mheon, line 84: I like this a lot... didn't think this would be possible, but retaining compat is very nice.

test/250-bridge-nftables.bats
```diff
@@ -1014,6 +1014,17 @@ net/ipv4/conf/podman1/rp_filter = 2"
     assert_json "$result" 'has("t1")' == "true" "t1 object key exists"
     assert_json "$result" 'has("t2")' == "true" "t2 object key exists"
 
+    # verify the setup order of the contianer interfaces by checking the interface index
```
mheon, line 1017: Might it be worth doing this twice, second time with a different network with the order of the vector swapped, to make sure it's actual order that affects it?

Luap99: yeah good point, I will add a test case with the inverse order in the config to confirm the behavior

src/network/types.rs
Luap99: unlike the go json library the rust serde (and serde_json) one has options for just about anything you can think off. Though this here is just like a custom UnmarshalJSON() in go.

Though I just noticed I need to drop the commented out line

test/250-bridge-nftables.bats
Luap99: done

Luap99:
@baude @mheon PTAL I like to move the card to done.

mheon:
LGTM

PR #1376, opened 2025-12-11
mozim: do not use netlink feature

Luap99:
mozim uses rtnetlink which we do not otherwise. All mozim uses it for is to get the interface index which can lookup easily ourselves.

This saves about 500KB in binary size.


mheon:
LGTM

Luap99:
@baude PTAL

PR #1379, opened 2025-12-15
Support multiple static IPs per subnet

Honny1:
Change from index-based IP/subnet pairing to subnet-aware matching. Now all static IPs that belong to a subnet are assigned to that subnet, allowing multiple IPs per subnet and proper multi-subnet distribution.

Fixes: https://issues.redhat.com/browse/RHEL-98277

### DIFF
--- a/plugin-API.md
+++ b/plugin-API.md
@@ -107,6 +107,14 @@ Example JSON input:
 }
 ```
 
+**Note on `static_ips`**: When multiple subnets are configured in the network, you can specify multiple static IP addresses. Each IP will be assigned to the subnet it belongs to. For example, with subnets `10.88.0.0/16` and `10.89.0.0/16`, you could specify:
+
+```
+"static_ips": ["10.88.0.50", "10.88.0.51", "10.89.0.100"]
+```
+
+All IPs matching a subnet's range will be assigned to that subnet. Networks with duplicate or overlapping subnet definitions will result in an error.
+
 Example JSON output:
 ```
 {
--- a/src/network/core_utils.rs
+++ b/src/network/core_utils.rs
@@ -62,6 +62,161 @@ where
     Ok(Some(val))
 }
 
+fn validate_subnets(subnets: &Option<Vec<types::Subnet>>) -> NetavarkResult<()> {
+    for (i, subnet1) in subnets.iter().flatten().enumerate() {
+        for subnet2 in subnets.iter().flatten().skip(i + 1) {
+            // Check for exact duplicate
+            if subnet1.subnet == subnet2.subnet {
+                return Err(NetavarkError::msg(format!(
+                    "duplicate subnet defined: {}",
+                    subnet1.subnet
+                )));
+            }
+            // Check for overlap (one contains the other)
+            if subnet1.subnet.contains(&subnet2.subnet.network())
+                || subnet2.subnet.contains(&subnet1.subnet.network())
+            {
+                return Err(NetavarkError::msg(format!(
+                    "overlapping subnets: {} and {}",
+                    subnet1.subnet, subnet2.subnet
+                )));
+            }
+        }
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    struct TestCase {
+        name: &'static str,
+        subnets: Option<Vec<(&'static str, &'static str)>>,
+        should_pass: bool,
+        expected_error: Option<&'static str>,
+    }
+
+    #[test]
+    fn test_validate_subnets() {
+        let test_cases = vec![
+            TestCase {
+                name: "empty subnets",
+                subnets: None,
+                should_pass: true,
+                expected_error: None,
+            },
+            TestCase {
+                name: "single subnet",
+                subnets: Some(vec![("10.0.0.0/24", "10.0.0.1")]),
+                should_pass: true,
+                expected_error: None,
+            },
+            TestCase {
+                name: "two non-overlapping subnets",
+                subnets: Some(vec![
+                    ("10.0.0.0/24", "10.0.0.1"),
+                    ("10.0.1.0/24", "10.0.1.1"),
+                ]),
+                should_pass: true,
+                expected_error: None,
+            },
+            TestCase {
+                name: "ipv4 and ipv6 subnets",
+                subnets: Some(vec![("10.0.0.0/24", "10.0.0.1"), ("fd00::/64", "fd00::1")]),
+                should_pass: true,
+                expected_error: None,
+            },
+            TestCase {
+                name: "duplicate subnets",
+                subnets: Some(vec![
+                    ("10.0.0.0/24", "10.0.0.1"),
+                    ("10.0.0.0/24", "10.0.0.1"),
+                ]),
+                should_pass: false,
+                expected_error: Some("duplicate subnet"),
+            },
+            TestCase {
+                name: "overlapping subnets",
+                subnets: Some(vec![
+                    ("10.0.0.0/16", "10.0.0.1"),
+                    ("10.0.1.0/24", "10.0.1.1"),
+                ]),
+                should_pass: false,
+                expected_error: Some("overlapping subnets"),
+            },
+            TestCase {
+                name: "ipv6 overlapping subnets",
+                subnets: Some(vec![
+                    ("fd00::/48", "fd00::1"),
+                    ("fd00:0:0:1::/64", "fd00:0:0:1::1"),
+                ]),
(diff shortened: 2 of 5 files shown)


test/400-ipvlan.bats
```diff
@@ -323,3 +323,74 @@ EOF
    run_in_container_netns ip -o link show
    assert "${#lines[@]}" == 2 "only two interfaces (lo, eth0) in the netns, the tmp ipvlan interface should be gone"
 }
+
+@test "ipvlan multiple static ips single subnet" {
```
vikas-goel, line 327: Will macvlan, too, support it?

Honny1: Yes, it should. Should I add tests for that?

vikas-goel: Yes, please.

plugin-API.md
```diff
@@ -107,6 +107,14 @@ Example JSON input:
 }
 ```
 
+**Note on `static_ips`**: When multiple subnets are configured in the network, you can specify multiple static IP addresses. Each IP will be assigned to the subnet it belongs to. For example, with subnets `10.88.0.0/16` and `10.89.0.0/16`, you could specify:
+
+```
+"static_ips": ["10.88.0.50", "10.88.0.51", "10.89.0.100"]
```
vikas-goel, line 113: Need ability to specify static IP address in CIDR format.

mheon: Are you saying that you need multiple static IPs that are *not in the same subnet*?

vikas-goel: Multiple static IP addresses that part of multiple smaller subnets of a large configured subnet?

mheon: Can you explain what the exact intention is here? I can't really understand the benefits of setting a subnet different from that of the actual network the interface is connected to, even if that subnet is part of the larger subnet the network uses.

test/400-ipvlan.bats
Honny1: I have added a macvlan test cases.

plugin-API.md
Luap99: I agree different subnets on a static ip address allocation is not something I like to support.
The way to do that would be multiple --subnet options on network create.

src/network/core_utils.rs
```diff
@@ -62,6 +62,31 @@ where
     Ok(Some(val))
 }
 
+fn validate_subnets(network: &types::Network) -> NetavarkResult<()> {
```
Luap99: Can you write some rust unit tests for this please. That allows us to cover some more subnet combinations and is faster than dealing with the more expensive integration tests

test/300-macvlan.bats
```diff
@@ -389,3 +389,76 @@ EOF
     assert_json "$default_route_v6" '.[0].dst' == "default" "Default route was selected"
     assert_json "$default_route_v6" '.[0].metric' == "200" "v6 route metric matches v4"
 }
+
+
+
+@test "macvlan multiple static ips single subnet" {
```
Luap99, line 395: There is a ton of duplication between the testfiles here, I would really prefer if we can start deduplicating some of this.

i.e. compare test_port_fw function to generate a config dynamically for tests.

Right now it duplicates all test files for each driver which is not great.

src/network/core_utils.rs
```diff
@@ -112,20 +140,25 @@ pub fn get_ipam_addresses<'a>(
...
-                net_addresses.push(types::NetAddress {
-                    gateway: subnet.gateway,
-                    ipnet: container_address,
-                });
+                // build up response information - add all static IPs that match this subnet
+                for static_ip in static_ips {
+                    // check if the static IP belongs to this subnet
+                    if subnet.subnet.contains(static_ip) {
+                        let container_address: ipnet::IpNet =
+                            match format!("{}/{}", static_ip, subnet_mask_cidr).parse() {
```
Luap99: pre existing but it makes no sense to combine this into a string to parse it again, we can create the subnet directly `ipnet::IpNet::new(ip, prefix_len)`

src/network/core_utils.rs
Honny1: I implemented a unit test.

test/300-macvlan.bats
Honny1: I deduplicated the tests.

Honny1:
PTAL @Luap99

src/network/core_utils.rs
```diff
@@ -62,6 +62,160 @@ where
...
+            },
+            TestCase {
+                name: "two non-overlapping subnets",
+                subnets: Some(vec![
+                    ("10.0.0.0/24", "10.0.0.1"),
+                    ("10.0.1.0/24", "10.0.1.1"),
+                ]),
+                should_pass: true,
+                expected_error: None,
+            },
```
TomSweeneyRedHat, line 123: It might be good to have two non-overlapping ipv6 subnest tested too.

src/network/core_utils.rs
```diff
@@ -62,6 +62,160 @@ where
     Ok(Some(val))
 }
 
+fn validate_subnets(network: &types::Network) -> NetavarkResult<()> {
+    let subnets: Vec<&types::Subnet> = network.subnets.iter().flatten().collect();
+    for (i, subnet1) in subnets.iter().enumerate() {
+        for subnet2 in subnets.iter().skip(i + 1) {
```
TomSweeneyRedHat: Not a rust expert, but would it be possible to use a map for one of these rather than iterating through both arrays?

src/network/core_utils.rs
Honny1: I added IPv6 scenarios.

src/network/core_utils.rs
Honny1: I think for correct inputs, we need to iterate over all unique pairs of subnets. While using a map or hash table would make detecting duplicates faster, we still need to check all pairs to find overlaps. Therefore, for valid inputs, using a map would simply add an extra pass.

src/network/core_utils.rs
```diff
@@ -62,6 +62,181 @@ where
     Ok(Some(val))
 }
 
+fn validate_subnets(network: &types::Network) -> NetavarkResult<()> {
+    let subnets: Vec<&types::Subnet> = network.subnets.iter().flatten().collect();
```
Luap99: this is effectively cloning the vector which seems unnecessary
if you put the loop into `if let Some(subnets) = &network.subnets {`
then we extract the some value as reference without cloning the vector.

src/network/core_utils.rs
```diff
@@ -62,6 +62,181 @@ where
...
+                name: "test".to_string(),
+                network_interface: None,
+                options: None,
+                ipam_options: None,
+                subnets,
+                routes: None,
+                network_dns_servers: None,
+            };
+
+            let result = validate_subnets(&network);
```
Luap99: I think you should just switch the function to accept `subnets: &Option<Vec<types::Subnet>>` Then you do not have to to construct the full network type for each case and we just pass the subnet

src/network/core_utils.rs
```diff
@@ -112,20 +290,27 @@ pub fn get_ipam_addresses<'a>(
...
-                container_addresses.push(container_address);
-                net_addresses.push(types::NetAddress {
-                    gateway: subnet.gateway,
-                    ipnet: container_address,
-                });
+                // build up response information - add all static IPs that match this subnet
+                for static_ip in static_ips {
+                    // check if the static IP belongs to this subnet
+                    if subnet.subnet.contains(static_ip) {
+                        let container_address =
```
Luap99, line 277: What happens if none of the given ips match a subnet, I guess the podman side should validate that normally but I think maybe we should have a check after the loop to ensure the length of container_addresses matches the length of static ips to make sure all were consumed?

mheon:
My original question about why multiple v4/v6 subnets are attached to the same interface remains, but the implementation LGTM, and I won't block a merge based on that question.

src/network/core_utils.rs
Honny1: I have added a check for that, along with the corresponding tests.

src/network/core_utils.rs
Honny1: I have changed the function to accept `subnets: &Option<Vec<types::Subnet>>`.

src/network/core_utils.rs
Honny1: After updating the function to accept `subnets: &Option<Vec<types::Subnet>>`, it appears that line is unnecessary. I am not a Rust expert, so I'm not sure if this implementation is fully optimized.

PR #1405, opened 2026-02-18
DNM cirrus: Enable Spot instances for ARM64 builds

timcoding1988:
Enable Spot pricing for t4g.xlarge instances used in ARM64
builds and tests


Luap99:
I am not a fan of only doing this for a few tasks, AFAIK spot instances work on GCE as well so the question why only focus on a single task?

In general I rather not change all the repos at once if we cannot tell the consequences yet, I merged https://github.com/containers/podman-machine-os/pull/222 so I rather have these sitting here first and let it try out a few weeks in machine-os to see if there are problem over there not with that approach

PR #1417, opened 2026-03-08
routes: Add blackhole route support

### LINKED ISSUE #930: Attaching container to a network with any routes defined results in fail
Creating network with `podman network create --route 10.1.0.0/24,1.1.1.1 --opt no_default_route=1 testserver`:
```json
[
     {
          "name": "testserver",
          "id": "dcc38ca4c22e02f94beb62aea608ef13ad14b466f9815bcc284a78e4da9af3a5",
          "driver": "bridge",
          "network_interface": "podman6",
          "created": "2024-02-19T02:36:20.158376838+01:00",
          "subnets": [
               {
                    "subnet": "10.89.6.0/24",
                    "gateway": "10.89.6.1"
               }
          ],
          "routes": [
               {
                    "destination": "10.1.0.0/24",
                    "gateway": "1.1.1.1"
               }
          ],
          "ipv6_enabled": false,
          "internal": false,
          "dns_enabled": true,
          "options": {
               "no_default_route": "true"
          },
          "ipam_options": {
               "driver": "host-local"
          }
     }
]
```
and then attaching a container to it results in `DEBU[0000] ExitCode msg: "netavark (exit code 1): netlink error: network is unreachable (os error 101)"`. This happens whenever I add any route to that network. Without it, everything works. Worth mentioning network is unavailable on a container after attaching it to this network (even though attach was unsuccessful).

When creating network with `podman network create --opt no_default_route=1 testserver` results in network:
```json
[
     {
          "name": "testserver",
          "id": "04435b5b0e18173c897137ead3387092fc27e9957d005ef6601180ea41597ed4",
          "driver": "bridge",
          "network_interface": "podman7",
          "created": "2024-02-19T10:54:13.700210416+01:00",
          "subnets": [
               {
                    "subnet": "10.89.7.0/24",
                    "gateway": "10.89.7.1"
               }
          ],
          "ipv6_enabled": false,
          "internal": false,
          "dns_enabled": true,
          "options": {
               "no_default_route": "true"
          },
          "ipam_options": {
               "driver": "host-local"
          }
     }
]
```
Attach works fine, network doesn't work as expected due to `no_default_route.`.

My goal is to block access to private network ranges via routes.

Am I doing something wrong or it's possibly netavark bug? Thank you in advance.

p12tic:
When the gateway field is omitted from a route definition, a blackhole route is created instead of a unicast route. Blackhole routes silently discard packets destined for the specified network.

The gateway field in the Route type is now Optional.

Closes #930

### DIFF
--- a/src/network/netlink_route.rs
+++ b/src/network/netlink_route.rs
@@ -43,31 +43,49 @@ pub enum LinkID {
 pub enum Route {
     Ipv4 {
         dest: ipnet::Ipv4Net,
-        gw: Ipv4Addr,
+        gw: Option<Ipv4Addr>,
         metric: Option<u32>,
+        route_type: RouteType,
     },
     Ipv6 {
         dest: ipnet::Ipv6Net,
-        gw: Ipv6Addr,
+        gw: Option<Ipv6Addr>,
         metric: Option<u32>,
+        route_type: RouteType,
     },
 }
 
 impl std::fmt::Display for Route {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let (dest, gw, metric) = match self {
-            Route::Ipv4 { dest, gw, metric } => (
+        let (dest, gw, metric, rtype) = match self {
+            Route::Ipv4 {
+                dest,
+                gw,
+                metric,
+                route_type,
+            } => (
                 dest.to_string(),
-                gw.to_string(),
+                gw.map(|g| g.to_string()),
                 metric.unwrap_or(constants::DEFAULT_METRIC),
+                route_type,
             ),
-            Route::Ipv6 { dest, gw, metric } => (
+            Route::Ipv6 {
+                dest,
+                gw,
+                metric,
+                route_type,
+            } => (
                 dest.to_string(),
-                gw.to_string(),
+                gw.map(|g| g.to_string()),
                 metric.unwrap_or(constants::DEFAULT_METRIC),
+                route_type,
             ),
         };
-        write!(f, "(dest: {dest} ,gw: {gw}, metric {metric})")
+        if let Some(gw) = gw {
+            write!(f, "(dest: {dest}, gw: {gw}, metric {metric})")
+        } else {
+            write!(f, "({rtype:?} dest: {dest}, metric {metric})")
+        }
     }
 }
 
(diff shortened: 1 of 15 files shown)


Luap99 (changes requested):
I don't think that is a great design choice, there are many different route types that could be used in the same spot, i.e. unreachable or prohibit.

I think it would be smarter to add an explicit type field and then let users request blackhole explicitly, that type field should then accept the same names like ip route basically.

p12tic:
@Luap99 Is the current version of the PR something what you had in mind? 

ip route additionally supports `local`, `multicast`, `broadcast`, `throw` and `nat` types. It doesn't make sense to add support for `throw` and `nat` and _probably_ doesn't make sense to add support for the rest. I didn't investigate deeper because in any case there's not even issue opened for these, so probably no one needs it.

p12tic:
This feature has been exposed in podman via https://github.com/containers/podman/pull/28230

p12tic:
@Luap99 Just a friendly ping :)

Luap99:
Note I am super busy, in particular I like to get https://github.com/containers/netavark/pull/1394 merged first then this PR can add the network creation validation logic for this option here as well.

Of course we still need the type changes in container-libs and podman but when we port the creating logic we should not add even more features in parallel to not complicated that work so this may be a bit stalled for a few weeks.

I try to ensure we still get this into netavark v2/podman v6

p12tic:
Thanks for explaining, I just wanted to make sure the PR is not forgotten for no reason :-)

Luap99:
https://github.com/containers/netavark/pull/1394 is merged

This should allow you to rebase and add the create time validation here

p12tic:
@Luap99 I have updated the PR and re-tested together with https://github.com/containers/podman/pull/28230

p12tic:
@Luap99 just a friendly ping :-)

Is there anything left to do with this PR?

src/network/netlink_route.rs
```diff
@@ -43,31 +43,49 @@ pub enum LinkID {
...
                 metric.unwrap_or(constants::DEFAULT_METRIC),
+                route_type,
             ),
         };
-        write!(f, "(dest: {dest} ,gw: {gw}, metric {metric})")
+        if gw.is_empty() {
+            write!(f, "({rtype:?} dest: {dest}, metric {metric})")
+        } else {
+            write!(f, "(dest: {dest}, gw: {gw}, metric {metric})")
+        }
```
Luap99: WDYT about this:

```diff
diff --git a/src/network/netlink_route.rs b/src/network/netlink_route.rs
index 04ca9b00..2bf549e6 100644
--- a/src/network/netlink_route.rs
+++ b/src/network/netlink_route.rs
@@ -1,5 +1,5 @@
 use std::{
-    net::{Ipv4Addr, Ipv6Addr},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr},
     os::fd::{AsFd, AsRawFd, BorrowedFd},
 };
 
@@ -65,7 +65,7 @@ impl std::fmt::Display for Route {
                 route_type,
             } => (
                 dest.to_string(),
-                gw.map_or(String::new(), |g| g.to_string()),
+                gw.map(|g| g.to_string()),
                 metric.unwrap_or(constants::DEFAULT_METRIC),
                 route_type,
             ),
@@ -76,15 +76,15 @@ impl std::fmt::Display for Route {
                 route_type,
             } => (
                 dest.to_string(),
-                gw.map_or(String::new(), |g| g.to_string()),
+                gw.map(|g| g.to_string()),
                 metric.unwrap_or(constants::DEFAULT_METRIC),
                 route_type,
             ),
         };
-        if gw.is_empty() {
-            write!(f, "({rtype:?} dest: {dest}, metric {metric})")
-        } else {
+        if let Some(gw) = gw {
             write!(f, "(dest: {dest}, gw: {gw}, metric {metric})")
+        } else {
+            write!(f, "({rtype:?} dest: {dest}, metric {metric})")
         }
     }
 }
```

I think keeping the option over an empty string to branch seem a bit more intuitive to me.

Luap99 (commented):
Thanks, one minor thing. LGTM overall

cc @mheon @ashley-cui

p12tic:
@Luap99 Thanks for review. Applied your suggestion and rebased on top of latest `main`.

p12tic:
One question for when this PR is being merged - how to integrate this feature in https://github.com/containers/podman/pull/28230, in particular tests? Is it appropriate to do local tests with development version of netavark, leaving the tests requiring newer netavark disabled with a version check? Or should I wait until netavark 2.0 is released before that PR can land?

mheon:
We can spin new images with a Netavark built from source when this merges; all the Podman main stuff is already doing that after the other changes that have merged in preparation for 6

mheon (approved):
LGTM

Luap99 (approved):
LGTM

Luap99:
We have a dev version in podman CI right now due other netavark changes we need. So we just need to rebuild a new CI image but the current one is blocked on some crun failures so we need to wait a bit longer.

@p12tic you do not have to do anything for the time being, I let you know once we have the new netavark in podman CI but it should happen before 6.0 final release

src/network/netlink_route.rs
```diff
@@ -43,31 +43,49 @@ pub enum LinkID {
 pub enum Route {
     Ipv4 {
         dest: ipnet::Ipv4Net,
-        gw: Ipv4Addr,
+        gw: Option<Ipv4Addr>,
         metric: Option<u32>,
+        route_type: RouteType,
```
mtrmac, line 48: A drive-by comment:

Dare I suggest that the gateway should be a field in the `RouteType` enum, instead of independently optional?

PR #1428, opened 2026-03-29
Fix: Add more descriptive netavark error message(s)

davchos:
In this patch, I am attempting to work on issue of #1392.

This PR add more descriptive netavark error message(s).

Fixes: #1392

Signed-off-by: David CHOSROVA <dchosrova@gmail.com>

### DIFF
--- a/src/error/mod.rs
+++ b/src/error/mod.rs
@@ -159,7 +159,18 @@ impl fmt::Display for NetavarkError {
                     Ok(())
                 }
             }
-            NetavarkError::Nftables(e) => write!(f, "nftables error: {e}"),
+            NetavarkError::Nftables(e) => match e {
+                nftables::helper::NftablesError::NftFailed {
+                    program,
+                    hint,
+                    stdout: _,
+                    stderr,
+                } => write!(
+                    f,
+                    "nftables error: {program:?} did not return successfully while {hint}: {stderr}",
+                ),
+                _ => write!(f, "nftables error: {e}"),
+            },
             NetavarkError::SubnetParse(e) => write!(f, "parsing IP subnet error: {e}"),
             NetavarkError::AddrParse(e) => write!(f, "parsing IP address error: {e}"),
         }
--- a/test/250-bridge-nftables.bats
+++ b/test/250-bridge-nftables.bats
@@ -1289,3 +1289,17 @@ function check_simple_bridge_nftables() {
     run_helper ps "$aardvark_pid"
     assert "${lines[1]}" =~ ".*aardvark-dns --config $NETAVARK_TMPDIR/config/aardvark-dns -p 53 run" "aardvark not running or bad options"
 }
+
+@test "$fw_driver - nft error" {
+    local nft="$NETAVARK_TMPDIR/nft"
+    cat > "$nft" <<EOF
+#!/usr/bin/env bash
+echo 'nft custom error message' 1>&2
+exit 1
+EOF
+
+    chmod +x $nft
+
+    PATH="$NETAVARK_TMPDIR:$PATH" expected_rc=1 run_netavark --file ${TESTSDIR}/testfiles/simplebridge.json setup $(get_container_netns_path)
+    assert_json ".error" 'nftables error: "nft" did not return successfully while getting the current ruleset: nft custom error message' "error message from nft is included"
+}


src/error/mod.rs
```diff
@@ -159,7 +159,18 @@ impl fmt::Display for NetavarkError {
                     Ok(())
                 }
             }
-            NetavarkError::Nftables(e) => write!(f, "nftables error: {e}"),
+            NetavarkError::Nftables(e) => match e {
+                nftables::helper::NftablesError::NftFailed {
+                    program: _program,
+                    hint: _hint,
+                    stdout: _stdout,
+                    stderr: _stderr,
```
Luap99: only unused variables should start with an underscore, but really no need to have this var at all then.
i.e. this should be written as
```
              nftables::helper::NftablesError::NftFailed {
                    program,
                    hint,
                    stdout: _,
                    stderr,
                } => write!(
                    f,
                    "nftables error: {program:?} did not return successfully while {hint} {stderr}",
                ),
```

Luap99 (changes requested):
Thanks, overall I think that is good if stderr contains useful info. Can you provide an example error? Ideally we have a way to check for that error in our tests, i.e. add a new test case in test/250-bridge-nftables.bats.

Also please include `Fixes: #1392` as part of the commit message body so the local git history has it as well.

davchos:
> Thanks, overall I think that is good if stderr contains useful info. Can you provide an example error? Ideally we have a way to check for that error in our tests, i.e. add a new test case in test/250-bridge-nftables.bats.
> 
> Also please include `Fixes: #1392` as part of the commit message body so the local git history has it as well.

@Luap99  Thanks for the feedback, I tested with nftables compiled without  json support. I get this error message:

```shell
  ~ sudo podman run  --replace --name podman-nginx -p 8083:80 -d docker.io/library/httpd
podman-nginx
Error: netavark: nftables error: "nft" did not return successfully while getting the current ruleset Error: JSON support not compiled-in
```

I will add a test case in test/250-bridge-nftables.bats (not for nftables json support I think).

src/error/mod.rs
```diff
@@ -159,7 +159,18 @@ impl fmt::Display for NetavarkError {
...
+                } => write!(
+                    f,
+                    "nftables error: {_program:?} did not return successfully while {_hint} {_stderr}",
```
Luap99:
```suggestion
                    "nftables error: {_program:?} did not return successfully while {_hint}: {_stderr}",
```

Would be good to have a clear cut off

Luap99 (commented):
> I will add a test case in test/250-bridge-nftables.bats (not for nftables json support I think).

You can use this, i.e. we but a custom nft script in $PATH that just prints a stderr message
```
@test "$fw_driver - nft error" {
    local nft="$NETAVARK_TMPDIR/nft"
    cat > "$nft" <<EOF
#!/usr/bin/env bash
echo 'nft custom error message' 1>&2
exit 1
EOF

    chmod +x $nft

    PATH="$NETAVARK_TMPDIR:$PATH" expected_rc=1 run_netavark --file ${TESTSDIR}/testfiles/simplebridge.json setup $(get_container_netns_path)
    assert_json ".error" 'netavark: nftables error: "nft" did not return successfully while getting the current ruleset: nft custom error message' "error message from nft is included"
}
```

davchos:
> > I will add a test case in test/250-bridge-nftables.bats (not for nftables json support I think).
> 
> You can use this, i.e. we but a custom nft script in $PATH that just prints a stderr message
> 
> ```
> @test "$fw_driver - nft error" {
>     local nft="$NETAVARK_TMPDIR/nft"
>     cat > "$nft" <<EOF
> #!/usr/bin/env bash
> echo 'nft custom error message' 1>&2
> exit 1
> EOF
> 
>     chmod +x $nft
> 
>     PATH="$NETAVARK_TMPDIR:$PATH" expected_rc=1 run_netavark --file ${TESTSDIR}/testfiles/simplebridge.json setup $(get_container_netns_path)
>     assert_json ".error" 'netavark: nftables error: "nft" did not return successfully while getting the current ruleset: nft custom error message' "error message from nft is included"
> }
> ```

mheon:
LGTM

src/error/mod.rs
```diff
@@ -159,7 +159,18 @@ impl fmt::Display for NetavarkError {
                     Ok(())
                 }
             }
-            NetavarkError::Nftables(e) => write!(f, "nftables error: {e}"),
+            NetavarkError::Nftables(e) => match e {
+                nftables::helper::NftablesError::NftFailed {
+                    program,
+                    hint,
+                    stdout: _stdout,
```
Luap99: this can just be `_`, no need to name it

Luap99 (changes requested):
As far as commit message goes the fixes should be in the body not title, i.e. something like

```
nft: Add more descriptive netavark error messages

Fixes: #1392

Signed-off-by: ...
```

Luap99 (approved):
LGTM

PR #1431, opened 2026-04-07
Add more Netavark create tests

ashley-cui:
Asked Claude to adapt missing tests from container-libs/common/libnetwork/netavark/create.go for more test coverage. In the process, it found two differences in validation, so now we validate subnet overlap and auto-detect ipv6 networks.

### DIFF
--- a/src/network/create_config/validation.rs
+++ b/src/network/create_config/validation.rs
@@ -119,6 +119,11 @@ pub fn validate_subnets(
 
     for subnet in subnets {
         validate_subnet(subnet, add_gateway, check_used, used_networks)?;
+
+        // Auto-detect IPv6 and set ipv6_enabled flag
+        if subnet.subnet.addr().is_ipv6() {
+            network.ipv6_enabled = true;
+        }
     }
 
     Ok(())
--- a/src/network/create_config/validation.rs
+++ b/src/network/create_config/validation.rs
@@ -1,5 +1,6 @@
 use crate::error::{NetavarkError, NetavarkResult};
 use crate::network::constants;
+use crate::network::core_utils;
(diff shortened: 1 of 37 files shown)


src/network/create_config/validation.rs
```diff
@@ -117,8 +117,37 @@ pub fn validate_subnets(
         return Ok(());
     };
 
+    // Check for duplicate and overlapping subnets within the same network
```
Luap99: we already have code for this in src/network/core_utils.rs validate_subnets() so we should just reuse it

test/700-create.bats
```diff
@@ -149,3 +156,238 @@ load helpers
...
+    run_netavark create < ${TESTSDIR}/testfiles/create/subnet-with-gateway.json
+    result="$output"
+    
+    assert_json "$result" ".subnets[0].subnet" == "10.100.0.0/24" "Subnet matches"
+    assert_json "$result" ".subnets[0].gateway" == "10.100.0.50" "Custom gateway matches"
+}
+
+@test "create - fail with gateway not in subnet" {
+    expected_rc=1 run_netavark create < ${TESTSDIR}/testfiles/create/gateway-not-in-subnet.json
+    assert_json "$output" ".error" "=~" "not in subnet" "Error message contains 'not in subnet'"
```
Luap99: I don't really have time to review all the test cases in detail but I guess I have to trust you do not add duplciates.

I do however have a comment on these error message checks, they should match the full error message or at least the important part. `not in subnet` can happen in three different error paths in this code base, it should check for what the test name asks for `gateway 10.200.0.1 not in subnet 10.100.0.0/24` to actually know we error on the right thing.

Same for many of the error messages here

Luap99 (approved):
LGTM

PR #1440, opened 2026-04-23
Improve error messages

ashley-cui:
Add more detail to errors for network create validation

### DIFF
--- a/test/700-create.bats
+++ b/test/700-create.bats
@@ -93,7 +93,7 @@ function setup() {
 
 @test "create - fail with invalid network name" {
     expected_rc=1 run_netavark create < ${TESTSDIR}/testfiles/create/invalid-name.json
-    assert_json ".error" "Invalid characters in network name: must match [a-zA-Z0-9][a-zA-Z0-9_.-]" "Error message contains 'Invalid characters in network name'"
+    assert_json ".error" "Invalid characters in network name \"badname!\": must match [a-zA-Z0-9][a-zA-Z0-9_.-]*" "Error message contains 'Invalid characters in network name'"
 }
 
 @test "create - fail with empty network ID" {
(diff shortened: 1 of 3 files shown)


ashley-cui:
Good to go, should fix the last failing test in https://github.com/containers/podman/pull/28570

@containers/netavark-maintainers PTAL

mtrmac (commented):
(I’m not a maintainer here, so I can’t approve / reject.)

Consider `{:?}`: if the value is already known to be not using the correct syntax, could it contain newlines, full sentences, or other surprising data?

Luap99:
> Consider {:?}: if the value is already known to be not using the correct syntax, could it contain newlines, full sentences, or other surprising data?

Well that is one of the nicer things about rust, String musts be valid utf8, so at least some dangerous bytes are skipped that way. Sure newlines and such can still happen so quoting them here does make sense I guess. I don't think we have paid much attention to any of this so far here.

My personal going so far is anything netavark receives is consider trusted input from a security POV. If something gives us garbage I am fine producing garbage.

test/700-create.bats
```diff
@@ -93,7 +93,7 @@ function setup() {
 
 @test "create - fail with invalid network name" {
     expected_rc=1 run_netavark create < ${TESTSDIR}/testfiles/create/invalid-name.json
-    assert_json ".error" "Invalid characters in network name: must match [a-zA-Z0-9][a-zA-Z0-9_.-]" "Error message contains 'Invalid characters in network name'"
+    assert_json ".error" "Invalid characters in network name badname!: must match [a-zA-Z0-9][a-zA-Z0-9_.-]*" "Error message contains 'Invalid characters in network name'"
```
Luap99: test need quotes around the name now

Luap99 (commented):
LGTM, minus the valid test

test/700-create.bats
ashley-cui: Fixed, tyvm!

Luap99 (approved):

PR #1442, opened 2026-04-30
Packit: Only create dist-git PRs for rawhide

lsm5:
We'll only be releasing breaking changes on rawhide so dist-git PRs should only be created for rawhide.


Luap99 (approved):
LGTM, your branch is way to outdated hence the cirrus AWS CI fails.
However since that does not matter for packit I guess I just force merge

PR #1445, opened 2026-05-08
bats: remove fd 3 work around

Luap99:
Do not close fd3 for the run call, bats needs it to report proper errors, i.e. missing command.

This is similar to the aardvark-dns fix[1] except we did not close the fd so bats does not fail just hide the errors which is not much better.

[1] https://github.com/containers/aardvark-dns/pull/686


mheon (approved):
LGTM

PR #1447, opened 2026-05-09
fix: reject auto-generated gateways outside subnets

immanuwell:
Small fix for create validation.

Before this, `/32` and `/128` subnets with no gateway slipped thru. Netavark generated `network + 1`, which is outside the subnet. Tiny footgun, tbh.

Repro:
1. keep the new regression tests, but drop the code change
2. run `PROTOC=/tmp/netavark-protoc/bin/protoc cargo test validate_subnet_rejects_ --lib`
3. both tests fail because `validate_subnet()` returns `Ok(())` for `10.100.0.1/32` and `fd00::1/128`

Now `first_ip_in_subnet()` checks the generated gateway is still inside the subnet and errors cleanly when it is not.

Tests:
- `PROTOC=/tmp/netavark-protoc/bin/protoc cargo test --lib`

### DIFF
--- a/src/network/create_config/validation.rs
+++ b/src/network/create_config/validation.rs
@@ -138,13 +138,13 @@ pub fn validate_subnets(
 
 /// Get the first IP address in a subnet (network address + 1).
 fn first_ip_in_subnet(network: &IpNet) -> NetavarkResult<IpAddr> {
-    match network {
+    let first_ip = match network {
         IpNet::V4(net_v4) => {
             let network_addr: u32 = net_v4.network().into();
             let first_ip = network_addr
                 .checked_add(1)
                 .ok_or_else(|| NetavarkError::msg("Subnet address overflow"))?;
-            Ok(IpAddr::V4(first_ip.into()))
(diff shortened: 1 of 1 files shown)


src/network/create_config/validation.rs
```diff
@@ -154,9 +154,18 @@ fn first_ip_in_subnet(network: &IpNet) -> NetavarkResult<IpAddr> {
...
+            IpAddr::V6(Ipv6Addr::from(first_ip_u128.to_be_bytes()))
         }
+    };
+
+    if !network.contains(&first_ip) {
+        return Err(NetavarkError::msg(format!(
+            "could not create gateway for subnet {}",
+            network
+        )));
     }
```
Luap99, line 166: instead of adding another check here could you not just swap the order in  validate_subnet() which already checks that as well.

Then again if we do not work with /32 or /128 then we might as well produce an actual error saying that explicitly at least.

mheon:
LGTM

Luap99:
this was done in https://github.com/containers/netavark/pull/1462

PR #1452, opened 2026-05-19
teardown: clean up host resources when netns is gone

### LINKED ISSUE #1451: teardown: host-side cleanup skipped if netns is missing
When the network namespace path is no longer valid at teardown time,  open_netlink_sockets
returns early. This means host-side cleanup (firewall rules, IPAM, DNS entries) — which  doesn't
actually need the netns — is also skipped.

This  would affect eg when the service using netavark (eg podman) is running under systemd with private mount namespaces (e.g. ProtectSystem=full, PrivateDevices=yes). When the service restarts, its mount
namespace would get destroyed along with any netns bind mounts it had created.

stuart-mclaren-hpe:
Change the NetworkDriver::teardown() signature to accept
an Option for the container netns netlink socket. When the
container network namespace is no longer available (e.g.
the process exited), drivers can still clean up host-side
resources such as firewall rules and bridge interfaces.

In teardown.rs, failure now falls back to opening only
the host netlink socket and passes None for the container
socket. Each driver handles the None case:

- bridge: skips veth/route deletion, still removes
  firewall rules, sysctl files, and empty bridges
- vlan/macvlan: returns Ok(()) since all state is
  netns-side

Note: netns error handling: for "does not exist" no error
is raised, other (potentially unrecoverable) errors
are raised.

Note: plugin is still invoked when netns does not exist.

Note: there may be a general principle here:
teardown should continue rather than error if something
it should remove is already gone.

Fixes: #1451

Signed-off-by: Stuart McLaren <stuart.mclaren@hpe.com>

### DIFF
--- a/src/commands/teardown.rs
+++ b/src/commands/teardown.rs
@@ -92,8 +92,24 @@ impl Teardown {
 
         let firewall_driver = firewall::get_supported_firewall_driver(firewall_driver)?;
 
-        let (mut hostns, mut netns) =
-            core_utils::open_netlink_sockets(&self.network_namespace_path)?;
+        let mut hostns = core_utils::open_host_netlink_socket()?;
+        let (netns_file, mut netns_netlink) =
+            match core_utils::open_netns_netlink_socket(&self.network_namespace_path, &hostns) {
+                Ok(netns) => (Some(netns.file), Some(netns.netlink)),
+                Err(e) => {
+                    match e.unwrap() {
+                        NetavarkError::Io(io_err)
+                            if io_err.kind() == std::io::ErrorKind::NotFound =>
+                        {
+                            log::warn!("container netns not found: {e}");
+                        }
+                        _ => {
+                            error_list.push(e);
+                        }
+                    }
+                    (None, None)
+                }
+            };
 
         for named_network_opts in &network_options.networks {
             let per_network_opts = &named_network_opts.opts;
@@ -116,7 +132,7 @@ impl Teardown {
                     container_hostname: &network_options.container_hostname,
                     container_dns_servers: &network_options.dns_servers,
                     netns_host: hostns.file.as_fd(),
-                    netns_container: netns.file.as_fd(),
+                    netns_container: netns_file.as_ref().map(|f| f.as_fd()), // None when no netns
                     netns_path: &self.network_namespace_path,
                     network,
                     per_network_opts,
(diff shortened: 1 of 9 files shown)


src/commands/teardown.rs
```diff
@@ -92,8 +92,15 @@ impl Teardown {
...
         let firewall_driver = firewall::get_supported_firewall_driver(firewall_driver)?;
 
-        let (mut hostns, mut netns) =
-            core_utils::open_netlink_sockets(&self.network_namespace_path)?;
+        let mut hostns = core_utils::open_host_netlink_socket()?;
+        let (netns_file, mut netns_netlink) =
+            match core_utils::open_netns_netlink_socket(&self.network_namespace_path, &hostns) {
+                Ok(netns) => (Some(netns.file), Some(netns.netlink)),
+                Err(e) => {
+                    debug!("container netns unavailable: {e}");
```
Luap99: This needs to raise a an error, please push an error to `error_list`, i.e. see a few llines above for an example.

src/network/bridge.rs
```diff
@@ -214,7 +214,9 @@ impl driver::NetworkDriver for Bridge<'_> {
             self.info.network.internal,
             self.info.rootless,
             self.info.netns_host,
-            self.info.netns_container,
+            self.info
+                .netns_container
+                .expect("netns_container required for setup"),
```
Luap99: expect panics and causes nv to not return a readable error to podman as we expect proper json, so end users will never see this.

As such this should return a error instead, in general I am somewhat unhappy with this being a option in the setup path but I guess there is no simple way to avoid thgat

src/network/plugin.rs
```diff
@@ -118,8 +118,11 @@ impl NetworkDriver for PluginDriver<'_> {
 
     fn teardown(
         &self,
-        _netlink_sockets: (&mut Socket<NetlinkRoute>, &mut Socket<NetlinkRoute>),
+        netlink_sockets: (&mut Socket<NetlinkRoute>, Option<&mut Socket<NetlinkRoute>>),
     ) -> NetavarkResult<()> {
+        if netlink_sockets.1.is_none() {
+            return Ok(());
+        }
```
Luap99: that leaves the plugin author no chance to cleanup so this seems wrong, I rather pass the non existing path to the plugin and let them deal with it like we do

src/commands/teardown.rs
stuart-mclaren-hpe: I did this, but I excluded the "not found" case, so that rc=0 unless
the error is something that is more likely to require an action.

Let me know if you think that makes sense.

src/network/bridge.rs
stuart-mclaren-hpe: Changed to error.

There is an equivalent expect in vlan.rs - should that become an error too?

mheon (approved):
LGTM

stuart-mclaren-hpe:
Thanks for your reviews.

Let me know if you need any further updates.

Luap99 (approved):
LGTM, sorry for the delay

PR #1455, opened 2026-05-29
Added a "set_link_ns_pid" in netlink_route.rs

tomis007:
Added a "set_link_ns_pid" function in netlink_route to expose the corresponding way to change the network namespace by a pid. The file descriptor option already exists in the code. I'm working on implementing a network plugin and noticed that this option isn't currently exposed through the netavark netlink internals. The change is quite small and straightforward and open to whatever changes you think are best, thanks !

### DIFF
--- a/src/network/netlink_route.rs
+++ b/src/network/netlink_route.rs
@@ -151,6 +151,16 @@ impl Socket<NetlinkRoute> {
         Ok(())
     }
 
+    pub fn set_link_ns_pid(&mut self, link_id: u32, pid: u32) -> NetavarkResult<()> {
+        let mut msg = LinkMessage::default();
+        msg.header.index = link_id;
+        msg.attributes.push(LinkAttribute::NetNsPid(pid));
+
+        let result = self.make_netlink_request(RouteNetlinkMessage::SetLink(msg), NLM_F_ACK)?;
+        expect_netlink_result!(result, 0);
+        Ok(())
+    }
+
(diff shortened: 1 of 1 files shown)


tomis007:
Looks like a few of the auto rpm jobs failed but I can't figure out how to see the logs to determine why, happy to make whatever changes are needed to get the builds working

Luap99:
/packit retest-failed

Looks like logs are just missing, likely some infra failures.

Mind sharing what you type of plugin you are working on? Just interest to know what is out there.

Luap99 (approved):
I don't mind a small addition like this but I like to highlight that I do not consider this a general purpose library so if large changes are needed that we do not use in netavark I likely reject them.

Technically this is not need because you could open /proc/$PID/ns/net and then use set_link_ns on the fd though I get if you have the pid you like zol directly call this and not have an extra open().

LGTM

tomis007:
> Technically this is not need because you could open /proc/$PID/ns/net and then use set_link_ns on the fd though I get if you have the pid you like zol directly call this and not have an extra open().

Yeah exactly

> Mind sharing what you type of plugin you are working on? Just interest to know what is out there.

Definitely ! I'm implementing another wireguard plugin that's similar in use to https://github.com/b-m-f/netavark-wireguard-plugin, but changing up how some of the internals work to better fit my use cases. I can share once I've gotten it finished if you're interested

PR #1460, opened 2026-06-08
update sha2 to 0.11

Luap99:
The update removed the default UpperHex implementation from the returned
type so we now need to explicitly import another crate to give a type
that implements the conversion.

base16ct seems to be recommended by the RustCrypto maintainers, see
https://github.com/RustCrypto/hashes/issues/185#issuecomment-4313220182

Also while at it rework the code a bit to use Sha512::digest() directly
and truncate the string in place with truncate() do not need a second
copy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -269,6 +269,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "base16ct"
+version = "1.0.0"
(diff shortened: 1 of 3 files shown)


Luap99:
@mheon PTAL

mheon (approved):
LGTM

PR #1462, opened 2026-06-09
prevent incorrect gateway assignment

Luap99:
In small subnets we cannot assign a gateway as it would not be valid address within the network.

Check these cases and return a sensible error.


Luap99:
@mheon PTAL

replaces https://github.com/containers/netavark/pull/1447

mheon:
LGTM

mheon (approved):

PR #1464, opened 2026-06-09
Release v2.0.0

Luap99:


Luap99:
So apparently we cannot test new releases by default because the aardvark-dns rpm in podman0next is still at 2.0 and thus fails the 2.1 spec requirement due the -dev bump.

On the other hand we can tell packit to test against a specific PR 


/packit test containers/aardvark-dns#706

Luap99:
Ok the direct link to the aardvark-dns does not seem to pick up the right rpm either. I cannot make sense of the packit/tmt logs here either. I guess we need to ask @lsm5 when he is back.

For now I suggest we force merge this, there are no real code changes here so tests should pass just like on main.

Luap99:
I forced push without the dev bump so we can reuse the rpms build here for machine os if needed

baude:
this LGTM; I like the way you did the release notes ... 2.0!

baude (approved):

mheon:
LGTM

PR #1467, opened 2026-06-17
rpm: drop unused protobuf-c dependency

yselkowitz:
protobuf-c is just a library; the intention was probably the protoc binary,
which is packaged as protobuf-compiler.


Luap99 (approved):
LGTM
@lsm5 PTAL

lsm5 (approved):
LGTM. Thanks @yselkowitz

PR #1480, opened 2026-07-23
Add IPv6 subnet pool allocation for network creation

thrawn-sh:
Add support for sequential IPv6 subnet allocation from configurable subnet pools, complementing the existing IPv4 pool allocation and random IPv6 subnet generation (RFC 4193).

When container-libs passes a `subnet_pools_v6` field in the create config, netavark now allocates IPv6 subnets sequentially from those pools — mirroring how `subnet_pools` works for IPv4. When the field is absent or empty, the existing random fd00::/8 allocation is used.

Key changes:

- types.rs: Add `subnet_pools_v6` field to `CreateOpts` with `#[serde(default)]` for backward compatibility with older container-libs versions that don't send this field.

- subnet.rs: Add `next_subnet_v6()` for incrementing IPv6 network addresses using u128 arithmetic, and `get_free_ipv6_network_subnet_from_pools()` for sequential allocation with a 10,000 iteration cap to prevent runaway scans in large address spaces (e.g. fd00::/8 with /64 subnets has 2^56 candidates vs 256 for typical IPv4 pools).

- driver.rs: Dispatch to pool-based allocation when `subnet_pools_v6` is non-empty, falling back to random generation otherwise.

This is the netavark counterpart to the container-libs configuration wiring in podman-container-tools/container-libs#995.

### DIFF
--- a/docs/netavark-create-input.md
+++ b/docs/netavark-create-input.md
@@ -43,7 +43,8 @@ The `options` object contains creation options and settings.
 
 | Field | Description | Required |
 |-------|-------------|----------|
-| `subnet_pools` | Array of subnet pools from which to allocate subnets. Each pool contains a `base` (CIDR) and `size` (subnet size in bits). | Yes |
+| `subnet_pools` | Array of IPv4 subnet pools from which to allocate subnets. Each pool contains a `base` (CIDR) and `size` (subnet size in bits). | Yes |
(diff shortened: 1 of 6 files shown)


TomSweeneyRedHat:
@thrawn-sh our CI allows only signed commits in.  Please `git commit --amend -s`, fetch, rebase and push please.

PR #1498, opened 2026-08-09
Don't leak sysctl.d drop-ins for internal networks

### LINKED ISSUE #1497: per-bridge sysctl.d drop-ins leaked for internal networks, disabling forwarding on unrelated external networks
I'm creating two networks in a podman compose, one internal (`internal: true`), one with default external config. I run this in a podman machine on a MacOS host. The first time around, this would always work fine. But when I repeatedly tear down the compose and re-up it, sometimes the external network would not forward my packages externally, other times it would work just fine.

## Deterministically reproducing the issue

This was all very non-deterministic, but I figured out a way to trigger the bug deterministically.

In my podman machine running under MacOS (see "My environment" below; should work on any linux hosts with systemd-sysctl and udev as well):
```console
# podman network create --internal a
a
# BR=$(podman network inspect a --format '{{.NetworkInterface}}')
# podman run -d --name c1 --network a docker.io/library/alpine sleep 60
30900c35fb544e101a9bf1eb9d5ec911d8b3b5265eeda945339c2fc994d80793

# cat /run/sysctl.d/10-netavark-$BR.conf
# autogenerated by netavark
net/ipv4/conf/podman3/forwarding = 0
net/ipv4/conf/podman3/rp_filter = 2

# podman rm -f -t0 c1 && podman network rm a
c1
a

# ls -l /run/sysctl.d/10-netavark-$BR.conf           # <-- this is left lying around from the deleted network
-rw-r--r--. 1 root root 101 Aug  9 12:41 /run/sysctl.d/10-netavark-podman3.conf

# podman network create b                            # reuses the bridge name
b
# podman run -d --name c2 --network b docker.io/library/alpine sleep 60
e365680526e5a84ac6aa30dd543e4f2937af5d07d8ad35d1033b9d20ddb19c5a

# cat /proc/sys/net/ipv4/conf/$BR/forwarding         # <-- external network, forwarding off
0
# podman exec c2 wget -qO- -T5 https://example.com   # <-- external connectivity broken
wget: download timed out
```

## What seems to be going on

1. Setting up an internal network, netavark writes `/run/sysctl.d/10-netavark-<bridge>.conf`
   containing `net/ipv4/conf/<bridge>/forwarding = 0` (and `rp_filter = 2`).
2. Tearing the network down removes the bridge but leaves the drop-in in `/run/sysctl.d/`.
3. Bridge names are reused.
4. systemd reapplies the drop-in as per some udev-rule to every newly added interface of the same name
5. netavark's own configuration of the new bridge and the udev-triggered `systemd-sysctl` run are
   concurrent and race. When `systemd-sysctl` wins the race `forwarding = 0` stays set.

## Expected behaviour

An external network's bridge must have `net.ipv4.conf.<bridge>.forwarding = 1`, regardless of what
any previously-deleted network with the same bridge name required.

## My environment

| | |
|---|---|
| Host | macOS (Apple Silicon, arm64), `podman machine` |
| VM | Fedora CoreOS 44.20260707.3.1 |
| Kernel | 7.1.3-200.fc44.aarch64 |
| podman | 6.0.2 (`b28edb9ad70ce4317dc762ee9ce0a6d081d154e9`, linux/arm64) |
| netavark | 2.0.0 |
| systemd | 259 (259.7-1.fc44) |

I doubt this is MacOS-specific; the whole topology is internal to the Linux VM in podman machine. Any rootful podman host with `systemd-sysctl` and udev should reproduce it.

## Workaround

Manually running

```sh
sysctl -w net.ipv4.conf.<bridge>.forwarding=1
```

fixes the issue for the currently configured network. Not a permanent fix, tearing down and re-upping it causes the bug to trigger again (sometimes; subject to race condition).

## AI use

Full disclosure: I had Claude Opus 5 help me debug this. Without it, this bug report would have ended after the first paragraph. I did write this entire report myself and I did manually verify and run everything described here, so I'm sure all of this is legit. Thanks to Opus, I also managed to build a fix for this, which I will submit a PR for shortly, when I'm done reviewing it.

hcebke:
Fixes: #1497

These changed were generated by Claude Opus and then reviewed, edited, and tested by me.


test/250-bridge-nftables.bats
```diff
@@ -32,6 +32,43 @@ export NETAVARK_FW=nftables
...
+net/ipv4/conf/podman0/forwarding = 0
+net/ipv4/conf/podman0/rp_filter = 2"
+   run_in_host_netns cat /run/sysctl.d/10-netavark-podman0.conf
+   assert "$output" == "$sysctl_content" "sysctl.d file content"
+
+   run_netavark --file ${TESTSDIR}/testfiles/internal.json teardown $(get_container_netns_path)
+
+   # The file must not be leaked. Bridge names are reused so systemd-sysctl would
+   # apply forwarding = 0 to the next, possibly not internal, network with that name.
+   expected_rc=1 run_in_host_netns cat /run/sysctl.d/10-netavark-podman0.conf
```
Luap99, line 42: this should just be added to an existing internal test

Luap99 (changes requested):
This is mixing way to many things at once, this needs to be split in proper commits with proper reasons why they are needed. Staring with the bug at hand of the wrong internal condition.

I am also not convinced we need the "atomic" write behavior, our writes are small enough. And enabling forwarding on that one bridge should not be needed as we already enable the default forwarding.

hcebke:
Fair point! I've split it up into multiple commits. I still think the atomic write is necessary (see my reasoning in the commit message). I removed emitting the forwarding = 1 setting for non-internal networks altogether, since it shouldn't be necessary as long as everything else operates as it should. (It would be kind of nice from a symmetry perspective, though, so maybe consider it. ;-))

Unit tests are green, Hoping the github tests will be too once they have finished running. 🤞

hcebke:
I did fat-finger the issue # in one commit message, though. :-(
Will force-push a revised commit message.

test/250-bridge-nftables.bats
hcebke: Done

src/network/bridge.rs
```diff
@@ -484,16 +484,19 @@ impl<'a> Bridge<'a> {
         };
 
         // 2. Tear down firewall & sysctl FIRST (while routes still exist)
-        if !self.info.network.internal && mode == BridgeMode::Managed {
-            match self.teardown_firewall(complete_teardown, bridge_name.clone()) {
-                Ok(_) => {}
-                Err(err) => {
-                    error_list.push(err);
+        if mode == BridgeMode::Managed {
```
mheon, line 488: Is this functionally different from what's already there? It does swap the order of the comparisons, but it could easily still do the two checks on a single line?

Luap99: yes see the second complete_teardown branch inside which is needed to remove the sysctl file

Luap99 (commented):
This needs a rebase, logic wise seem fine now for me

src/network/bridge.rs
hcebke: To elaborate on Paul's point: the `!self.info.network.internal` condition used to guard the entire block, now it's only guarding the `teardown_firewall` part. The `sysctl.d` drop-in deletion now happens regardless of the network being `internal` or not.

hcebke:
Rebased. Checks currently in progress.

mheon:
LGTM

hcebke:
I rebased once more. If you don't have any more change requests, would you mind merging this?

mheon:
@Luap99 PTAL

hcebke:
Rebased once again. I know you guys are busy, so if there is anything I can do to help you with this PR, let me know!

PR #1500, opened 2026-08-10
bridge: add DNS firewall rules for internal networks

### LINKED ISSUE #1051: No DNS in internal networks when dns_port is not set to 53
## Observed behavior

When setting `dns_bind_port` in `/etc/containers/containers.conf` to anything other than 53, containers in internal networks have no container-to-container name resolution (e.g. you can't `ping database` from a `web` container).

Inside containers `/etc/resolv.conf` is configured properly to the IP of the aardvark-dns resolver, but cannot specify a non-default port: DNS requests try to go to port 53 and fail. The non-default port is however perfectly reachable (using e.g. `dig` or `nslookup`).

## Expected behavior

The behavior should be the same as with the default DNS port.

In non-internal networks, the following iptables rules are added to remap the non-default DNS port as 53 (YMMV), but these are missing in internal networks:

```
iptables -I INPUT 1 -s 10.89.0.0/24 -d 10.89.0.0/24 -p udp --dport 
20053 -j ACCEPT
iptables -t nat -I PREROUTING 1 -m addrtype --dst-type LOCAL -p udp 
-d 10.89.0.0/24 --dport 53 -j DNAT --to-destination 10.89.0.1:20053
iptables -t nat -I OUTPUT 1 -m addrtype --dst-type LOCAL -p udp -d 
10.89.0.0/24 --dport 53 -j DNAT --to-destination 10.89.0.1:20053
```

----

Versions on the host:

* OS: Archlinux (Linux 6.10.3-arch1-2 at the moment)
* `podman`: 5.1.2
* `aardvark-dns`: 1.11.0
* Extra software: `ufw` firewall with default deny for incoming and forwarded packets

Note: See [discussion on the mailing list](https://lists.podman.io/archives/list/podman@lists.podman.io/message/OR7CY2AKD4VAY3S67EHLJJI62KQHROZW/) for more details.

### LINKED ISSUE #1055: Cannot resolve container hostnames in internal network
### Issue Description

The DNS resolution in an internal network doesn't work. The DNS server is unreachable.

Steps to reproduce the issue. Run all commands as root:

1.  Create internal network:
```
# podman network create internal-network --internal
```
2. Create first container:
```
# podman run -d --rm --name container1 --network internal-network fedora sleep infinity
```
3. In a second container, try to resolve the first container hostname:
```
# podman run -ti --rm --network internal-network quay.io/jorti/debug
[root@5823c479811c /]# cat /etc/resolv.conf 
search dns.podman
nameserver 10.89.2.1
[root@5823c479811c /]# dig container1
;; communications error to 10.89.2.1#53: host unreachable
;; communications error to 10.89.2.1#53: host unreachable
;; communications error to 10.89.2.1#53: host unreachable

; <<>> DiG 9.18.26 <<>> container1
;; global options: +cmd
;; no servers could be reached
```

Internal DNS server is unreachable. I cannot resolve other container hostnames in an internal network.

This network uses the netavark backend and the bridge driver, so I expect a working DNS functionality.

### podman info output

```yaml
host:
  arch: amd64
  buildahVersion: 1.35.4
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - rdma
  - misc
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.10-1.fc40.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.10, commit: '
  cpuUtilization:
    idlePercent: 86.94
    systemPercent: 2.59
    userPercent: 10.47
  cpus: 8
  databaseBackend: boltdb
  distribution:
    distribution: fedora
    variant: workstation
    version: "40"
  eventLogger: journald
  freeLocks: 2045
  hostname: _REDACTED_
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 6.8.10-300.fc40.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 6752358400
  memTotal: 33483927552
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    dns:
      package: aardvark-dns-1.10.0-1.fc40.x86_64
      path: /usr/libexec/podman/aardvark-dns
      version: aardvark-dns 1.10.0
    package: netavark-1.10.3-3.fc40.x86_64
    path: /usr/libexec/podman/netavark
    version: netavark 1.10.3
  ociRuntime:
    name: crun
    package: crun-1.15-1.fc40.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.15
      commit: e6eacaf4034e84185fd8780ac9262bbf57082278
      rundir: /run/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL
  os: linux
  pasta:
    executable: /usr/bin/pasta
    package: passt-0^20240510.g7288448-1.fc40.x86_64
    version: |
      pasta 0^20240510.g7288448-1.fc40.x86_64
      Copyright Red Hat
      GNU General Public License, version 2 or later
        <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>
      This is free software: you are free to change and redistribute it.
      There is NO WARRANTY, to the extent permitted by law.
  remoteSocket:
    exists: false
    path: /run/podman/podman.sock
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: false
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: true
  serviceIsRemote: false
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.2.2-2.fc40.x86_64
    version: |-
      slirp4netns version 1.2.2
      commit: 0ee2d87523e906518d34a6b423271e4826f71faf
      libslirp: 4.7.0
      SLIRP_CONFIG_VERSION_MAX: 4
      libseccomp: 2.5.3
  swapFree: 8589930496
  swapTotal: 8589930496
  uptime: 1h 39m 8.00s (Approximately 0.04 days)
  variant: ""
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  - ipvlan
  volume:
  - local
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - docker.io
  - quay.io
store:
  configFile: /usr/share/containers/storage.conf
  containerStore:
    number: 3
    paused: 0
    running: 1
    stopped: 2
  graphDriverName: overlay
  graphOptions:
    overlay.imagestore: /usr/lib/containers/storage
    overlay.mountopt: nodev,metacopy=on
  graphRoot: /var/lib/containers/storage
  graphRootAllocated: 510389125120
  graphRootUsed: 474386456576
  graphStatus:
    Backing Filesystem: btrfs
    Native Overlay Diff: "false"
    Supports d_type: "true"
    Supports shifting: "true"
    Supports volatile: "true"
    Using metacopy: "true"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 9
  runRoot: /run/containers/storage
  transientStore: false
  volumePath: /var/lib/containers/storage/volumes
version:
  APIVersion: 5.0.3
  Built: 1715299200
  BuiltTime: Fri May 10 02:00:00 2024
  GitCommit: ""
  GoVersion: go1.22.2
  Os: linux
  OsArch: linux/amd64
  Version: 5.0.3
```

### Podman in a container

No

### Privileged Or Rootless

Privileged

### Upstream Latest Release

Yes

_No response_

### Additional information

_No response_

ajoshua2004:
Internal networks skipped firewall setup entirely, so the DNS input-accept rule was never created: container `hostname` resolution didn't work in internal networks. The DNS redirect (dport 53 -> dns_port) was never created, so a non-default `dns-port` didn't work. 

For nftables, internal networks will get the DNS accept and redirect redirect rules, but will still skip all 
masquerade/forward/port forwarding rules to keep the network isolation. 

For firewalld, internal networks will skip setup/tear down. 

### Tests
`sudo bats -f "internal network dns with default drop policy" test/250-bridge-nftables.bats`

Fixes: https://github.com/containers/netavark/issues/1051
Fixes: https://github.com/containers/netavark/issues/1055

### DIFF
--- a/src/network/bridge.rs
+++ b/src/network/bridge.rs
@@ -362,8 +362,9 @@ impl driver::NetworkDriver for Bridge<'_> {
         };
 
         if let BridgeMode::Managed = data.mode {
-            // if the network is internal do not setup firewall rules
-            if !self.info.network.internal {
+            // Internal networks normally skip firewall setup, but when DNS is
+            // enabled, we still need the DNS rules so aardvark-dns is reachable.
+            if !self.info.network.internal || self.info.network.dns_enabled {
                 self.setup_firewall(data)?
             }
         }
@@ -484,7 +485,9 @@ impl<'a> Bridge<'a> {
         };
 
         // 2. Tear down firewall & sysctl FIRST (while routes still exist)
-        if !self.info.network.internal && mode == BridgeMode::Managed {
+        if (!self.info.network.internal || self.info.network.dns_enabled)
+            && mode == BridgeMode::Managed
+        {
             match self.teardown_firewall(complete_teardown, bridge_name.clone()) {
                 Ok(_) => {}
                 Err(err) => {
@@ -543,6 +546,7 @@ impl<'a> Bridge<'a> {
             network_hash_name: id_network_hash.clone(),
             isolation: isolate,
             dns_port: self.info.dns_port,
+            internal: self.info.network.internal,
             outbound_addr4,
             outbound_addr6,
         };
@@ -586,6 +590,7 @@ impl<'a> Bridge<'a> {
             subnet_v6: net_v6,
             dns_port: self.info.dns_port,
             dns_server_ips: nameservers,
+            internal: self.info.network.internal,
         };
         Ok((sn, spf))
     }
--- a/test/250-bridge-nftables.bats
+++ b/test/250-bridge-nftables.bats
@@ -1426,3 +1426,115 @@ EOF
     run_helper ls "$NETAVARK_TMPDIR/config/aardvark-dns"
     assert "$output" == "" "no aardvark entries written on rejection"
 }
+
+# https://github.com/containers/netavark/issues/1055
+@test "$fw_driver - internal network dns with default drop policy" {
+    run_netavark --file ${TESTSDIR}/testfiles/internal-dns.json setup $(get_container_netns_path)
+
+    # internal networks get ONLY the DNS input accept rules, no routing/NAT rules
+
+    # INPUT: base chain + one DNS accept per subnet (v4, v6), nothing else
+    run_in_host_netns nft list chain inet netavark INPUT
+    assert "${lines[3]}" =~ "ip saddr 10.89.3.0/24 meta l4proto \{ tcp, udp \} th dport 53 accept" "ipv4 DNS accept rule"
+    assert "${lines[4]}" =~ "ip6 saddr fd10:88:a::/64 meta l4proto \{ tcp, udp \} th dport 53 accept" "ipv6 DNS accept rule"
+    assert "${#lines[@]}" = 7 "only DNS accept rules in INPUT chain"
+
+    # per-subnet chains must be empty: no daddr accept, no masquerade/SNAT
+    run_in_host_netns nft list chain inet netavark nv_ec79dd0c_10_89_3_0_nm24
+    assert "${#lines[@]}" = 4 "ipv4 subnet chain is empty for internal network"
+    run_in_host_netns nft list chain inet netavark nv_ec79dd0c_fd10-88-a--_nm64
+    assert "${#lines[@]}" = 4 "ipv6 subnet chain is empty for internal network"
+
+    # FORWARD: only base rules, no per-network accept rules
+    run_in_host_netns nft list chain inet netavark FORWARD
+    assert "${lines[3]}" =~ "ct state invalid drop" "CT state invalid rule"
+    assert "${lines[4]}" =~ "ct mark & 0x00001000 == 0x00001000 accept" "accept dnat traffic"
+    assert "${lines[5]}" =~ "jump NETAVARK-ISOLATION-1" "jump to isolation chain"
+    assert "${#lines[@]}" = 8 "no per-network FORWARD rules for internal network"
+
+    # POSTROUTING: only the base mark-masquerade rule, no per-network jump
+    run_in_host_netns nft list chain inet netavark POSTROUTING
+    assert "${lines[3]}" =~ "meta mark & 0x00002000 == 0x00002000 masquerade" "Mark-masquerade rule"
+    assert "${#lines[@]}" = 6 "no per-network POSTROUTING rule for internal network"
+
+    # a port is published, but internal networks must skip container port-forwarding
+    # (and there is no DNS redirect on the default port), so the chain stays empty
+    run_in_host_netns nft list chain inet netavark NETAVARK-HOSTPORT-DNAT
+    assert "${#lines[@]}" = 4 "no port-forwarding rules for internal network"
+
+    # internal isolation is still enforced: no default route
+    run_in_container_netns ip route show
+    assert "$output" "!~" "default" "No default route for internal networks"
+    run_in_container_netns ip -6 route show
+    assert "$output" "!~" "default" "No default route for internal networks (ipv6)"
+
+    # enforce a default drop policy and make sure DNS still resolves
+    run_in_host_netns nft add chain inet netavark INPUT \{ type filter hook input priority 0 \; policy drop \; \}
+    run_in_host_netns nft add rule inet netavark INPUT ct state related,established accept
+    run_in_host_netns nft add rule inet netavark INPUT meta l4proto ipv6-icmp accept
+
+    run_in_container_netns dig +short "somename.dns.podman" @10.89.3.1 A "somename.dns.podman" @10.89.3.1 AAAA
+    assert "${lines[0]}" =~ "10.89.3.2" "ipv4 dns resolution works 1/2"
+    assert "${lines[1]}" =~ "fd10:88:a::2" "ipv6 dns resolution works 2/2"
+
+    run_in_container_netns dig +short "somename.dns.podman" @fd10:88:a::1
+    assert "${lines[0]}" =~ "10.89.3.2" "ipv6 dns resolution works"
+
+    run_netavark --file ${TESTSDIR}/testfiles/internal-dns.json teardown $(get_container_netns_path)
+}
+
+# https://github.com/containers/netavark/issues/1051
+@test "$fw_driver - internal network dns with default drop policy and non-default dns port" {
+    dns_port=$((RANDOM+10000))
+
+    NETAVARK_DNS_PORT="$dns_port" run_netavark --file ${TESTSDIR}/testfiles/internal-dns.json setup $(get_container_netns_path)
+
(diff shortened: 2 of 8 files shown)


test/250-bridge-nftables.bats
```diff
@@ -1426,3 +1426,70 @@ EOF
...
+    run_netavark --file ${TESTSDIR}/testfiles/internal-dns.json setup $(get_container_netns_path)
+
+    # internal networks must get ONLY the DNS input accept, no routing/NAT rules
+    run_in_host_netns nft list table inet netavark
+    assert "$output" "=~" "ip saddr 10.89.3.0/24 meta l4proto \{ tcp, udp \} th dport 53 accept" "ipv4 DNS accept rule present"
+    assert "$output" "=~" "ip6 saddr fd10:88:a::/64 meta l4proto \{ tcp, udp \} th dport 53 accept" "ipv6 DNS accept rule present"
+    assert "$output" "!~" "224.0.0.0/4" "no ipv4 masquerade rule for internal network"
+    assert "$output" "!~" "ff00::/8" "no ipv6 masquerade rule for internal network"
+    assert "$output" "!~" "ip saddr 10.89.3.0/24 accept" "no forward saddr accept for internal network"
+    assert "$output" "!~" "ip daddr 10.89.3.0/24 ct state established,related accept" "no forward daddr accept for internal network"
```
Luap99: the rule output is deterministic enough, I would check the correct line output here and then check against the total line count like we have in other tests. The `!~` are fragile as it does not prevent someone from adding/chaining other rules. So having a hard count on what rules we expect is best

Luap99 (commented):
I think the setup_port_forward also needs a condition to skip the port forwarding rules for internal

Luap99:
cc @mheon

Luap99:
And I suppose this must be wired into the firewalld backend as well, because right now we would create all normal rules there

ajoshua2004:
> And I suppose this must be wired into the firewalld backend as well, because right now we would create all normal rules there

I'm thinking a separate internal zone that only allows DNS to the host could be an option?

mheon:
Didn't we turn this off due to a CVE, data exfiltration from internal networks via DNS request? Probably still safe since we turned that off in Aardvark as well...

Luap99:
> Didn't we turn this off due to a CVE, data exfiltration from internal networks via DNS request? Probably still safe since we turned that off in Aardvark as well...

You made aarvark-dns block external dns requests yes. but that unreleated to this one here. internal dns already works as long as no custom cnd port is given and the default firewall rules accept the traffic. So thread model wise this here changes nothing it just adds the missing rules to make internal dns always work correctly.

Luap99:
> > And I suppose this must be wired into the firewalld backend as well, because right now we would create all normal rules there
> 
> I'm thinking a separate internal zone that only allows DNS to the host could be an option?

I have never touched the firewalld code really, I am not happy with having these extra implementations. I think @mheon suggested he would be fine dropping it at some point but I doubt we can just do that as likely some user rely on it.

I think for this here alone the simple thing would be ensure that if internal do not alter any firewalld rules to keep the status quo. With the current implementation since you call into the firewall code and there is no internal condition for firewalld it means all the regular rules will be created which is definitely wrong.

ajoshua2004:
> > > And I suppose this must be wired into the firewalld backend as well, because right now we would create all normal rules there
> > 
> > 
> > I'm thinking a separate internal zone that only allows DNS to the host could be an option?
> 
> I have never touched the firewalld code really, I am not happy with having these extra implementations. I think @mheon suggested he would be fine dropping it at some point but I doubt we can just do that as likely some user rely on it.
> 
> I think for this here alone the simple thing would be ensure that if internal do not alter any firewalld rules to keep the status quo. With the current implementation since you call into the firewall code and there is no internal condition for firewalld it means all the regular rules will be created which is definitely wrong.

What i just did was just skip the firewalld setup and tear down so internal networks shouldnt get the regular rules, added an extra test too

src/network/bridge.rs
```diff
@@ -362,8 +362,9 @@ impl driver::NetworkDriver for Bridge<'_> {
         };
 
         if let BridgeMode::Managed = data.mode {
-            // if the network is internal do not setup firewall rules
-            if !self.info.network.internal {
+            // Internal networks normally skip firewall setup, but when DNS is
+            // enabled, we still need the DNS rules so aardvark-dns is reachable.
+            if !self.info.network.internal || self.info.network.dns_enabled {
                 self.setup_firewall(data)?
```
mheon, line 368: I think that's still going to configure full forwarding, not just DNS

mheon: (At least for firewalld)

ajoshua2004: I added a guard in `setup_network`, `teardown_network`, `setup_port_forward`, `teardown_port_forward`  in `src/firewall/firewalld.rs` to `return Ok` if the network was internal. The bats tests confirmed that no forwarding gets configured and the subnet doesn't get added to a firewall zone, and no netavark policy is created.

mheon (approved):
LGTM

test/250-bridge-nftables.bats
```diff
@@ -1426,3 +1426,115 @@ EOF
...
+    assert "${lines[4]}" =~ "ct mark & 0x00001000 == 0x00001000 accept" "accept dnat traffic"
+    assert "${lines[5]}" =~ "jump NETAVARK-ISOLATION-1" "jump to isolation chain"
+    assert "${#lines[@]}" = 8 "no per-network FORWARD rules for internal network"
+
+    # POSTROUTING: only the base mark-masquerade rule, no per-network jump
+    run_in_host_netns nft list chain inet netavark POSTROUTING
+    assert "${lines[3]}" =~ "meta mark & 0x00002000 == 0x00002000 masquerade" "Mark-masquerade rule"
+    assert "${#lines[@]}" = 6 "no per-network POSTROUTING rule for internal network"
+
+    # after the redirect DNS arrives on dns_port, so accept that under the drop policy
```
Luap99, line 1525: can you file a new bug for that, I would say we need to create that rule ourselves in nv

Luap99 (approved):
LGTM

PR #1503, opened 2026-08-17
Makefile: drop redundant $(CARGO_TARGET_DIR) rule

### LINKED ISSUE #1333: No need for ` mkdir -p -$(CARGO_TARGET_DIR)` in Makefile
In `Makefile` the target `$(CARGO_TARGET_DIR): mkdir -p $@` is not necessary.  With this change:
```diff
diff --git a/Makefile b/Makefile
--- a/Makefile
+++ b/Makefile
@@ -49,11 +49,8 @@ bin:
        mkdir -p $@
 
-$(CARGO_TARGET_DIR):
-       mkdir -p $@
-
 .PHONY: build
 build: bin/netavark
 
-bin/netavark: $(SOURCES) bin $(CARGO_TARGET_DIR)
+bin/netavark: $(SOURCES) bin
        $(CARGO) build $(release)
        cp $(CARGO_TARGET_DIR)/$(profile)/netavark bin/netavark
```

`CARGO_TARGET_DIR=abc make`, `make CARGO_TARGET_DIR=def` and `make` do create the target directory as expected using cargo 1.89.0.

magic-peach:
cargo creates its own target directory, so the `mkdir -p` rule for `$(CARGO_TARGET_DIR)` is not doing anything.

The diff in the issue only drops the rule and the prerequisite on `bin/netavark`, but six other targets also list `$(CARGO_TARGET_DIR)` as a prerequisite: `examples`, `client`, `build_unit`, `unit`, `integration` and `validate`. With the rule gone and those left alone they all fail with `No rule to make target 'targets'` on a fresh checkout, so this drops the prerequisite everywhere.

The variable stays, it is still used for the `cp` paths in `bin/netavark` and by `clean`.

Testing: `make -n` on all seven targets with no `targets/` present, and a real build with `CARGO_TARGET_DIR` pointing at a path that did not exist, which cargo created before copying the binaries. `make unit` gives the same 7 passed / 5 failed as unmodified main in my container, the failures are `unshare(CLONE_NEWNET): EPERM` and unrelated to this. `cargo fmt --all -- --check` is clean. I have not run the bats suite.

Fixes: #1333

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,13 +48,10 @@ all: build docs
 bin:
 	mkdir -p $@
 
-$(CARGO_TARGET_DIR):
-	mkdir -p $@
-
 .PHONY: build
 build: bin/netavark
 
-bin/netavark: $(SOURCES) bin $(CARGO_TARGET_DIR)
+bin/netavark: $(SOURCES) bin
 	$(CARGO) build $(release)
 	cp $(CARGO_TARGET_DIR)/$(profile)/netavark bin/netavark
 	cp $(CARGO_TARGET_DIR)/$(profile)/netavark-dhcp-proxy-client bin/netavark-dhcp-proxy-client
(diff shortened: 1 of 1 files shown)


mheon:
Sure, LGTM

Luap99 (approved):
LGTM

PR #1507, opened 2026-08-18
bump netlink-packet-{core,route}

Luap99:
Update netlink-packet-route to 0.33.0 and netlink-packet-core to 0.9.0 as they depend on each other.


mheon (approved):
LGTM

PR #1509, opened 2026-08-20
firewall: fix DNS accept rule port on internal networks with custom dns_port

### LINKED ISSUE #1508: Internal networks missing mark-masquerade rule for DNS under drop policy
Follow up to #1500 

For internal networks with DNS enabled under a drop firewall policy, the generated POSTROUTING chain is missing a mark-masquerade rule. `test/250-bridge-nftables.bats` only showed us 5 rule lines instead of the expected 6 missing `meta mark & 0x00002000 == 0x00002000 masquerade`  This rule is needed to properly masquerade DNS reply traffic that's been marked (0x2000) as allowed under the drop policy for internal networks. Without it, the return path for DNS traffic on internal networks isn't correctly rewritten. Netavark should generate this per network POSTROUTING rule itself for internal networks rather than assuming it exists elsewhere in the ruleset.

ajoshua2004:
The INPUT accept rule was hardcoded to check port 53, but when a custom dns_port is configured the DNAT redirect rewrites the destination port before the packet reaches INPUT, so the accept rule never matched. Changed it to use `network_setup.dns_port` instead.

Fixes: #1508

### DIFF
--- a/src/firewall/nft.rs
+++ b/src/firewall/nft.rs
@@ -523,7 +523,7 @@ impl firewall::FirewallDriver for Nftables {
                 }
 
                 // Next, populate basic chains with forwarding rules
-                // Input chain: ip saddr <subnet> udp dport 53 accept
+                // Input chain: ip saddr <subnet> udp dport <dns_port> accept
                 batch.add(make_rule(
                     Cow::Borrowed(INPUTCHAIN),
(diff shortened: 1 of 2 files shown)


mheon:
Sure, LGTM

Luap99 (approved):
LGTM

